PR #3, opened 2015-04-09
Avoid modifications in dictionary during iteration

SylvainCorlay:
Calls to monkey-patched _notify_trait resulting from cross validation may cause the `cache` dictionary to be modified.


minrk:
If this fixes a bug, can you add a test for the bug?

SylvainCorlay:
`IntSlider(min=1)` triggers the bug. I will add a test.

SylvainCorlay:
Added a test.

PR #7, opened 2015-04-17
add traitlets.utils.importstring

minrk:
moved from genutils, since I think it makes more sense as part of traitlets,
since most usage is traitlets-related, even when not in traitlets itself.

import_item is available as a public top-level API as `traitlets.import_item`

I also created a utils subpackage for the few utils so far.


Carreau:
I suppose you will open a PR on genutils to remove it, one this is merged ?

+1

minrk:
I think so.

PR #10, opened 2015-04-21
Adding key-specific dict validation

SylvainCorlay:
Now, Dict trait type's ctor has an additional optional argument, `traits`. `traits` is a dictionary of trait types. 
When provided, the element validation stage first checks for the traittype of the corresponding key, before defaulting to the general `trait` used for other keys.
Moreover, the corresponding trait type may have a `required` metadata. If True, the corresponding key will be required.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1717,15 +1717,22 @@ class Dict(Instance):
     """An instance of a Python dict."""
     _trait = None
 
-    def __init__(self, trait=None, default_value=NoDefaultSpecified, **metadata):
+    def __init__(self, trait=None, traits=None, default_value=NoDefaultSpecified,
+                 **metadata):
         """Create a dict trait type from a dict.
 
         The default value is created by doing ``dict(default_value)``,
         which creates a copy of the ``default_value``.
 
         trait : TraitType [ optional ]
-            the type for restricting the contents of the Container.  If unspecified,
-            types are not checked.
+            The type for restricting the contents of the Container. If
+            unspecified, types are not checked.
+
+        traits : Dictionary of trait types [optional]
+            The type for restricting the content of the Dictionary for certain
+            keys.
+            If the trait types are marked as `required`, the validation stage
+            will check that a value is present for that key.
 
         default_value : SequenceType [ optional ]
             The default value for the Dict.  Must be dict, tuple, or None, and
@@ -1736,10 +1743,13 @@ class Dict(Instance):
             Whether to allow the value to be None
 
         """
+        # Handling positional arguments
         if default_value is NoDefaultSpecified and trait is not None:
             if not is_trait(trait):
                 default_value = trait
                 trait = None
+
+        # Handling default value
         if default_value is NoDefaultSpecified:
             default_value = {}
         if default_value is None:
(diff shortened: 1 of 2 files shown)


takluyver:
What's the use case for this? It seems like the sort of thing we had said should be a subclass in widgets, rather than adding more functionality to traitlets itself.

SylvainCorlay:
The usecase is to use trait types to specify a schema

```Python
class Foo(HasTraits):
    bar = Dict(traits={
        'name': Unicode(required=True),
        'age': Int(),
        'preferences': Dict(traits={
            'language': Unicode('English'),
        }, required=True),
    })
```

It is not specific to widgets but my use cases are motivated by widgets. 

There might be other places where you would want to enforce a schema for a Dict attribute (such as the config system.)

takluyver:
We've got by without that in config so far, though. My understanding is that new features motivated by widget use cases should go in subclasses within widgets, even if they are more generally applicable in principle.

SylvainCorlay:
Indeed, in every case, with valid values, things should work without this kind of validation :) 
It is probably more valuable for things like widgets because it provides meaningful error messages in interactive mode, but would still be useful elsewhere. In any case, one can have their own trait type doing the same thing.

SylvainCorlay:
@takluyver, @minrk what's your take on this finally? This would allow specifying simple schemas for HasTrait attributes.

minrk:
I think making Dict match List in terms of validation makes a certain amount of sense. I would probably only implement the simplest case, and not include the `required` metadata feature here. More advanced validation can be implemented much more easily with custom trait types. I do not think we should be implementing a full schema system inside the Dict trait. I think custom validation callbacks and/or trait subclasses implement this much better than elaborate trait instantiation.

The cases I think are reasonable to cover in the base Dict trait:

- specify value_type (only thing currently supported, I think)
- specify key_type, value_type (like `Map[key_type] => value_type` in other languages)
- (maybe) specify keys and the types of their values (all keys required)

SylvainCorlay:
Ok, thanks. I will remove the `required` feature. 

I don't have a usecase for specifying `key_type` since json requires strings, but it could be a good idea for the example that you are describing (associating a key type to a value type). I will open a PR.

minrk:
I think leaving out key_type would be fine, too. I wouldn't really expect people to use it in Python. I just got the idea from mapping types in other languages.

SylvainCorlay:
Removed the `required` keyword. Now all specified keys are considered required. I am not sure if I should rather consider everything as non-required.

SylvainCorlay:
up

traitlets/traitlets.py
```diff
@@ -1751,13 +1759,19 @@ def __init__(self, trait=None, default_value=NoDefaultSpecified, **metadata):
         else:
             raise TypeError('default value of Dict was %s' % default_value)
 
+        # Case where a type of TraitType is provided eather than an instance
```
minrk: s/eather/rather/

minrk:
Sorry about that. Found a small typo, but otherwise looks good. Thanks for the ping.

SylvainCorlay:
Rebased - fixing the typo.

burnpanck:
I wasn't aware that I'm again disagreeing to some extent with a feature by you @SylvainCorlay, so please don't take it personal ;-).

I think the specification of schemas should go to a subclass `Schema` of the `Dict` trait. Personally, I'd most often use a class for such a use-case instead. I already see an opportunity for some meta-programming (yay!):
```python
class Foo(HasTraits):
    @Schema(default_value={})
    class Bar:
        name=Unicode()
        age=Int()
        @Schema()
        class preferences:
            language = Unicode('English')
```
Alternatively, without the decorator saving you one line (but with some `__new__` magic): `class Bar(Schema(default_value={}).tag()): ...`. (Should I attempt a PR with this?)

More importantly, I think the `key_type` and `value_type` sub-traits are much more important. I often use a `Dict` as a part of an algorithm or complex data-structure, where the keys could be ID's or even instances of another class and so on. I would like to be able to specify these things in the trait. A quick grep over my personal code-base shows that essentially all of my `Dict`s (the original ones from enthought) have both a key and a value trait and and only a small part of the key traits are strings.

PR #14, opened 2015-04-29
Ignore unspecified default value until first access

takluyver:
Alternative to #13

Previously, if:

- The traitlet class does not specify a default value,
- The use of the traitlet does not provide a default value,
- The class using the traitlet does not have a _foo_default method, and
- The instantiation of that class does not provide a value for the
traitlet

then the traitlet would try to set its value to the undefined default, usually causing an error. With this change, it will skip that, so the traitlet will have no value. If it is accessed before a value has been
set, the error will occur then instead.

This allows Enums with no default value, which I want for kernelspec.

`Instance` traitlets are not affected - they will still create an instance of the specified class when the class using them is instantiated.

Ping @ssanderson @SylvainCorlay

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -382,6 +382,20 @@ class TraitType(object):
         obj._trait_values[self.name] = value
         return value
 
+    def _setup_dynamic_initializer(self, obj):
+        # Check for a deferred initializer defined in the same class as the
+        # trait declaration or above.
+        mro = type(obj).mro()
+        meth_name = '_%s_default' % self.name
+        for cls in mro[:mro.index(self.this_class)+1]:
+            if meth_name in cls.__dict__:
+                break
+        else:
+            return False
+        # Complete the dynamic initialization.
+        obj._trait_dyn_inits[self.name] = meth_name
+        return True
+
     def set_default_value(self, obj):
         """Set the default value on a per instance basis.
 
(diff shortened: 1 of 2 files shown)


takluyver:
Don't merge this just yet, it appears to break starting IPython. Will investigate.

takluyver:
OK, the problem was that a traitlet where `allow_none` was True was not getting a default value. I've changed it so that None is the default default if `allow_none` is True.

SylvainCorlay:
Does it make sense to have the default value depend on `allow_none`?

SylvainCorlay:
Actually In general, I would be in favor of both #13 (the default for an enum being the first item unless specified otherwise), and something deferring instantiation / validation of  defaults to first access.

takluyver:
Well, the case I hit was pylab, where the default definitely should be None. I think changing the default to the first value would be surprising in cases like that.

I would look at the current state not as the default depending on allow_none, but as it having a valid default state if allow_none is True. i.e. it's always unset by default, but if allow_none is True, that's an acceptable state in which to use it.

takluyver:
Can you give some examples of other cases where delaying instantiation or validation to first access makes sense? I might be able to work them in here.

SylvainCorlay:
- Re: pylab, then the code should probably do `default_value=None` right?
- Re: delaying default values. The main use case I have in mind is the case of an Instance trait type. 

```Python
class Foo(Widget):
    bar= Instance(Bar)
```
Then

```Python
# The following works
bar= Foo(bar=SomeBar()) 

# The following tries instantiating an invalid default value for `bar`
# even though it is not ever used
foo= Foo()  # raises TraitError
foo.bar= SomeBar()
```

takluyver:
In that case, where you just define it as `bar = Instance(Bar)`, the default value will be `None`. If I changed that to `Undefined` in this PR, it would behave like Enums, with no default value and an error if you try to access it before it's set to something.

If you provide arguments, like `bar = Instance(Bar, (1,))`, it will instantiate a default value with `Bar(1)`. Do you still want that to happen at instantiation, or should it also be delayed until first access?

I think `None` is a bit confusing here, because there's multiple levels where we want to allow something to be unspecified, so we now have three sentinels in traitlets: None, NoDefaultSpecified and Undefined. I wonder if we could get rid of one of those...

SylvainCorlay:
I agree that `None` does not make sense as a default value for `Instance`. Undefined would make more sense.

minrk:
@takluyver can you add a couple of tests to exercise this? It's working form me with:

```python
class C(HasTraits):
    t = Enum(['a', 'b'])

c = C()
c.t  = 'a'
assert c.t == 'a'

c = C()
with assertRaises(TraitError):
    t = c.t

c = C(t='b')
assert c.t == 'b'
```

So I think we can merge once there is a test for the new behavior.

takluyver:
Added that bit of code as a test, and it passes. :-)

traitlets/traitlets.py
```diff
@@ -1419,8 +1431,10 @@ def validate(self, obj, value):
 class Enum(TraitType):
     """An enum that whose value must be in a given sequence."""
 
-    def __init__(self, values, default_value=None, **metadata):
+    def __init__(self, values, default_value=NoDefaultSpecified, **metadata):
         self.values = values
+        if metadata.get('allow_none', False) and default_value is NoDefaultSpecified:
+            default_value = None
```
SylvainCorlay: This makes the default value depend on the value of `allow_none`. Why not `Undefined`?

PR #17, opened 2015-05-20
Create docs about traitlets

takluyver:
- Sphinx 1.3 has a new extension called [napoleon](http://sphinx-doc.org/ext/napoleon.html), which aims to parse numpy and Google style docstrings. I tried using it instead of numpydoc. It appears to work for the stuff in traitlets.

- I documented the four possible signatures for `_traitname_changed` callbacks. Do we want to take the refactoring as an opportunity to deprecate some of them? I'm inclined to leave them as they are, but it's worth considering.

- Some work is needed on the config docs (which were in the repo before this PR) to separate the description of the config system from information about how IPython uses it (e.g. profiles). But I think that can be a separate PR.

### DIFF
--- a/docs/source/trait_types.rst
+++ b/docs/source/trait_types.rst
@@ -11,16 +11,20 @@ Trait Types
 
 Numbers
 -------
-.. autoclass:: Int
-
-.. class:: Long
-
-   Integers ``> sys.maxsize`` on Python 2. Alias for :class:`Int` on Python 3.
 
 .. class:: Integer
 
-   Uses the int or long types as appropriate on Python 2. Alias for :class:`Int`
-   on Python 3.
+   An integer trait. On Python 2, this automatically uses the ``int`` or
+   ``long`` types as necessary.
+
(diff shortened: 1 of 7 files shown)


docs/source/trait_types.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Numbers
+-------
+.. autoclass:: Int
+
+.. class:: Long
+
+   Integers ``> sys.maxsize`` on Python 2. Alias for :class:`Int` on Python 3.
+
+.. class:: Integer
```
minrk: Can there be a note on Int and Long that under ~all circumstances, Integer should be used instead? Int and Long should only be used when the opposite would be invalid, which is quite rare.

takluyver: Done.

Carreau:
Should we merge that and iterate ?

takluyver:
That sounds good. There's nothing more I want to do on this before merge.

Issue #20, opened 2015-05-27
Deprecate/remove help_classes, config_classes?


takluyver:
At some point, we added functionality to describe different lists of classes in the `--help-all` output and the config file. I think this might have been when frontend config could be passed to kernels, but we didn't want to show those options in help output.

We're not using this functionality now, as far as I can see, and it's a little bit of extra confusing indirection to follow. I'd suggest we get rid of it again and just use `classes`.

minrk:
I think you are right - this was only used in ConsoleApp as far as I can tell, where CLI args could affect the kernel, but config files could not. ConsoleApp no longer has this behavior.

PR #21, opened 2015-05-28
API to generate rST docs for configurable options

takluyver:
Corresponding PR to ipython/ipython coming up.

I discussed this yesterday with @minrk , and I think that each config option should only be described once, on the most general class that it applies to, and neither the config documentation nor the generated default config file should try to describe inheritance relationships. To this end, there are a couple of new methods:

- `Application._classes_inc_parents()` walks through the list of configurable classes for the application, yielding them along with all configurable parents.
- `HasTraits.class_own_traits()` is like `.class_traits()`, but only takes the traits defined on this class, not on its parents.

Also gets rid of help_classes and config_classes (#20).

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -372,44 +372,11 @@ class Application(SingletonConfigurable):
 
         Returns a multiline string.
         """
-        lines = []
-        for cls in self.classes:
-            classname = cls.__name__
-            for k, trait in sorted(cls.class_traits(config=True).items()):
-                ttype = trait.__class__.__name__
+        chunks = []
+        for cls in self._classes_inc_parents():
+            chunks.append(cls.class_config_rst_doc())
 
-                termline = classname + '.' + trait.name
-
-                # Choices or type
-                if 'Enum' in ttype:
-                    # include Enum choices
-                    termline += ' : ' + '|'.join(repr(x) for x in trait.values)
-                else:
-                    termline += ' : ' + ttype
-                lines.append(termline)
-
-                # Default value
-                try:
-                    dv = trait.get_default_value()
-                    dvr = repr(dv)
-                except Exception:
-                    dvr = dv = None # ignore defaults we can't construct
-                if (dv is not None) and (dvr is not None):
-                    if len(dvr) > 64:
-                        dvr = dvr[:61]+'...'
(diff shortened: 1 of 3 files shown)


Carreau:
That will be of interest for @jhamrick and nbgrader.

traitlets/config/application.py
```diff
@@ -362,6 +367,17 @@ def print_help(self, classes=False):
...
+    def document_config_options(self):
+        """Generate rST format documentation for the config options this application
+
+        Returns a multiline string.
+        """
+        chunks = []
+        for cls in self._classes_inc_parents():
+            chunks.append(cls.class_config_rst_doc())
+
+        return '\n'.join(chunks)
```
Carreau: I'm sad that `_classes_inc_parents` is a generator and that you for loop and not `map(cls, lambda cls:cls.class_config_rst_doc)`. I'm sure there is even a way to get rid of the lambda.

traitlets/config/application.py
```diff
@@ -128,15 +128,20 @@ class Application(SingletonConfigurable):
...
-        return getattr(self, 'config_classes', self.classes)
+
+    def _classes_inc_parents(self):
+        """Iterate through configurable classes, including configurable parents
+
+        Children should always be after parents, and each class should only be
+        yielded once.
+        """
+        seen = set()
+        for c in self.classes:
```
Carreau: Want to make sure this is sorted to have doc order stability ?

traitlets/config/application.py
takluyver: I could do a genexp here, would that make you happier?

traitlets/config/application.py
takluyver: classes is a list, and I think we do use the order already in that list to group things meaningfully.

traitlets/config/application.py
Carreau: yield from all the things.

traitlets/config/application.py
Carreau: Ah, yes indeed. fine then.

takluyver:
Made the code slightly briefer by using a genexp and a dict comprehension.

Carreau:
+1

PR #22, opened 2015-05-29
Create BaseDescriptor parent class to TraitType and add obj argument to instance_init

SylvainCorlay:
As I am working on widget signals and slots, this is meant to simplify the extension of traitlets in widgets.

- created a `BaseDescriptor` class, a parent to TraitType which is used in the metaclass - so that one does not need to subclass the metaclass for signals and slots.
- `instance_init` now takes an `obj` argument, that is the underlying `HasTraits` instance. On this regard, the documentation was already saying:

    ```
    Part of the initialization which may depend on the underlying HasTrait instance
    ```

    and it was not the case anymore.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -300,8 +300,14 @@ dlink = directional_link
 # Base TraitType for all traits
 #-----------------------------------------------------------------------------
 
+class BaseTraitType(object):
+    """Base class for TraitType
 
-class TraitType(object):
+    This allows other classes than TraitType to hook into the MetaHasTraits
+    metaclass machinery.
+    """
+
+class TraitType(BaseTraitType):
     """A base class for all trait descriptors.
 
     Notes
@@ -365,8 +371,8 @@ class TraitType(object):
         """Create a new instance of the default value."""
         return self.default_value
 
-    def instance_init(self):
-        """Part of the initialization which may depends on the underlying
+    def instance_init(self, obj):
+        """Part of the initialization which may depend on the underlying
         HasTraits instance.
 
         It is typically overloaded for specific trait types.
@@ -542,7 +548,7 @@ class MetaHasTraits(type):
         # print "MetaHasTraitlets (bases): ", bases
         # print "MetaHasTraitlets (classdict): ", classdict
         for k,v in iteritems(classdict):
-            if isinstance(v, TraitType):
+            if isinstance(v, BaseTraitType):
                 v.name = k
             elif inspect.isclass(v):
                 if issubclass(v, TraitType):
@@ -590,9 +596,9 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
             except AttributeError:
                 pass
             else:
-                if isinstance(value, TraitType):
-                    value.instance_init()
-                    if key not in kw:
+                if isinstance(value, BaseTraitType):
+                    value.instance_init(inst)
+                    if isinstance(value, TraitType) and key not in kw:
                         value._set_default_value_at_instance_init(inst)
         inst._cross_validation_lock = False
         return inst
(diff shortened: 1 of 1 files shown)


minrk:
What does BaseTraitType being an empty unused class accomplish?

SylvainCorlay:
Any class inheriting from it has its attributes `this_class` and `name` set.
I use it in https://github.com/ipython/ipywidgets/pull/39 for signals and slots, which inherit from `BaseTraitType`.

minrk:
That should probably be explained and declared as attributes on the class, rather than an unused empty class with no API. Are any methods called on this class? If I create a BaseTraitType and declare no attributes or methods, will that work?

SylvainCorlay:
This class has no attribute and no method. It is an empty type used for semantic dispatch by the metaclass.

I don't understand what you mean by *'declared as attributes on the class'* . What is the problem with an empty semantic class with no API? 
(btw we do this all the time for metaprog in c++, e.g. for tag-dispatching `struct input_iterator_tag { };`)

minrk:
> What is the problem with an empty semantic class with no API? 

Then it's super unclear what it's for, what it does, and what APIs on it will be called. How do I know what a valid sublcass is? Looking at the class, I see no information about what it could possibly be used for. Should I expect a working instance of this class to have certain attributes or methods? If so, the class should have those attributes and methods defined.

SylvainCorlay:
OK, I will elaborate in the docstring and provide some empty implementation of some methods.

minrk:
I would expect `class MyThing(BaseTraitType): pass` to either:

1. be completely valid, or
2. provide a complete interface, with must-override methods raising `NotImplementedError`.

takluyver:
This is again the kind of discussion that prompted us to propose giving widgets its own copy of traitlets. IIRC, we decided not to because we thought most of the things widgets needed could be done by subclassing and extending.

SylvainCorlay:
This change is meant to provide the hooks so that I can do all the extensions in widgets, with minimal changes in the traitlets. 
 - the addition of the base class really cannot have an impact functionally. 
 - the change on instance_init makes it closer to what it was before, and more consistent with the documentation.

minrk:
I think instance_init getting `obj` makes sense, given what it's for. I'm still a little unclear on the BaseTraitType not seeming to have anything to do with traits. @SylvainCorlay Have you looked at providing that functionality in a subclass of MetaHasTraits? Or does it really have to be an empty base class beneath TraitType?

SylvainCorlay:
Actually `BaseTraitType` would probably have an empty implementation of `instance_init`, since it is the only thing in common (and required) between `TraitType` on the one hand and `Signal` or `Slot` on the other hand (besides having `this_class` and `name`).

SylvainCorlay:
Regarding subclassing MetaHasTraits, I need to think about how this would work. It seems complicated though because the base Widget class could not inherit from HasTraits anymore. 

That is why I thought I had a solution with minimal impact by creating a base for TraitType.

SylvainCorlay:
BaseTraitType could be called Descriptor or something like this. It just takes what is required by the metaclass.

traitlets/traitlets.py
```diff
@@ -319,6 +320,23 @@ class that does the following:
        accept superclasses for :class:`This` values.
     """
 
+    def instance_init(self, obj):
```
minrk: can you add a `name = None` class attribute here, as documentation that this will be assigned by the metaclass?

minrk:
@SylvainCorlay thanks, I think maybe BaseDescriptor is a good choice.

traitlets/traitlets.py
SylvainCorlay: Done (and for this_class too)
- although what is set by the metaclass is an instance attribute. Let me know if you want me to remove it.

SylvainCorlay:
@minrk done. Let me know what you think.

traitlets/traitlets.py
minrk: This is good, thanks. I know it's an instance attribute, but it's a nice convention to define instance attributes as class attributes, as a sort of minimal documentation of the interface.

minrk:
Thanks.

PR #23, opened 2015-06-02
Simplify deferred creation of default values

SylvainCorlay:
This is an attempt to simplify initialization of default values. 

The behavior is slightly different in that now:
```Python
class Foo(HasTraits):
    bar = Int()

# triggers a trait notification
foo = Foo()
foo.bar = 0

# does not triggers a trait notification
foo = Foo(bar=0)
foo.bar = 0

# accessing the uninitialized trait initiates the default value
foo = Foo()
print(foo.bar)  # prints 0
```

Previously, the first example would not trigger a trait notification since 0 is equal to the default value which was initialized. I realize this is quite a change, but it really simplifies the code.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -84,11 +84,6 @@ class TestTraitType(TestCase):
         a = A()
         self.assertEqual(a.tt, 10)
 
-        # Defaults are validated when the HasTraits is instantiated
-        class B(HasTraits):
-            tt = MyIntTT('bad default')
-        self.assertRaises(TraitError, B)
-
     def test_info(self):
         class A(HasTraits):
             tt = TraitType
@@ -118,7 +113,6 @@ class TestTraitType(TestCase):
         self.assertEqual(a.x, 11)
         self.assertEqual(a._trait_values, {'x': 11})
         b = B()
-        self.assertEqual(b._trait_values, {'x': 20})
         self.assertEqual(list(a._trait_dyn_inits.keys()), ['x'])
         self.assertEqual(b.x, 20)
         c = C()
@@ -200,9 +194,9 @@ class TestHasTraitsNotify(TestCase):
         a = A()
         a.on_trait_change(self.notify1)
         a.a = 0
-        self.assertEqual(len(self._notify1),0)
+        self.assertEqual(len(self._notify1), 1)
         a.b = 0.0
-        self.assertEqual(len(self._notify1),0)
+        self.assertEqual(len(self._notify1), 2)
         a.a = 10
         self.assertTrue(('a',0,10) in self._notify1)
         a.b = 10.0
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
@takluyver

SylvainCorlay:
(This actually does what @minrk was suggesting there https://github.com/ipython/ipython/pull/8107#issuecomment-87050310)

traitlets/traitlets.py
```diff
@@ -865,11 +832,10 @@ def trait_metadata(self, traitname, key, default=None):
         else:
             return trait.get_metadata(key, default)
 
-    def add_trait(self, traitname, trait):
+    def add_trait(self, name, trait):
         """Dynamically add a trait attribute to the HasTraits instance."""
         self.__class__ = type(self.__class__.__name__, (self.__class__,),
-                              {traitname: trait})
-        trait.set_default_value(self)
+                              {name: trait})
```
SylvainCorlay: Btw, this works for any descriptor.

SylvainCorlay:
Just rebased to come on top of https://github.com/ipython/traitlets/pull/26.

traitlets/tests/test_traitlets.py
```diff
@@ -600,23 +589,6 @@ class C(HasTraits):
...
-
-        self.assertRaises(TraitError, A)
-
-    def test_instance(self):
-        class Foo(object): pass
-
-        def inner():
-            class A(HasTraits):
-                inst = Instance(Foo())
-
```
minrk: It seems like these removed tests should still be failing, rather than removed.

minrk: If it requires adding a request for the trait, that's okay, too, but the tests shouldn't be removed altogether.

SylvainCorlay:
Rebased upon #28

takluyver:
I think we can avoid the behaviour change if, in `__set__`, we change this:

```python
try:
    old_value = obj._trait_values[self.name]
except KeyError:
    old_value = Undefined
```

to this:

```python
try:
    old_value = obj._trait_values[self.name]
except KeyError:
    old_value = self.default_value
```

`self.default_value` is `Undefined` in cases where there's no obvious default, but it is e.g. `0` for an integer if no default was supplied.

SylvainCorlay:
@takluyver I tried your suggestion but it still fails the test for bad defaults.

traitlets/traitlets.py
```diff
@@ -311,15 +311,25 @@ class that does the following:
     name = None
     this_class = None
 
+    def top_instance_init(self, obj, **kwargs):
```
takluyver: I'd like to avoid adding yet another method, especially in a PR called 'Simplify...'

takluyver:
It looks like you've got rid of all the changes to tests - does that mean that it no longer changes behaviour?

SylvainCorlay:
No, a commit is missing. Just need a minute.

traitlets/traitlets.py
SylvainCorlay: This is part of #28, can we debate this there?

takluyver: It seemed like these two PRs were part of the same thing, but I guess that's not how you see it? Should I ignore this one until we have reached a decision on #28?

SylvainCorlay: I guess so.

SylvainCorlay:
Closing this one for now, since it changes the behavior and we are close to 4.0.

PR #24, opened 2015-06-03
move login from jupyter_core

Carreau:
To keep file_name in sync and correct.

cc @takluyver sharp-eye


takluyver:
Seems reasonable.

PR #26, opened 2015-06-04
TraitType -> BaseDescriptor, and replace call to set_default_value with instance_init

SylvainCorlay:
- Missed a replacement of TraitType by BaseDescriptor.
- add_trait should call `instance_init`, not `set_default_value`. Then there is no remaining call to set_default_value, which I remove.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -551,11 +551,11 @@ class MetaHasTraits(type):
     def __init__(cls, name, bases, classdict):
         """Finish initializing the HasTraits class.
 
-        This sets the :attr:`this_class` attribute of each TraitType in the
+        This sets the :attr:`this_class` attribute of each BaseDescriptor in the
         class dict to the newly created class ``cls``.
         """
         for k, v in iteritems(classdict):
-            if isinstance(v, TraitType):
+            if isinstance(v, BaseDescriptor):
                 v.this_class = cls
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@minrk

SylvainCorlay:
Ping @minrk, @takluyver ? this one is quite trivial.

minrk:
It's surprising to me that simply removing the call to set_default_value has has no effect. instance_init doesn't appear to set the default value, so functionality does seem to be removed. Can you explain that a little bit?

SylvainCorlay:
- `instance_init` should clearly be called in `add_trait`. (e.g. for setting `name` and `this_class` of sub-traits)

- The last call to the `set_default_value` function was in `add_trait`. It is actually not needed. 

*Other calls were removed by earlier PRs. A call was removed when we decided to wait to see if a value was provided in a keyword argument of the constructor of the HasTraits instance. Another call was removed when Thomas when he introduced the `_set_default_value_at_instance_init` method.*

- Example after this PR: 
```Python
from traitlets import *
class Foo(HasTraits):
    pass
foo = Foo()
foo.add_trait('bar', Int(1)) 
print(foo.bar )  # prints 1 because self.init_default_value is called in `__get__`.
```

minrk:
ok. That seems fine, then.

PR #27, opened 2015-06-05
Allow adding multiple traits at once

SylvainCorlay:
As suggested by @sccolbert, allow adding multiple traits at once. This avoids creating too many new types when adding multiple decorators, since every call to add_trait was creating a new one.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -845,11 +845,12 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         else:
             return trait.get_metadata(key, default)
 
-    def add_trait(self, name, trait):
-        """Dynamically add a trait attribute to the HasTraits instance."""
+    def add_traits(self, **traits):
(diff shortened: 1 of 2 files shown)


traitlets/traitlets.py
```diff
@@ -845,10 +845,10 @@ def trait_metadata(self, traitname, key, default=None):
...
             return trait.get_metadata(key, default)
 
-    def add_trait(self, name, trait):
-        """Dynamically add a trait attribute to the HasTraits instance."""
+    def add_traits(self, **traits):
+        """Dynamically add trait attributes to the HasTraits instance."""
         self.__class__ = type(self.__class__.__name__, (self.__class__,),
-                              {name: trait})
+                              traits)
         trait.instance_init(self)
```
minrk: You need to iterate over `traits.values()` here.

minrk:
Other than a missing iteration now that it's a container instead of a single trait, this makes sense to me.

SylvainCorlay:
Thanks,

PR #28, opened 2015-06-05
_setup_dynamic_initializer simplified

SylvainCorlay:
`_setup_dynamic_initializer ` used to return a boolean on whether there is a dynamic initializer for a given trait. It does not anymore.

Two remarks though:
- `instance_init` is supposed to correspond to the part or the initialization that depends on the `HasTraits` instance. 
    All of what is currently in the implementation of `instance_init` for our trait types does not actually depend on `obj`. It could simply be part of `__init__`. There is no real reason to keep it there.
- After we move this, the call to `_setup_dynamic_initializer` could be moved to `TraitType.instance_init`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -365,11 +365,6 @@ class TraitType(BaseDescriptor):
         else:
             self._metadata = self.metadata
 
-        self.init()
-
-    def init(self):
-        pass
-
     def get_default_value(self):
         """Create a new instance of the default value."""
         return self.default_value
@@ -390,20 +385,18 @@ class TraitType(BaseDescriptor):
         # trait declaration or above.
         mro = type(obj).mro()
         meth_name = '_%s_default' % self.name
-        for cls in mro[:mro.index(self.this_class)+1]:
+        for cls in mro[:mro.index(self.this_class) + 1]:
             if meth_name in cls.__dict__:
                 break
         else:
-            return False
-        # Complete the dynamic initialization.
+            return
         obj._trait_dyn_inits[self.name] = meth_name
-        return True
 
     def _set_default_value_at_instance_init(self, obj):
         # As above, but if no default was specified, don't try to set it.
         # If the trait is accessed before it is given a value, init_default_value
         # will be called at that point.
-        if (not self._setup_dynamic_initializer(obj)) \
+        if (self.name not in obj._trait_dyn_inits) \
                 and (self.default_value is not Undefined):
             self.init_default_value(obj)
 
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -390,20 +385,18 @@ def _setup_dynamic_initializer(self, obj):
...
+            return
         obj._trait_dyn_inits[self.name] = meth_name
-        return True
 
     def _set_default_value_at_instance_init(self, obj):
         # As above, but if no default was specified, don't try to set it.
         # If the trait is accessed before it is given a value, init_default_value
         # will be called at that point.
-        if (not self._setup_dynamic_initializer(obj)) \
+        if (self.name not in obj._trait_dyn_inits) \
```
takluyver: I dislike this change - it makes the API more stateful, because `_set_default_value_at_instance_init()` now relies on `_setup_dynamic_initializer` having been called first. But in #23, you remove this method entirely. The two PRs will conflict - what is the relationship between them?

SylvainCorlay:
Thanks for looking at this. I guess I am trying to split things too much into small changes to them in. 

So here is the idea: `instance_init` is supposed to handle the *part of the initialization which may depend on the underlying HasTraits instance*. What `_setup_dynamic_initializer` does falls very much into that category. Hence, this should probably be called in `instance_init`.

One issue is that `instance_init` is also called by descriptors holding other descriptors like `List(Int())`, and we don't want these to be set up by dynamic initializers... Hence there should maybe be two levels for such initialization: one called at the top level only, i.e. in `__new__`, which does a bit more than the one also called for sub-traits.

SylvainCorlay:
Btw, he current content of `instance_init` actually does a lot of other things which actually do no require the HasTraits instance, so much that for a while we removed the corresponding `obj` argument from the signature. The only part that required it, setting up default values got removed. Most of it could actually be done in `TraitType.__init__`.

traitlets/traitlets.py
SylvainCorlay: I already rebased the #23 on this PR. `_trait_dyn_inits` is also used in __get__ directly. `_setup_dynamic_initializer`'s role is to set this variable.

takluyver: OK, should we just be reviewing #23 as a single unit now?

SylvainCorlay: - #28 does not change behavior. It is more of a cosmetic change and could be merge more easily.
- #23 is more involving since it changes how default values are initialized.

takluyver:
As mentioned on #23, I'm not happy about a simplification PR that adds a new public API method. From the diff, this does not now appear to make the code simpler - has what you're trying to achieve changed?

SylvainCorlay:
Too me, having if statements based on `isinstance` is a sign of bad design. It make sense in the context of metaclass magics, but this is better solved with polymorphism.

PR #33, opened 2015-06-15
Bounded Float and Int

SylvainCorlay:
@minrk This allows defining a `min` and `max` attribute to a `Float` trait type.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1274,12 +1274,22 @@ class Float(TraitType):
     default_value = 0.0
     info_text = 'a float'
 
+    def __init__(self, default_value=NoDefaultSpecified,
+                 allow_none=None, **kwargs):
+        self.min = kwargs.pop('min', -float('inf'))
+        self.max = kwargs.pop('max', float('inf'))
+        super(Float, self).__init__(default_value=default_value, 
+                                    allow_none=allow_none, **kwargs)
+
     def validate(self, obj, value):
-        if isinstance(value, float):
-            return value
         if isinstance(value, int):
-            return float(value)
-        self.error(obj, value)
+            value = float(value)
+        if not isinstance(value, float):
+            self.error(obj, value)
+        if value > self.max or value < self.min:
+           raise TraitError("The value of the '%s' trait of %s instance should be between %s and %s"\
+               % (self.name, class_of(obj), self.min, self.max))
+        return value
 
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
(I will add a test if you are fine with this.)

minrk:
I think it makes sense. I would do the same on Integer, though.

SylvainCorlay:
Added a bounded Int. In Python 3.0, `int` don't really have a maximum value. The implementation is a bit different than for Float.

traitlets/traitlets.py
```diff
@@ -1204,10 +1204,24 @@ class Int(TraitType):
...
-        if isinstance(value, int):
-            return value
-        self.error(obj, value)
+        if not isinstance(value, int):
+            self.error(obj, value)
+        if self.max is not None and value > self.max:        
+            raise TraitError("The value of the '%s' trait of %s instance should not be greater than %s"\
+                % (self.name, class_of(obj), self.max))
+        if self.min is not None and value < self.min:
+            raise TraitError("The value of the '%s' trait of %s instance should not be lesser than %s"\
```
takluyver: lesser -> less

takluyver:
The error messages should also include the value that it was trying to set, I think, by analogy with the type mismatch error messages:

"The '%s' trait of %s instance must be %s, but a value of %s was specified."

SylvainCorlay:
@takluyver as per you suggestions.

traitlets/traitlets.py
```diff
@@ -1204,10 +1204,28 @@ class Int(TraitType):
...
-        if isinstance(value, int):
-            return value
-        self.error(obj, value)
+        if not isinstance(value, int):
+            self.error(obj, value)
+        if self.max is not None and value > self.max:        
+            raise TraitError("The value of the '%s' trait of %s instance should "
+                             "not be greater than %s, but a value of %s was "
+                             "specified" % (self.name, class_of(obj),
+                                            self.max, repr_type(value)))
```
takluyver: `repr_type()` will return something like `200 <class 'int'>`. Given that at this point it must be the correct type, we can just show its repr with `%r` in the format string.

SylvainCorlay: done.

takluyver:
:+1:

SylvainCorlay:
green button?

Issue #32, opened 2015-06-14
Deprecate get_default_value method?


takluyver:
From PR #31 - the `get_default_value` method on traits now just returns `self.default_value` in all the traits we define. Dynamically created defaults (e.g. for `Instance` traitlets) are built via a different mechanism, so that we can avoid creating them until they're needed.

If there are no other traitlets defining or using `get_default_value`, we can deprecate it and just use `self.default_value` instead. Widgets [doesn't appear to](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/trait_types.py). Are there any other users of widgets we know about?

CC @minrk @SylvainCorlay

SylvainCorlay:
I am not aware of any code depending on `get_default_value`. :+1: to deprecate it.

PR #34, opened 2015-06-15
Deprecate get_default_value()

takluyver:
Closes gh-32

I also removed validate_default_value(), which I just added in PR #31. Its implementation is a single line, so I decided it was clearer to inline it where it's used.


takluyver:
Don't merge this yet, just about to open a conflicting PR

takluyver:
Rebased now that #35 is merged. @SylvainCorlay , do you want to check that this doesn't break anything with widgets?

minrk:
:+1: will merge after confirmation from @SylvainCorlay

SylvainCorlay:
Sorry for the late reply. All the tests run OK.

PR #35, opened 2015-06-15
Set trait values from static defaults at instance_init

takluyver:
Should fix the problem introduced in gh-31 for the widgets. @SylvainCorlay 

One small behavioural change: traitlets with no default value will raise TraitError when accessed instead of returning Undefined.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -49,7 +49,8 @@ class TestTraitType(TestCase):
         class A(HasTraits):
             a = TraitType
         a = A()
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
I checked out thomas' branch and I can confirm that it fixes the issue.

takluyver:
Thanks. Again, I'd like @minrk to have a quick look before we merge this, but it's a fairly small change.

SylvainCorlay:
Well, in any case, there is no risk I merge anything in traitlets :-)

minrk:
Thanks, guys.

PR #37, opened 2015-06-16
trait_medatada exception

SylvainCorlay:
If `allow_raise` is set to `False`, the method `trait_metadata` will return the default value even if `traitname` is not a valid trait name.

The goal is to optimize things like

```Python
if name in Foo.traits(bar=True):
```
which creates a dictionary and then checks if `name` is in it. Now one can simply do

```Python
if Foo.trait_metadata(name, 'bar', allow_raise=False):
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -841,13 +841,16 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
 
         return result
 
-    def trait_metadata(self, traitname, key, default=None):
+    def trait_metadata(self, traitname, key, default=None, allow_raise=True):
         """Get metadata values for trait by key."""
         try:
             trait = getattr(self.__class__, traitname)
         except AttributeError:
-            raise TraitError("Class %s does not have a trait named %s" %
-                                (self.__class__.__name__, traitname))
+            if allow_raise:
+                raise TraitError("Class %s does not have a trait named %s" %
+                                    (self.__class__.__name__, traitname))
+            else:
+                return default
         else:
             return trait.get_metadata(key, default)
 


takluyver:
`if isinstance(getattr(type(Foo), name, None), TraitType):`?

SylvainCorlay:
What do you mean?

takluyver:
What you want to test is 'does this object have a trait called name?', right? The snippet I posted should do that, though I haven't tested it. It seems preferable to me to do that test explicitly rather than asking for metadata to test if a trait exists.

If this is a common enough case, we could have a `has_trait` method that does this, but I think that's probably not needed. In most cases, the type of an object defines what traits it has, just like it defines what methods it has.

SylvainCorlay:
We currently have a filter `Foo.traits(bar=True)` that creates the Dict of all the trait attributes that have a metadata 'bar' equal to True.

What I am testing is if the trait_metadata 'bar' for 'name' is True. If 'name' is not a trait at all, I want False, not an exception.

takluyver:
OK. So that's implicitly `if has_trait(foo, 'name') and foo.trait_metadata('name', 'bar'):`, right?

I'm wary of making dynamic things too easy, because in the major use cases for traitlets, they are relatively static things which are defined by the class.

Also, if we do this, I think the parameter name should be clearer. I'll have a think about what name I'd want.

SylvainCorlay:
Ok, I think that it is fine. Did no think of the `has_trait` method.

takluyver:
That doesn't actually exist, I was doing pseudocode ;-). But it's an easy addition if you think it's necessary.

SylvainCorlay:
Well, that might be the reason :) opening a PR.

PR #38, opened 2015-06-16
has_trait function

SylvainCorlay:
As per discussion with @takluyver .

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -689,7 +689,6 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                 raise TraitError('a trait changed callback '
                                     'must be callable.')
 
-
     def _add_notifiers(self, handler, name):
         if name not in self._trait_notifiers:
             nlist = []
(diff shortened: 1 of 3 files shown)


minrk:
:+1:, but can you add a simple exercise test, for good measure?

SylvainCorlay:
alright :)

traitlets/traitlets.py
```diff
@@ -801,6 +800,12 @@ def class_own_traits(cls, **metadata):
         return {n: t for (n, t) in cls.class_traits(**metadata).items()
                 if getattr(sup, n, None) is not t}
 
+    def has_trait(self, name):
+        """Returns True if the specified name is the name of a trait
+        attribute."""
+        return (hasattr(self.__class__, name) and
+                isinstance(getattr(self.__class__, name), TraitType))
```
takluyver: getattr takes an optional default parameter, so we should be able to save the separate `hasattr` check and just do `isinstance(getattr(self.__class__, name, None), TraitType)`.

traitlets/traitlets.py
```diff
@@ -801,6 +800,12 @@ def class_own_traits(cls, **metadata):
         return {n: t for (n, t) in cls.class_traits(**metadata).items()
                 if getattr(sup, n, None) is not t}
 
+    def has_trait(self, name):
+        """Returns True if the specified name is the name of a trait
+        attribute."""
```
takluyver: The summary in a docstring should generally be kept to one line. Something like "Returns True if the object has a trait with the specified name."

SylvainCorlay:
done, squashed in a single commit.

takluyver:
Can you also add the method to the HasTraits description in the docs (`using_traitlets.rst`)? Sorry I didn't think of that before?

SylvainCorlay:
Done

takluyver:
Thanks

PR #39, opened 2015-06-17
Use NoDefaultSpecified as sentinel for Type default

takluyver:
Instead of None

Prerequisite for fixing jupyter/nbconvert#41

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -453,6 +453,41 @@ class TestType(TestCase):
         self.assertEqual(a.klass, B)
         self.assertRaises(TraitError, setattr, a, 'klass', 10)
 
+    def test_default_options(self):
+
+        class B(object): pass
+        class C(B): pass
+        class A(HasTraits):
+            # Different possible combinations of options for default_value
+            # and klass. default_value=None is only valid with allow_none=True.
+            k1 = Type()
+            k2 = Type(None, allow_none=True)
+            k3 = Type(B)
+            k4 = Type(klass=B)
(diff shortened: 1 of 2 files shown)


takluyver:
WIP

takluyver:
OK, this is ready for review now. The logic around default values for `Type()` traitlets is a bit hairy, but I think what's there now should work.

klass | default_value | desired k | desired dv
-------|------------------|--------------|-------------
None | NDS | object | object
None | None | object | None
None | value | value | value
value | NDS | value | value
value | None | value | None
value | value2 | value | value2

takluyver:
This is necessary because if we use `None` as the value meaning 'nothing was specified', there's no way to explicitly ask for a default of None, which is what we really want to do in nbconvert.

jdfreder:
Yay two null types, like Javascript : ) .  Have you thought about naming this as `Undefined` instead of `NoDefaultSpecified`?  That really is what `undefined` is for...

takluyver:
There's Undefined in there as well. I'd like to collapse Undefined and NoDefaultSpecified down to one object, but that's another PR. For the moment, Undefined is what's used as an attribute, and NoDefaultSpecified is used in constructor parameters.

SylvainCorlay:
+1 to make it a different PR for collapsing. I recall that I looked into that and there was a lot more to change.

minrk:
Makes sense. Want to add some tests to exercise the different cases? I could imagine getting a little mixed up and not noticing that we broke something here for a while.

minrk:
Also :+1: for collapsing Undefined and NoDefaultSpecified.

takluyver:
Added tests

minrk:
huzzah!

PR #41, opened 2015-06-29
Don't deepcopy items when merging two config objects

minrk:
creates unnecessary duplicates, and causes problems with LazyConfigValues

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -184,14 +184,14 @@ class Config(dict):
         to_update = {}
         for k, v in iteritems(other):
             if k not in self:
-                to_update[k] = copy.deepcopy(v)
+                to_update[k] = v
             else: # I have this key
                 if isinstance(v, Config) and isinstance(self[k], Config):
                     # Recursively merge common sub Configs
                     self[k].merge(v)
                 else:
                     # Plain updates for non-Configs
-                    to_update[k] = copy.deepcopy(v)
+                    to_update[k] = v
 
         self.update(to_update)
     
--- a/traitlets/config/tests/test_loader.py
+++ b/traitlets/config/tests/test_loader.py
@@ -407,13 +407,13 @@ class TestConfig(TestCase):
         assert isinstance(foo, LazyConfigValue)
         self.assertIn('foo', cfg)
     
-    def test_merge_copies(self):
+    def test_merge_no_copies(self):
         c = Config()
         c2 = Config()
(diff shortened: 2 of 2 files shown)


takluyver:
Before we merge new things into traitlets: version numbering. After releasing 4.0, I set the version number to `4.1.0.dev`, because bugfix changes seemed more likely than breaking changes. We should keep in mind that we should branch before we merge the first change that needs to be 5.0.

This sounds to me like a bugfix change that belongs in 4.1, but are there any use cases it could break?

takluyver:
I've just created 4.1 and 5.0 milestones - please mark this with which one you think it fits in.

minrk:
@takluyver we should probably bring that up about the projects in general, but I'm on the same page: For our more slow-moving projects, master will typically be stable + 0.1. The faster moving guys (widgets, notebook) will probably hit stable + 1 more quickly.

As for this PR, I could imagine a contrived scenario that depended on the existing behavior, but I don't believe any exist at this point.

takluyver:
Great, thanks. I've put that one the Hackpad for tomorrow. Merging this.

takluyver:
For a fix like this, it would also be good to know how much pressure it adds to make a new release - does this block 4.0 releases of other projects, or can it sit and wait while other fixes accumulate?

minrk:
The effect that prompted this PR is that LazyConfigValues used in config files:

```python
c.NotebookApp.server_extensions.append('myextension')
```

end up being added ~three times, due to how we load config files. LazyConfigValues memoize their evaluated value, but when we are loading config, we are always copying the config object in the merge, so the memo is never re-used. This ends up with

```python
server_extensions = ['myextension', 'myextension', 'myextension']
```

I don't think this is a huge deal because:

0. in dict/set traits, it doesn't have any effect, only lists
0. In the majority of configurable list traits I'm aware of, adding an item multiple times doesn't actually have a negative effect, just a minor inefficiency
0. It's entirely possible that nobody but me actually uses this feature

I noticed this when I was working on the ipyparallel server extension, which I saw was getting loaded multiple times when I started adding extra debug logging.

takluyver:
Got it, thanks. I think it's fine for this to sit until there's a few more things to make 4.1, then.

takluyver:
This appears to have caused test failures for IPython in completion - the test is calling `completer.update_config()`, and it seems like the changes are not getting applied.

minrk:
fixed by #42

PR #44, opened 2015-07-08
Read-only trait attributes

SylvainCorlay:
Read-only attributes are something that was recently added in widgets for attributes that can only be set by the JavaScript side.

Another use-case that I encountered is for building a traitlets API to various sensors: forcibly setting the trait value does not make sense either, unless it is to indicate the value measured by the sensor. Hence it is probably useful outside of widgets.

To forcibly set a read-only trait attribute, use `HasTraits.set_trait`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -330,9 +330,10 @@ class TraitType(BaseDescriptor):
     metadata = {}
     default_value = Undefined
     allow_none = False
+    read_only = False
     info_text = 'any value'
 
-    def __init__(self, default_value=Undefined, allow_none=None, **metadata):
+    def __init__(self, default_value=Undefined, allow_none=None, read_only=None, **metadata):
         """Declare a traitlet.
 
         If *allow_none* is True, None is a valid value in addition to any
@@ -347,6 +348,8 @@ class TraitType(BaseDescriptor):
             self.default_value = default_value
         if allow_none is not None:
             self.allow_none = allow_none
+        if read_only is not None:
+            self.read_only = read_only
 
         if 'default' in metadata:
             # Warn the user that they probably meant default_value.
@@ -446,7 +449,7 @@ class TraitType(BaseDescriptor):
             else:
                 return value
 
-    def __set__(self, obj, value):
+    def set(self, obj, value):
         new_value = self._validate(obj, value)
         try:
             old_value = obj._trait_values[self.name]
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -644,6 +648,17 @@ def hold(*a):
                 for v in cache.values():
                     self._notify_trait(*v)
 
+    @contextlib.contextmanager
+    def _allow_write(self):
```
minrk: What's the utility of this context manager? If it's read-only, shouldn't it be read-only?

SylvainCorlay: - In widgets, the context manager is used to allow the widget machinery to write when receiving state from the js side. It is only read-only for the Python user.
- In the usecase with sensors, the thread polling the sensors will use that context manager to write sensor values.

takluyver: Having state on the trait for this seems like the wrong way to do this. E.g. any on_change callbacks for the traitlet will be able to modify it, because they'll be called while it's 'unlocked'. Likewise, if there's more than one thread that might try to change it, some changes that should raise an error could randomly succeed.

I'm not convinced this belongs in traitlets at all, but if it does, I think the design should be fundamentally different. For instance, a main interface using properties that is always read only, and a different 'back door' interface that can be used to update the stored value.

traitlets/traitlets.py
```diff
@@ -447,22 +447,25 @@ def __get__(self, obj, cls=None):
                 return value
 
     def __set__(self, obj, value):
-        new_value = self._validate(obj, value)
-        try:
-            old_value = obj._trait_values[self.name]
-        except KeyError:
-            old_value = self.default_value
+        if self._metadata.get('read_only') and obj._read_only_enabled:
```
takluyver: I'm also not terribly pleased that this requires an extra check every time any trait gets set. I know it's not much overhead, it just seems like bad design.

traitlets/traitlets.py
SylvainCorlay: I don't quite understand what you mean regarding the `on_change` handler. Why would it be unlocked?
Regarding the multi-threading that you are concerned about, the way I was thinking of doing this is by scheduling a thread-safe set of the value in the zmq event loop. what would be the issue then exactly?

takluyver: I'm sure you can come up with workarounds for both of the problems I described. My point is that doing this with a stateful 'is this currently unlocked?' API is probably not the best way to do this, because it will need workarounds for problems like these. I think more elegant, stateless designs are possible.

SylvainCorlay: That is right. It would make sense to have a different setter for read-only attributes. It might still require a boolean check in `__set__` though.

SylvainCorlay:
Rebased. Now, one can use `HasTraits.set_trait` to forcibly set read-only trait attribute.

PR #56, opened 2015-07-23
Promote help to be a constructor argument and a top-level member

jasongrout:
This continues the work in #53 (and includes it) to implement the decisions in #52.


jasongrout:
Rebased on master, now that #53 is merged.

Issue #57, opened 2015-08-02
import string bug (Py2 only)


rmorshea:
Matplotlib is supporting unicode literals rather than string literals and during my traitlets refactor I've come across this issue where `traitlets/traitlets/utils/importstring.py` will throw an error because the `fromlist` argument in `__import__` doesn't accept unicode:

```python
from __future__ import unicode_literals
from traitlets import Instance, HasTraits

class Artist(HasTraits):
    axes = Instance('matplotlib.axes.Axes')

a = Artist()
```

Even though `__import__` in Py3 accepts unicode, dealing with the unicode literals in Py2 would require that every class (that needs to be imported with an import string) instead read:

```python
class Artist(HasTraits):
    axes = Instance(str('matplotlib.axes.Axes'))
```

If there isn't a way around this problem it might be worth fixing.

PR #58, opened 2015-08-02
DEV: Handle both string types in import_item.

ssanderson:
- In PY2, coerce unicode to bytes.
- In PY3, coerce bytes to unicode.
- Add tests for the above behavior.

Fixes https://github.com/ipython/traitlets/issues/57

### DIFF
--- a/traitlets/utils/importstring.py
+++ b/traitlets/utils/importstring.py
@@ -2,9 +2,35 @@
 """
 A simple utility to import something by its string name.
 """
-
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
+from sys import version_info
+
+
+def _check_stringlike(s, types):
+    if not isinstance(s, types):
+        raise TypeError(
+            "import_item expected input of type "
+            "{0[0]} or {0[1]}, but got '{1}' instead.".format(
+                types, type(s),
+            )
+        )
+
+
+if version_info.major == 2:
+    # PY2 __import__ expects bytes
+    def _to_module_str(s):
+        _check_stringlike(s, (bytes, unicode))
+        if isinstance(s, unicode):
+            s = s.encode('utf-8')
+        return s
+elif version_info.major == 3:
+    # PY3 __import__ expects unicode
+    def _to_module_str(s):
+        _check_stringlike(s, (bytes, str))
+        if isinstance(s, bytes):
+            s = s.decode('utf-8')
+        return s
 
 
 def import_item(name):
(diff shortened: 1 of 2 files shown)


traitlets/utils/importstring.py
```diff
@@ -2,9 +2,35 @@
...
+
+
+if version_info.major == 2:
+    # PY2 __import__ expects bytes
+    def _to_module_str(s):
+        _check_stringlike(s, (bytes, unicode))
+        if isinstance(s, unicode):
+            s = s.encode('utf-8')
+        return s
+elif version_info.major == 3:
```
minrk: This should be a single ` if < 3...else` rather than elif.

traitlets/utils/tests/test_importstring.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestImportItem(TestCase):
+
+    def test_import_unicode(self):
+        self.assertIs(os, import_item(u'os'))
+        self.assertIs(os.path, import_item(u'os.path'))
+        self.assertIs(os.path.join, import_item(u'os.path.join'))
+
+    def test_import_bytes(self):
+        self.assertIs(os, import_item(b'os'))
```
minrk: `import_item(bytes)` *should* fail on Python 3. Is there any case where that makes sense as a thing to do?

traitlets/utils/importstring.py
```diff
@@ -23,7 +49,7 @@ def import_item(name):
     mod : module object
        The module that was imported.
     """
-    
+    name = _to_module_str(name)
```
minrk: I think we've already got the function we want as `py3compat.cast_bytes_py2`, which converts unicode to str on Python 2 and leaves unicode/str alone on Python 3. I don't think we want to be converting bytes to str on Python 3.

traitlets/utils/tests/test_importstring.py
ssanderson: It seems odd to me to accept `bytes` or `unicode` in PY2 but not in PY3.  Having the behavior of `import_item` be that it consistently accepts string-like types felt like the cleanest way to do this to me.

minrk: Python 2 has two string types. Python 3 only has one. `bytes` is not a string type on Python 3.

ssanderson: Fair enough.

minrk:
@ssanderson did you want to finish this one?

ssanderson:
Hey, sorry, I lost track of this.  Feel free to tie it off yourself if it's blocking something.  Otherwise I'll try to take finish it up this week.

minrk:
No problem. I can finish it up if you are busy.

PR #60, opened 2015-08-04
add notification muting alongside holding

rmorshea:
```python
from traitlets import HasTraits, Int

class Test(HasTraits):
    
    i = Int()
    
    def _i_changed(self, name, old, new):
        print(name+' : '+str(old)+' >>> '+str(new))
        
t = Test()

with t.hold_trait_notifications():
    t.i = 1
    print('held')

with t.mute_trait_notifications():
    t.i = 0
    print('muted')

t.i = 1
```

print results

```
held
i : 0 >>> 1
muted
i : 0 >>> 1
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -628,16 +628,15 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                 setattr(self, key, value)
 
     @contextlib.contextmanager
-    def hold_trait_notifications(self):
+    def mute_trait_notifications(self):
         """Context manager for bundling trait change notifications and cross
         validation.
-
         Use this when doing multiple trait assignments (init, config), to avoid
         race conditions in trait notifiers requesting other trait values.
         All trait notifications will fire after all values have been assigned.
         """
         if self._cross_validation_lock is True:
-            yield
+            yield {}
             return
         else:
             cache = {}
@@ -656,7 +655,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
             try:
                 self._notify_trait = hold
                 self._cross_validation_lock = True
-                yield
+                yield cache
                 for name in list(cache.keys()):
                     if hasattr(self, '_%s_validate' % name):
                         cross_validate = getattr(self, '_%s_validate' % name)
@@ -668,7 +667,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                         setattr(self, name, value[1])
                     else:
                         self._trait_values.pop(name)
-                cache = {}
+                cache.clear()
                 raise e
             finally:
                 self._notify_trait = _notify_trait
@@ -680,9 +679,21 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                     # (only used to preserve pickleability on Python < 3.4)
                     self.__dict__.pop('_notify_trait', None)
 
-                # trigger delayed notifications
-                for v in cache.values():
-                    self._notify_trait(*v)
+    @contextlib.contextmanager
+    def hold_trait_notifications(self):
+        """Context manager for bundling trait change notifications and cross
+        validation.
+        Use this when doing multiple trait assignments (init, config), to avoid
+        race conditions in trait notifiers requesting other trait values.
+        All trait notifications will fire after all values have been assigned.
+        """
+        try:
+            with self.mute_trait_notifications() as cache:
+                yield
+        finally:
+            self._cross_validation_lock = False
+            for v in cache.values():
+                self._notify_trait(*v)
 
     def _notify_trait(self, name, old_value, new_value):
 
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -691,7 +691,6 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
             with self.mute_trait_notifications() as cache:
                 yield
         finally:
-            self._cross_validation_lock = False
             for v in cache.values():
                 self._notify_trait(*v)
 
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1449,49 +1449,63 @@ def test_pickle_hastraits():
         nt.assert_equal(c2.j, c.j)
 
 
-def test_hold_trait_notifications():
-    changes = []
+class TestMutingAndHolding(TestCase):
 
-    class Test(HasTraits):
-        a = Integer(0)
-        b = Integer(0)
+    def _base_evaluation(self, method_name):
+        changes = []
 
-        def _a_changed(self, name, old, new):
-            changes.append((old, new))
+        class Test(HasTraits):
+            a = Integer(0)
+            b = Integer(0)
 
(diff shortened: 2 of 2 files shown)


traitlets/traitlets.py
```diff
@@ -680,9 +679,21 @@ def hold(*a):
...
+        validation.
+        Use this when doing multiple trait assignments (init, config), to avoid
+        race conditions in trait notifiers requesting other trait values.
+        All trait notifications will fire after all values have been assigned.
+        """
+        try:
+            with self.mute_trait_notifications() as cache:
+                yield
+        finally:
+            self._cross_validation_lock = False
```
minrk: I'm not sure about this cross-validation logic line. I wouldn't think it should be needed since it's already restored in the finally block of mute_trait_notifications

rmorshea: Yes, that's a mistake, thanks.

rmorshea:
@SylvainCorlay @jasongrout any suggestions or corrections regarding this?

traitlets/traitlets.py
```diff
@@ -628,16 +628,15 @@ def __init__(self, *args, **kw):
...
 
     @contextlib.contextmanager
-    def hold_trait_notifications(self):
+    def mute_trait_notifications(self):
         """Context manager for bundling trait change notifications and cross
         validation.
-
         Use this when doing multiple trait assignments (init, config), to avoid
         race conditions in trait notifiers requesting other trait values.
         All trait notifications will fire after all values have been assigned.
```
SylvainCorlay: The docstring should probably be modified to reflect what `mute_trait_notifications` does.

SylvainCorlay:
Tthis seems to work. What is a typical use case for muting trait notifications?

rmorshea:
It'd be for user defined defaults in which callbacks shouldn't be fired.
Or for matplotlib, the below might be useful to reduce the number of times
a callback for drawing is fired when many traits are updated.

```python
from traitlets import HasTraits, Int

class UseCaseParent(object):
    
    def __init__(self):
        self.trigger_count = 0
    
    def trigger(self):
        self.trigger_count += 1

class UseCaseChild(HasTraits):
    
    n = Int(0, trigger=True)
    m = Int(0, trigger=True)
    # more traits with trigger=True ...
    
    def __init__(self, parent, **kwargs):
        self.parent = parent
        names = self.trait_names(trigger=True)
        self.on_trait_change(self._trigger_parent,names)
    
    def update(self, **values):
        #with self.mute_trait_notifications():
            for k,v in values.items():
                setattr(self, k, v)
        #self.parent.trigger()
        
    def _trigger_parent(self, name):
        self.parent.trigger()
        
ucp = UseCaseParent()
ucc = UseCaseChild(ucp)
ucc.update(n=1, m=1)
```

rmorshea:
So I'm thinking that in addition to or instead of the context manager a `silent_set` method might be appropriate. Holding trait notifications only makes sense when you're assigning things in bulk, but there are a lot of cases with matplotlib where I need to be muting notifications for a single attribute and using the `with` syntax for that just seems a bit messy.

SylvainCorlay:
It the goal is to *reduce* the number of notification, using the `hold_trait_notifications` context manager is the right solution because it will send everything in batch and accumulate subsequent changes of the same attributes into a single notification.

jasongrout:
I'm also trying to understand exactly what is being proposed and the use cases for it.  It sounds like you might be shifting what you are proposing in the `silent_set` comment above.  I think you want a way to set defaults for a traitlet without triggering notifications?  Is that right?

rmorshea:
So having dug deeper into the issue I was encountering with matplotlib I think I need more than just silent setting. Essentially it the problem is that I'm trying to emulate a getters/setters with traitlets, and the fact that with getters/setters you still have access to the "private" value that's hidden behind them. It means that with setters/getters you can do four things:

+ normal setter method use
+ normal getter method use
+ underscore value setting
+ underscore value getting

The first capability can be emulated by traitlets, but the remainder, at the moment cannot.

SylvainCorlay:
I am not convinced of the need of a silent setter in general, and would be inclined to think that there would be a more natural solution to your problem.

rmorshea:
At the moment matplotlib is using all four of those capabilities, however I'm not sure whether all the use cases there are features and not bugs. In any case though, if that's not something matplotlib is willing to changed I'm fairly certain that I'd need to include `on_trait_set` and `on_trait_get` event handlers along with `silent_set` and `silent_get` to make the refactor happen.

rmorshea:
I guess on your end it's a question of whether you want to enable those things or not.

jasongrout:
The purpose of the traitlet observer pattern is to enable notifications.  I'm still trying to understand the use of using a traitlet when you don't want notifications.

rmorshea:
```python
from traitlets import HasTraits, Int

class Thing(HasTraits):

	n = Int()
	
	def __init__(self):
		self._reset_enabled = False
		
	def _n_changed(self):
		self._reset_enabled = True

def get_and_reset(obj):
	if obj._reset_enabled:
		return obj.n
	else:
		obj.n = 0
		return 0
		
def feature(obj1, obj2):
	return obj1.n + obj2.n
```

If I remember correctly this is similar to a piece of code which was throwing an error. Basically if the callback was fired it would changed the behavior of a handler that would receive the object later on. In the vast majority of cases one behavior was used while the other was ignored, however in the few cases where the other one was required, it was never enabled because the callback would always fire.

tacaswell:
@rmorshea Can you point me to the mpl use case for this?  I suspect that the fix might be on the mpl side.

rmorshea:
@tacaswell in tests/test_artist.py: `test_patch_transform_of_none`, `test_collection_transform_of_none`
which ultimately drills down to a statement in axes/_base.py:

```python
    # on line 782
    def _set_artist_props(self, a):
        """set the boilerplate props for artists added to axes"""
        a.set_figure(self.figure)
        if not a.is_transform_set():
            a.set_transform(self.transData)

        a.axes = self
```

where `update_from` on line 870 in artist.py can cause problems due to underscore assignments (at least that was my daignosis of the problem). This, in addition to various instances of underscore access instead of through the getter (hard to tell how many of those underscore accesses make a difference).

Even if the underscored access problem can be solved though, I think that "on set/get" notifiers are probably needed since  `_transformSet` (the underscore value retrieved by `is_transform_set`) needs to be updated regardless of a value change in `transform`, the getter may return a different value from `_transform`, and the `as_mpl_transform` api seems to require that the transform occur "on get" rather than "on set". Asside from these cases they would be useful as a way to do simple conversions on set like going from `'-'` to `MarkerStyle('-')` without explicitly defining that accommodation in a `MarkerStyle` TraitType or simply requiring that the input always be `MarkerStyle('-')`.

I should have a pull request by the end of the day that incorporates these notifiers (without muting).

rmorshea:
There's also a case in quiver.py and transforms.py (line 452 and 1569 respectively) where `transform` is used as an attribute wholey apart from `_transform` that was throwing me off for a while as well.

rmorshea:
get/set notifiers: #63

jasongrout:
@rmorshea, @tacaswell - it sounds like the appropriate fix is in matplotlib.  I'll close this in a bit unless you guys want to pursue this further.

PR #61, opened 2015-08-04
Deprecate on_trait_change and magic change handlers

SylvainCorlay:
####  Deprecation of `on_trait_change` and the signature of change handlers

As of now, this PR deprecates `on_trait_change` and the magic `_*foo*_changed` handlers, to the benefit of new `observe` and `unobserve` method. The signature of the handlers passed to `observe` is either
 - no argument
 - a single `dict` argument containing  'name', 'old', 'new' and 'owner', the HasTraits instance emitting the event.

See the examples below.

```Python
from __future__ import print_function
from traitlets import *

class Foo(HasTraits):
    bar = Int()
    
    def test_bar(self):
        print('foobar')
    
foo = Foo()    

def cb(change):
    print(change)
    
foo.observe(cb)
foo.bar = 3
```

#### Implementation of an `observe` decorator like in [Atom](https://github.com/nucleic/atom):
On the other side, atom implements a `observe` decorator 

```Python
class Person(Atom):
    age = Int()

    @observe('age')
    def debug_print(self, change):
        print(change['object'].last_name)
```

The problem is that the registration of the callback requires access `self` attribute, which is not available when the decorator is called. The way it works in atom is that the `@observe('age')` decorator replaces the `debug_print` method with a placeholder object `ObserveHandler` holding the method to be registered, which is to be picked up and unwrapped by the `AtomMeta` metaclass when instantiating the instance of (the counterpart to) HasTraits.

Rather than augmenting the metaclass, I could achieve the same result by making our `ObserveHandler` inherit from `BaseDescriptor`, and implement the custom logic in `instance_init`.

#### Example with traitlets:

```Python
class Foo(HasTraits):
    bar = Int()
    
    @observe('bar')
    def test(self):
        print('foobar')

foo = Foo()
foo.bar = 2
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -552,8 +552,12 @@ class TraitType(BaseDescriptor):
 # The HasTraits implementation
 #-----------------------------------------------------------------------------
 
-class _CbWrapper(object):
+class _CallbackWrapper(object):
+    """An object adapting a on_trait_change callback into an observe callback.
     
+    The comparison operator __eq__ is implemented to enable removal of wrapped
+    callbacks.
+    """ 
 
     def __init__(self, cb):
         if callable(cb):
@@ -568,13 +572,13 @@ class _CbWrapper(object):
 
     def __eq__(self, other):
         # The wrapper is equal to the wrapped element
-        if isinstance(other, _CbWrapper):
+        if isinstance(other, _CallbackWrapper):
             return self.cb == other.cb
         else:
             return self.cb == other
 
     def __call__(self, change):
-        # The wrapper is callabled
+        # The wrapper is callable
         if self.nargs == 0:
             self.cb()
         elif self.nargs == 1:
@@ -586,11 +590,11 @@ class _CbWrapper(object):
         elif self.nargs == 4:
             self.cb(change['name'], change['old'], change['new'], change['object'])
 
-def _cb_wrapper(cb):
-    if isinstance(cb, _CbWrapper):
+def _callback_wrapper(cb):
+    if isinstance(cb, _CallbackWrapper):
         return cb
     else:
-        return _CbWrapper(cb)
+        return _CallbackWrapper(cb)
 
 
 
@@ -635,6 +639,13 @@ class MetaHasTraits(type):
 
 
 def observe(*names):
+    """ A decorator which can be used to observe members on a class.
+
+    Parameters
+    ----------
+    *names
+        The str names of the attributes to observe on the object.
+    """
     return ObserveHandler(names)
 
 
@@ -767,7 +778,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         else:
             warn("_[traitname]_changed change handlers are deprecated: use observe instead", 
                  DeprecationWarning, stacklevel=2)
-            callables.append(_cb_wrapper(cb))
+            callables.append(_callback_wrapper(cb))
 
         # Call them all now
         for c in callables:
@@ -776,8 +787,8 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
                 # Bound methods have an additional 'self' argument.
                 offset = -1 if isinstance(c, types.MethodType) else 0
 
-                if isinstance(c, _CbWrapper):
-                    # _CbWrappers are not compatible with getargspec and have one argument
+                if isinstance(c, _CallbackWrapper):
+                    # _CallbackWrappers are not compatible with getargspec and have one argument
                     nargs = 1
                 else:
                     nargs = len(getargspec(c)[0]) + offset
@@ -846,7 +857,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         """
         warn("on_trait_change is deprecated: use observe instead",
              DeprecationWarning, stacklevel=2)
-        self.observe(_cb_wrapper(handler), name=name, remove=remove)
+        self.observe(_callback_wrapper(handler), name=name, remove=remove)
 
     def observe(self, handler, name=None, remove=False):
         """Setup a handler to be called when a trait changes.
@@ -858,18 +869,18 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         handler : callable, None
             A callable that is called when a trait changes.  Its
             signature can be handler() or handler(change), where change is a
-            dictionary with the following optional keys:
+            dictionary with the following keys:
                 - object : the HasTraits instance
                 - old : the old value of the modified trait attribute
                 - new : the new value of the modified trait attribute
-                - name : the name ofthe modified trait attribute.
+                - name : the name of the modified trait attribute.
         name : list, str, None
             If None, the handler will apply to all traits.  If a list
             of str, handler will apply to all names in the list.  If a
             str, the handler will apply just to that name.
         remove : bool
             If False (the default), then install the handler.  If True
-            then unintall it. If `remove` is True and `handler` is not
+            then uninstall it. If `remove` is True and `handler` is not
             specified, all change handlers for the specified name are
             uninstalled.
         """
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -209,8 +209,8 @@ class link(object):
         try:
             setattr(target[0], target[1], getattr(source[0], source[1]))
         finally:
-            source[0].on_trait_change(self._update_target, source[1])
-            target[0].on_trait_change(self._update_source, target[1])
+            source[0].observe(self._update_target, name=source[1])
+            target[0].observe(self._update_source, name=target[1])
 
     @contextlib.contextmanager
     def _busy_updating(self):
@@ -220,21 +220,21 @@ class link(object):
         finally:
             self.updating = False
 
-    def _update_target(self, name, old, new):
+    def _update_target(self, change):
         if self.updating:
             return
         with self._busy_updating():
-            setattr(self.target[0], self.target[1], new)
+            setattr(self.target[0], self.target[1], change['new'])
 
-    def _update_source(self, name, old, new):
+    def _update_source(self, change):
         if self.updating:
             return
         with self._busy_updating():
-            setattr(self.source[0], self.source[1], new)
+            setattr(self.source[0], self.source[1], change['new'])
 
     def unlink(self):
-        self.source[0].on_trait_change(self._update_target, self.source[1], remove=True)
-        self.target[0].on_trait_change(self._update_source, self.target[1], remove=True)
+        self.source[0].observe(self._update_target, name=self.source[1], remove=True)
+        self.target[0].observe(self._update_source, name=self.target[1], remove=True)
         self.source, self.target = None, None
 
 
@@ -261,7 +261,7 @@ class directional_link(object):
         try:
             setattr(target[0], target[1], getattr(source[0], source[1]))
         finally:
-            self.source[0].on_trait_change(self._update, self.source[1])
+            self.source[0].observe(self._update, name=self.source[1])
 
     @contextlib.contextmanager
     def _busy_updating(self):
@@ -271,14 +271,14 @@ class directional_link(object):
         finally:
             self.updating = False
 
-    def _update(self, name, old, new):
+    def _update(self, change):
         if self.updating:
             return
         with self._busy_updating():
-            setattr(self.target[0], self.target[1], new)
+            setattr(self.target[0], self.target[1], change['new'])
 
     def unlink(self):
-        self.source[0].on_trait_change(self._update, self.source[1], remove=True)
+        self.source[0].observe(self._update, name=self.source[1], remove=True)
         self.source, self.target = None, None
 
 dlink = directional_link
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -634,6 +634,28 @@ class MetaHasTraits(type):
         super(MetaHasTraits, cls).__init__(name, bases, classdict)
 
 
+def observe(*names):
+    return ObserveHandler(names)
+
+
+class ObserveHandler(BaseDescriptor):
+
+    def __init__(self, names=None):
+        if names is None:
+            self.names=[None]
+        else:
+            self.names = names
+
+    def __call__(self, func):
+        self.func = func
+        return self
+
+    def instance_init(self, inst):
+        setattr(inst, self.name, self.func)
+        for name in self.names:
+            inst.observe(self.func, name=name)
+
+
 class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
     """The base class for all classes that have traitlets.
     """
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -233,8 +233,8 @@ class link(object):
             setattr(self.source[0], self.source[1], change['new'])
 
     def unlink(self):
-        self.source[0].observe(self._update_target, name=self.source[1], remove=True)
-        self.target[0].observe(self._update_source, name=self.target[1], remove=True)
+        self.source[0].unobserve(self._update_target, name=self.source[1])
+        self.target[0].unobserve(self._update_source, name=self.target[1])
         self.source, self.target = None, None
 
 
@@ -278,7 +278,7 @@ class directional_link(object):
             setattr(self.target[0], self.target[1], change['new'])
 
     def unlink(self):
-        self.source[0].observe(self._update, name=self.source[1], remove=True)
+        self.source[0].unobserve(self._update, name=self.source[1])
         self.source, self.target = None, None
 
 dlink = directional_link
@@ -857,16 +857,19 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
         """
         warn("on_trait_change is deprecated: use observe instead",
              DeprecationWarning, stacklevel=2)
-        self.observe(_callback_wrapper(handler), name=name, remove=remove)
+        if remove:
+            self.unobserve(_callback_wrapper(handler), name=name)
+        else:
+            self.observe(_callback_wrapper(handler), name=name)
 
-    def observe(self, handler, name=None, remove=False):
+    def observe(self, handler, name=None):
         """Setup a handler to be called when a trait changes.
 
         This is used to setup dynamic notifications of trait changes.
 
         Parameters
         ----------
-        handler : callable, None
+        handler : callable
             A callable that is called when a trait changes.  Its
             signature can be handler() or handler(change), where change is a
             dictionary with the following keys:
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
To handle the removal of the wrapped callbacks, I will need to write a bit more code. Hence the test failure. Looking into it.

SylvainCorlay:
#### Back to the `observe` decorator like in [Atom](https://github.com/nucleic/atom):

I can actually make the wrapper object inherit from the base descriptor. I think it is going to be very clean...

traitlets/traitlets.py
```diff
@@ -771,6 +829,33 @@ def on_trait_change(self, handler, name=None, remove=False):
...
+        Parameters
+        ----------
+        handler : callable
+            A callable that is called when a trait changes.  Its
+            signature can be handler() or handler(change), where change is a
+            dictionary with the following optional keys:
+                - object : the HasTraits instance
+                - old : the old value of the modified trait attribute
+                - new : the new value of the modified trait attribute
+                - name : the name ofthe modified trait attribute.
```
blink1073: Missing space

traitlets/traitlets.py
```diff
@@ -771,6 +829,33 @@ def on_trait_change(self, handler, name=None, remove=False):
...
+        name : list, str, None
+            If None, the handler will apply to all traits.  If a list
+            of str, handler will apply to all names in the list.  If a
+            str, the handler will apply just to that name.
+        remove : bool
+            If False (the default), then install the handler.  If True
+            then unintall it.
```
blink1073: Spelling

SylvainCorlay:
@ellisonbg @sccolbert as per earlier discussion.

blink1073:
@SylvainCorlay, tested both constructs, :+1:

blink1073:
The old constructs worked too, but I didn't get any warnings when run in an IPython console.

SylvainCorlay:
Hum, I am seeing the warnings in the notebook.

blink1073:
For whatever reason my session is not printing `DeprecationWarnings` in the console or the notebook, I must have a global setting somewhere...

jasongrout:
By default, deprecation warnings are not printed in python.  You have to either manually set the warnings filter to allow them, or start python with a special switch.

traitlets/traitlets.py
```diff
@@ -547,6 +547,51 @@ def tag(self, **metadata):
 # The HasTraits implementation
 #-----------------------------------------------------------------------------
 
+class _CbWrapper(object):
```
jasongrout: Can we make this `_CallbackWrapper`?  It feels more pythonic than using abbreviations.

traitlets/traitlets.py
```diff
@@ -547,6 +547,51 @@ def tag(self, **metadata):
...
+
+    def __eq__(self, other):
+        # The wrapper is equal to the wrapped element
+        if isinstance(other, _CbWrapper):
+            return self.cb == other.cb
+        else:
+            return self.cb == other
+
+    def __call__(self, change):
+        # The wrapper is callabled
```
jasongrout: `callabled` -> `callable`

traitlets/traitlets.py
```diff
@@ -547,6 +547,51 @@ def tag(self, **metadata):
...
+        elif self.nargs == 1:
+            self.cb(change['name'])
+        elif self.nargs == 2:
+            self.cb(change['name'], change['new'])
+        elif self.nargs == 3:
+            self.cb(change['name'], change['old'], change['new'])
+        elif self.nargs == 4:
+            self.cb(change['name'], change['old'], change['new'], change['object'])
+
+def _cb_wrapper(cb):
```
jasongrout: I like `_callback_wrapper`...

traitlets/traitlets.py
```diff
@@ -771,6 +840,33 @@ def on_trait_change(self, handler, name=None, remove=False):
...
+        """Setup a handler to be called when a trait changes.
+
+        This is used to setup dynamic notifications of trait changes.
+
+        Parameters
+        ----------
+        handler : callable
+            A callable that is called when a trait changes.  Its
+            signature can be handler() or handler(change), where change is a
+            dictionary with the following optional keys:
```
jasongrout: Are the keys really optional?

jasongrout:
@minrk, @ellisonbg - when you guys have a minute, it'd be great to hear your comments.

SylvainCorlay:
Let me know what you think.
- While keeping the `@observe` decorator, we can rename the `.observe` *method* into `.on_change`.
- Another idea would be to remove the `remove=False/True` argument from the signature, and use a different method to unregister change hanldlers, like `off_change`.

ellisonbg:
OK I see this one and will have a look...

On Sun, Aug 9, 2015 at 1:49 PM, Sylvain Corlay <notifications@github.com>
wrote:

> Let me know what you think.
>
>    - While keeping the @observe decorator, we can rename the .observe
>    *method* into .on_change.
>    - Another idea would be to remove the remove=False/True argument from
>    the signature, and use a different method to unregister change hanldlers,
>    like off_change.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-129237052>.
>



-- 
Brian E. Granger

ellisonbg:
I think the name of the key in the change dict of `object` is too vague. What about `source` or `parent`?

SylvainCorlay:
Sounds good. I picked `object` like in Atom, but did not quite use the same names for the other keys. I like `source`.

minrk:
source is probably better - parent is already used in config, and would get confusing.

sccolbert:
perhaps `owner`?

minrk:
owner makes sense to me

SylvainCorlay:
Any take on the other two questions
 - while keeping the `observe` *decorator*, the `observe` *function* could be named `on_change.`
 - use a separate function for deregistration? `off_change` ?

ellisonbg:
I like the symmetry of having both called `observe`

For the counterpart, maybe `unobserve`?

On Sat, Aug 22, 2015 at 12:05 AM, Sylvain Corlay <notifications@github.com>
wrote:

> Any take on the other two questions
>
>    - while keeping the observe *decorator*, the observe *function* could
>    be named on_change.
>    - use a separate function for deregistration? off_change ?
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-133647861>.
>



-- 
Brian E. Granger

SylvainCorlay:
`register/unregister`?

blink1073:
`watch/unwatch`?

sccolbert:
Atom uses `observe/unobserve`

SylvainCorlay:
Now using `observe/unobserve`. Remains the naming decision between `'object'` and `'owner'`. Otherwise, it is ready for review.

blink1073:
I vote for `owner` because `object` is a builtin.

SylvainCorlay:
(Good point, although it is just a string)

SylvainCorlay:
@minrk, @ellisonbg do you have other things you would like to see in this one?

minrk:
Other than picking 'owner' over 'object' (I don't have a real preference - object has the potential namespace issue, but I think it's actually a clearer key than owner), I think this is in good shape.

SylvainCorlay:
ok, I don't have a preference of `owner` or `object`. I picked `object` because of Atom - the APIs are really close now.

blink1073:
Long live `object` (your point about it being a string escaped me this morning).

ellisonbg:
I think I prefer source and owner over object, with owner being my fav.

On Tue, Aug 25, 2015 at 1:32 PM, Steven Silvester <notifications@github.com>
wrote:

> Long live object (your point about it being a string escaped me this
> morning).
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-134731560>.
>



-- 
Brian E. Granger

jasongrout:
+1 to `source` over `object` or `owner`.

SylvainCorlay:
I slightly lean towards `object`.

ellisonbg:
I think we have a perfectly uniform distribution over names (just an aside,
I hate naming :) - do we want to get together on a call and try to push
through to a decision on this? Other option is to BDFL it up to @fperez

On Tue, Aug 25, 2015 at 3:21 PM, Sylvain Corlay <notifications@github.com>
wrote:

> I slightly lean towards object.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-134758739>.
>



-- 
Brian E. Granger

fperez:
Better if you guys can make a decision, but obviously ping me if a knot needs to be cut in half :)

tacaswell:
From the peanut gallery, I would go with `owner` over `source` over `object`.  

Object is way too overloaded (is it the traitlet object or the configurable object? in addition to the brain-namespace collision with built-in `object`) and source sounds like it is providing the new values, not the object that the traitlet object is attached to.

SylvainCorlay:
PS: there cannot be a collision with the build-in `object` since it is a string, used as a key in a dictionary.

ellisonbg:
From my quick counting, I think we have 3 votes for owner, 2 for object and
1 for source. Any other one want to weigh in on this?

On Wed, Aug 26, 2015 at 6:00 AM, Sylvain Corlay <notifications@github.com>
wrote:

> PS: there cannot be a collision with the build-in object since it is a
> string, used as a key in a dictionary.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-134999806>.
>



-- 
Brian E. Granger

fperez:
Non-BDFL, pure 'just another developer' vote, count me in for 'owner'.

rmorshea:
If you kept with the naming scheme used in `link` it would probably be 'source', but in the context of change handlers I think 'owner' is better. Something like 'initiator' might make sense for both though.

ellisonbg:
Are folks OK moving forward with owner? That is the closest thing to a
consensus we have at this point. Otherwise we should probably hop on a call
and try to talk through things though.

On Wed, Aug 26, 2015 at 3:32 PM, Ryan Morshead <notifications@github.com>
wrote:

> If you kept with the naming scheme used in link it would probably be
> 'source', but in the context of change handlers I think 'owner' is better.
> Something like 'initiator' might make sense for both though.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135196545>.
>



-- 
Brian E. Granger

sccolbert:
I'm perfectly happy with `owner`.

SylvainCorlay:
I will push the change whenever I get close to a keyboard.

ellisonbg:
OK great. Looking forward to seeing this land - I think matplotlib will
start to use this API almost instantly!

On Wed, Aug 26, 2015 at 5:05 PM, Sylvain Corlay <notifications@github.com>
wrote:

> I will push the change whenever I get close to a keyboard.
> On Aug 26, 2015 7:04 PM, "S. Chris Colbert" <notifications@github.com>
> wrote:
>
> > I'm perfectly happy with owner.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ipython/traitlets/pull/61#issuecomment-135201073>.
>
> >
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135211389>.
>



-- 
Brian E. Granger

SylvainCorlay:
Done.

SylvainCorlay:
I just thought of `emitter` by the way. What do you guys think of it?

ellisonbg:
I like it almost as much as owner....but am tired of naming things :)

On Thu, Aug 27, 2015 at 8:06 AM, Sylvain Corlay <notifications@github.com>
wrote:

> I just thought of emitter by the way. What do you guys think of it?
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135462888>.
>



-- 
Brian E. Granger

blink1073:
The benefit of `emitter` is that `phosphor-signaling` calls it the same thing: http://phosphorjs.github.io/phosphor-signaling/api/globals.html#emitter

SylvainCorlay:
Yes, it would make sense.

sccolbert:
Yeah, I was avoiding mentioning `emitter`, because I wanted to avoid the discussion about the subtle differences between signals and traits. Where signals are state-less, traits have state, so a richer payload is warranted IMO. If you're already creating a dict for the payload, might as well include the owner. 

As long as you stay `<=` 5 items in the dict, you don't pay any extra allocation costs and the items will fit in the inlined small table. Once you hit 6 items, you'll get a rehash and a secondary heap allocation, which can have a big impact on dispatch performance:

```python
>>> import sys
>>> a = dict(zip(range(4), range(4)))
>>> b = dict(zip(range(5), range(5)))
>>> c = dict(zip(range(6), range(6)))
>>> sys.getsizeof(a)
280
>>> sys.getsizeof(b)
280
>>> sys.getsizeof(c)
1048
```

ellisonbg:
Good point!

On Thu, Aug 27, 2015 at 8:43 AM, Sylvain Corlay <notifications@github.com>
wrote:

> Yes, it would make sense.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135473093>.
>



-- 
Brian E. Granger

SylvainCorlay:
So are we moving to emitter?

ellisonbg:
Hmm, yeah, I can see how "emitter" has a "signal oriented" connotation
rather than the stateful model of traitlets which I think it better
described by "owner" (IOW, the "owner" owns that bit of state).

On Thu, Aug 27, 2015 at 8:49 AM, Sylvain Corlay <notifications@github.com>
wrote:

> So are we moving to emitter?
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135474354>.
>



-- 
Brian E. Granger

SylvainCorlay:
In the signaling PR, there are two bridges between traits and signaling:

- a widget attribute can be used as a slot (an `Int()` widget attribute can receive value from a signal having and `Int()` signature.)
- every widget has a global `state_changed` signal.

In the latter case especially, mixing two systems where one uses `owner` and the other one uses `emitter` may seem cumbersome.

jasongrout:
+1 on emitter because it's more like 'source' :)

SylvainCorlay:
Pushed `emitter` instead of `owner`. Let's wait for @minrk to merge or ask to change back to 'owner' :)

ellisonbg:
To clarify @minrk I think you are the right person to make this call. In
summary, there was a mild consensus around "owner" but "emitter" has
emerged as a late and popular option.

I am personally 50/50 on owner versus emitter.

On Thu, Aug 27, 2015 at 11:10 AM, Sylvain Corlay <notifications@github.com>
wrote:

> Pushed emitter instead of owner. Let's wait for @minrk
> <https://github.com/minrk> to merge or ask to change back to 'owner' :)
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135510250>.
>



-- 
Brian E. Granger

rmorshea:
'source' sounded generic and ambiguous, 'emitter' seems better.

minrk:
Owner also seems slightly preferable to emitter, but both seem just fine. I'm okay with letting the current state of this PR make the choice.

ellisonbg:
We are getting @jdfreder to weigh in right now...

On Thu, Aug 27, 2015 at 11:54 AM, Min RK <notifications@github.com> wrote:

> Owner also seems slightly preferable to emitter, but both seem just fine.
> I'm okay with letting the current state of this PR make the choice.
>
> —
> <https://github.com/ipython/traitlets/pull/61#issuecomment-135522325>.
>



-- 
Brian E. Granger

ellisonbg:
@jdfreder voted for emitter. I am still 50/50.

On Thu, Aug 27, 2015 at 11:58 AM, Brian Granger <ellisonbg@gmail.com> wrote:

> We are getting @jdfreder to weigh in right now...
>
> On Thu, Aug 27, 2015 at 11:54 AM, Min RK <notifications@github.com> wrote:
>
>> Owner also seems slightly preferable to emitter, but both seem just fine.
>> I'm okay with letting the current state of this PR make the choice.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/ipython/traitlets/pull/61#issuecomment-135522325>.
>>
>
>
>
> --
> Brian E. Granger
> Associate Professor of Physics and Data Science
> Cal Poly State University, San Luis Obispo
> @ellisonbg on Twitter and GitHub
> bgranger@calpoly.edu and ellisonbg@gmail.com
>



-- 
Brian E. Granger

SylvainCorlay:
@ellisonbg I updated the PR as per what we agreed on during the call.

ellisonbg:
Just to update the community - on a phone call we reached consensus on using owner for this as this usage of more about a property pattern than a signaling one. If we ever have a signally abstraction we all agree that emitter would be good in that context.

ellisonbg:
And great work!

traitlets/traitlets.py
```diff
@@ -593,6 +637,35 @@ class dict to the newly created class ``cls``.
...
+        if names is None:
+            self.names=[None]
+        else:
+            self.names = names
+
+    def __call__(self, func):
+        self.func = func
+        return self
+
+    def instance_init(self, inst):
```
rmorshea: It turns out that decorators prevent functions from being registered as methods at run time. `instance_init` should instead read:

```python
        self.func = types.MethodType(self.func, inst)
        for name in self.names:
            inst.observe(self.func, name= name)

File: traitlets/traitlets.py [hunk 8]
```diff
@@ -593,6 +637,35 @@ class dict to the newly created class ``cls``.
...
+        setattr(inst, self.name, self.func)
```
sccolbert: What is the intent of this `setattr` call? It's seems be assigning the bound method to the instance which a) wastes space, and b) causes a circular reference.

It seems like omitting the `setattr` line should work?

rmorshea:
@sccolbert I think the problem with that is that `__call__` returns `self` rather than the function.

traitlets/traitlets.py
SylvainCorlay: Hum then you could not call the method directly right?

SylvainCorlay:
@rmorshea you are right. We should make it a bound method.

rmorshea:
Sure

sccolbert:
Yep, I didn't see that `__call__` returns self.

No worries though, simply reimplement `__get__` on the descriptor to return a new bound method when accessed:

```python
def __get__(self, inst, cls):
    if inst is None:
        return self
    return MethodType(self.func, inst)
```

The way it currently sits, it allocates a new bound method for *each* function for *each* instance. It then also consumes *more* space by storing the method on the instance, plus circular ref, etc. That's pretty wasteful and can cause some undue pain when people start creating lots of objects.

rmorshea:
I think that's a cleaner way to do it. What's the "circular ref" though, wouldn't the ObserveHander instance be garbage collected?

sccolbert:
The circular ref would be caused by the bound method holding a reference to the instance, and instance holding a reference to the bound method. It's not the end of the world, and the garbage collector would probably get to it eventually (unless the instance has a `__del__` method). That said, circular refs should be avoided when possible because they delay the cleanup of objects (forcing it upon the GC) and are generally a sign of a code-smell.

rmorshea:
Gotcha, wasn't looking for it in the right place.

rmorshea:
New PR for that?

PR #62, opened 2015-08-07
ensure stable order of config output

minrk:
config file output was subject to dict ordering, resulting in reordering config file content.

closes ipython/ipython#8702


Carreau:
Need rebase, sorry about that.

minrk:
rebased

PR #63, opened 2015-08-10
add trait setters/getters

rmorshea:
example:

```python
from traitlets import HasTraits, Int

class Test(HasTraits):
    
    n = Int()
    
    def _n_setter(self, value):
        return int(value)
        
    def _n_getter(self, value):
        return float(value)

t = Test()
t.n = 1.2

print(t.n)
```

print result: `1.0`

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -507,6 +507,61 @@ class TestHasTraits(TestCase):
         # should raise TypeError if no positional arg given
         self.assertRaises(TypeError, A)
 
+    def test_getset_modifiers(self):
+
+        def setUp(self):
+            self.setter_trigger = False
+            self.getter_trigger = False
+
+        class A(HasTraits):
+
+            a = Int(0)
+            b = Int(0)
+            c = Int(1)
+
+            def _setter(self, value, trait):
+                self.setter_trigger = True
+                return int(value)
+            def _getter(self, value, trait):
+                self.getter_trigger = True
+                return value+self.c
+
+            def _a_setter(self, value, trait):
+                return self._setter(value, trait)
+            def _a_getter(self, value, trait):
+                return self._getter(value, trait)
+            def _c_setter(self, value, trait):
+                return float(value)
+        
+        a = A()
+        a.a = '1'
+
+        self.assertEqual(a._trait_values['a'], 1)
+        self.assertEqual(a.a, 2)
+        self.assertEqual(a.setter_trigger, True)
+        self.assertEqual(a.getter_trigger, True)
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
You could also use custom validators to achieve the same thing.

```Python
from traitlets import HasTraits, Float
import math

class Test(HasTraits):

    n = Float()

    def _n_validate(self, value, trait):
        return math.floor(value)

t = Test()
t.n = 1.2

print(t.n)
```

SylvainCorlay:
Actually, custom validators might be what you where looking for since the beginning. You can coerce/validate value using other attributes' values before trait notifications fire. Hence it is rather flexible.

rmorshea:
That example was more a way to get the idea across, so I guess a more exhaustive demo for the setter would be.

```python
from traitlets import HasTraits, Float

class Test(HasTraits):

    n = Float()

    def _n_setter(self, value):
        return float(value)

t = Test()
t.n = '1.2'

print(t.n)
```

rmorshea:
the main issue with cross validation is that it occurs after the main validation does

SylvainCorlay:
You can also create a CFloat trait type that cast passed values to Float... For example I often use a CInstance trai type that casts to the specified type...

rmorshea:
So this gets into some of the nuance I think. The use case I posted on the #60 involved an attribute `_transformSet` whose value needed to be updated regardless of whether the `_transform` had changed. This would be fine except for the fact that using `with self.hold_trait_notifications():` would prevent that call from being fired if I were to use `_name_validate` as a way to trigger that update.

SylvainCorlay:
`self.hold_trait_notifications()` will run the cross validation when the context manager is released.

SylvainCorlay:
Since I don't know the specifics of the matplotlib use case, can you post a more complete toy example of what you want to achieve. I still think that it is likely that you can already do everything you need with custom trait types and/or custom validation.

rmorshea:
I forgot about that. I think the setters may be covered by `_name_validate`. The only differences being that the setter here would basically allow for trivial conversions without inheritance and that I (maybe presumptuously) included an `on_trait_set` function. The only way that I can see to cover the getter though would be to have linked change notifiers that span both instances and attributes. As cool as that would be, I think it would be a bit hairy to actually implement.

rmorshea:
maybe that's not as bad as I'm thinking it would be though...

SylvainCorlay:
Should we close this one?

rmorshea:
I think so.

rmorshea:
the getters might resurface though if underscore access is a feature in matplotlib rather than a bug

Issue #64, opened 2015-08-13
Using repr of class name


jhamrick:
It appears that when a default config file is generated (or you just generate command line help), that class names are now represented no longer as just dotted strings, but of the python repr, e.g.:

```
# c.FormgradeApp.authenticator_class = <class 'nbgrader.auth.noauth.NoAuth'>
```

Given that this is not valid python syntax, were this line to be uncommented, the config file would not be able to be loaded.

PR #66, opened 2015-08-14
Set default value for containers from dynamic default

### LINKED ISSUE #65: Lists become 'traitlets.Undefined' in config files
If you generate a default config file (or print command line help), it appears that lists get converted to `traitlets.Undefined`, e.g.:

```
# c.NbConvertApp.notebooks = traitlets.Undefined
```

If this line is uncommented, then the config file cannot be loaded because `traitlets` is undefined. Previously, the config files used to display the actual list.

jhamrick:
Fixes #65 

I don't really know very much about the inner workings of traitlets, so this might not be the right approach. If that's the case, I'm happy to change this if someone can point me in the right direction.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1591,6 +1591,8 @@ class Container(Instance):
 
         super(Container,self).__init__(klass=self.klass, args=args, **metadata)
 
+        self.default_value = self.make_dynamic_default()
+
     def element_error(self, obj, element, validator):
         e = "Element of the '%s' trait of %s instance must be %s, but a value of %s was specified." \
             % (self.name, class_of(obj), validator.info(), repr_type(element))


minrk:
I think the fix might be to check for `Undefined` in `Configurable.class_config_section`, rather than generate the default on init, which is actually avoided on purpose now.

minrk:
And as you mentioned in #65, something similar is likely wanted in `class_get_trait_help` as well.

jhamrick:
Actually, looking into #64 as well, I'm wondering if a slightly more general approach would be more appropriate -- because I can see the case where there will be custom traitlets whose default values don't necessarily have an appropriate repr, and we definitely wouldn't want to special case those.

What about a solution like defining a method in `TraitType` called something like `default_value_repr`, which defaults to just `repr(self.default_value)`, but can be overwridden by traits that need to do so. So, for `Container` traits, they can do `repr(self.make_dynamic_default())`, and for `Type` traits, they can make the appropriate dotted object name instead of `<class '...'>`.

SylvainCorlay:
The problem is that these functions in the `config` directly access the `default_value` attribute of the trait type. 

This may suggest that `default_value` should be considered a private argument, and these functions should thn call an getter, that generates the default value if it is undefined.

Besides, some trait types like `Type` don't have any valid default value, and making this explicit with by throwing an exception in the getter would make things more explicit.

jhamrick:
Yes, that's similar to what I was suggesting, though for `Type` I think you do legitimately want to have default values -- you should be able to specify a default class name. But in that case, you don't want it to be formatted like `<class '...'>` in the config file, you just want the dotted object name. So just having a getter for the default value wouldn't completely solve both #64 and #65.

I think there used to be a getter for `default_value` but it got removed (I don't know why).

SylvainCorlay:
Yes, we are on the same page.

I was just thinking of the special case where someone declares something like
```Python
class Foo(HasTraits):
    bar = Type()
```
Then one *must* specify a type at construction of instances of `Foo`, like
```Python
foo = Foo(bar=float)  # ok
foo = Foo()           # raises a TraitError
```
In this case, the getter for the default value should raise.

But indeed, must of the time, a default type would be specified.`bar=Type(float)`

jhamrick:
Ahh, I see what you mean -- and yes, I agree. I guess there's a distinction between "default" in the sense that the programmer has not supplied a default, and "default" in the sense that the programmer has a given a default but the user has not overwritten it.

minrk:
@jhamrick if you want to add an overridable getter for the default value repr, that seems like a reasonable solution to me.

jhamrick:
Closing in favor of #68

PR #67, opened 2015-08-19
fix config file generation

Carreau:
My fault, merge #62 too fast.


Carreau:
Need test.

minrk:
My fault for the bad rebase. Sorry.

Carreau:
Blame shared then. Should have catch that up.

PR #68, opened 2015-08-19
Add default_value_repr method

### LINKED ISSUE #64: Using repr of class name

### LINKED ISSUE #65: Lists become 'traitlets.Undefined' in config files

jhamrick:
Fixes #65 and fixes #64 

See also #66 

This adds a `default_value_repr` method that by default just returns the repr of `default_value`, but which can be overridden for traits that need special casing (e.g. dotted object names, or containers).

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -209,7 +209,7 @@ class Configurable(HasTraits):
             lines.append(indent('Current: %r' % getattr(inst, trait.name), 4))
         else:
             try:
-                dvr = repr(trait.default_value)
+                dvr = trait.default_value_repr()
             except Exception:
                 dvr = None # ignore defaults we can't construct
             if dvr is not None:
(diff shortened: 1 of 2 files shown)


minrk:
Somehow this appears to need a rebase.

minrk:
:+1: once it's mergeable. Thanks, @jhamrick

jhamrick:
Ok, done.

minrk:
Thanks!

PR #72, opened 2015-08-28
Making a bound method in the @observe decorator

SylvainCorlay:
The solution proposed by @sccolbert to the problem with unbound method in the observe decorator. @rmorshea.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -45,7 +45,6 @@ import inspect
 import re
 import sys
 import types
-from types import FunctionType
 try:
     from types import ClassType, InstanceType
     ClassTypes = (ClassType, type)
@@ -663,10 +662,14 @@ class ObserveHandler(BaseDescriptor):
         self.func = func
         return self
 
+    def __get__(self, inst, cls):
+        if inst is None:
+            return self
+        return types.MethodType(self.func, inst)
+
     def instance_init(self, inst):
-        setattr(inst, self.name, self.func)
         for name in self.names:
-            inst.observe(self.func, name=name)
+            inst.observe(types.MethodType(self.func, inst), name=name)
 
 
 class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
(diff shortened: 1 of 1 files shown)


sccolbert:
Looks good to me, based on the previous discussion.

One more thing to think about: do we really want to be holding on to bound methods as listeners? This means the lifetime of the bound instance is tied to the lifetime of the object which is being observed. In Atom, I decompose a bound method into its constituent function and instance, and only keep a weakref to the instance. When invoking that connections, I first deref the weakref. If the instance is still alive, the method is invoked, otherwise the connection is removed. 

This pattern allows a listener to be garbage collection independently of the thing emitting the change.

References (C++ but the idea is the same):
https://github.com/nucleic/atom/blob/master/atom/src/methodwrapper.cpp#L90
https://github.com/nucleic/atom/blob/master/atom/src/methodwrapper.cpp#L47
https://github.com/nucleic/atom/blob/master/atom/src/observerpool.cpp#L220
https://github.com/nucleic/atom/blob/master/atom/src/observerpool.cpp#L227

tacaswell:
mpl has a call back registry which does the same decomposition  (https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/cbook.py#L454 and the class right above it).

SylvainCorlay:
We allow using any function as listener when using the `observe` method instead of the decorator.

rmorshea:
SylvainCorlay/traitlets#1 replaces the api for cross validation which currently uses `_name_validate`.

sccolbert:
> We allow using any function as listener when using the observe method instead of the decorator.

We do the same thing in Atom. The `observe` function checks if the listener is a plain function. If it is, it's just added to the listener list. Otherwise, a method is broken apart and wrapped in an object as described above. The object is callable, so the dispatching code can treat it the same as function. 

Additionally, it reimplements `__eq__` to return  `False` when the weakref'd instance is collected so the manager knows the wrapper can be removed. It looks very similar to the class @tacaswell posted. Atom's is contained in the `methodwrapper.cpp` listed above.

SylvainCorlay:
It should be a simple addition though.

sccolbert:
Sure - I just brought it up while we had attention to the topic.

SylvainCorlay:
Anyone with commit rights wants to hit the green button?

PR #74, opened 2015-08-28
Remove cross validation lock from HasTraits::__new__

SylvainCorlay:
Not much there. I just think that it makes it clearer that you cannot cross-validate when the instance is incomplete.


minrk:
tests suggest this isn't quite complete.

SylvainCorlay:
Closing. I will make this part of the other PR.

PR #75, opened 2015-09-03
Allow Dynamic Creation of Trait Validators

rmorshea:
Make validator registration public.

```python
   
def _odd_validator(commit):
   if not commit['value'] % 2:
       raise TraitError('Expected an odd number')
   return value
   
class A(HasTraits):
   i = Int()

a = A()
a.register_validator(_odd_validator, 'i')

a.value = 1
try:
    a.value = 2
except TraitError:
    pass
else:
    raise
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1601,7 +1601,7 @@ class Union(TraitType):
 
     def class_init(self, cls, name):
         for trait_type in self.trait_types:
-            trait_type.class_init(cls, name)
+            trait_type.class_init(cls, None)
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Since the validator registration was privatized, we should probably not do this right away.

rmorshea:
So close for now?

SylvainCorlay:
So this is mostly making `_register_validator` public.
I am not too sure about the example because there would probably be a much simpler way to do exactly the same result with a single cross-validator that reads the parity from the object.

PR #77, opened 2015-09-07
Cast unicode module name to str on Python 2

### LINKED ISSUE #57: import string bug (Py2 only)

minrk:
Finish up #58

closes #57, closes #58

### DIFF
--- a/traitlets/utils/importstring.py
+++ b/traitlets/utils/importstring.py
@@ -5,7 +5,7 @@ A simple utility to import something by its string name.
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from ipython_genutils.py3compat import cast_bytes_py2
+from ipython_genutils.py3compat import cast_bytes_py2, string_types
 
 def import_item(name):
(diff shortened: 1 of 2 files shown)


Carreau:
Travis seem genuinely not happy.

minrk:
Fixed.

Carreau:
Thanks !

PR #78, opened 2015-09-09
Add Descriptor Installation Phase

rmorshea:
At the moment appending attributes to an instance that are required by descriptors during `instance_init` involves copying and pasting `__new__` into the inherited class with that additional bit of logic. This pr makes no changes to the logic of `__new__` and simply compartmentalizing descriptor installation in a hook that can be overwritten to add these kinds of attributes. The most immediate use case I can Imagine for this change would be custom event handlers which build on the logic of the new decorator api to store callbacks on the instance.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -684,23 +684,18 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, object)):
             inst = new_meth(cls)
         else:
             inst = new_meth(cls, **kw)
-        inst._trait_values = {}
-        inst._trait_notifiers = {}
+
+        # Some descriptors raise AttributeError like zope.interface's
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@rmorshea I am changing this logic in https://github.com/ipython/traitlets/pull/70

rmorshea:
@SylvainCorlay I hadn't seen the changes you made there, but I did it this way because I thought it seemed redundant to make that check on the first, and all subsequent descriptors that require a particular attribute.

However keeping things on the Descriptors makes things cleaner. Maybe there should be a descriptor class method that's called once per type of descriptor attached to an instance?

rmorshea:
I'll close this for now though and duplicate this comment on your branch.

PR #80, opened 2015-09-15
Pass Trait to Validator Method

rmorshea:
Keeps with the old validator signature by including the corresponding TraitType in the proposal dict instead of the trait name. Accessing the metadata on a particular TraitType seems more important than passing the trait name which you can get from `trait.name` anyway.


SylvainCorlay:
Makes sense.

rmorshea:
@minrk seem reasonable?

minrk:
Yes, I think so.

PR #82, opened 2015-09-16
Modularize TraitType Get

rmorshea:
Abstract `get` from `__get__` similarly to `set` and `__set__`

Example involving an "on get" notifier using `@retrieve` like `@observe`:

```python
# The `@retrieve` decorator
def retrieve(name):
    return RetrieveHandler(name)

# Registers notifiers which are called 'on get'
class RetrieveHandler(EventHandler):

    def __init__(self, name):
        self._name = name

        if not hasattr(inst, '_retrieve_handlers'):
            inst._retrieve_handlers = {}
        if self._name in inst._retrieve_handlers:
            raise TraitError("A retriever for the trait '%s' has "
                             "already been registered" % self._name)
        method = types.MethodType(self.func, inst)
        inst._retrieve_handlers[self._name] = method

# Mixin for TraitTypes enabling them to work with `@retrieve`
class OnGetMixin(object):

    def __get__(self, obj, cls=None):
        if obj is None:
            return self
        value = super(OnGetMixin,self).get(obj, cls)
        if self.name in obj._retrieve_handlers:
            method = obj._retrieve_handlers[self.name]
            value = method({'value': value, 'owner':obj, 'trait':self})
        return value
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -425,6 +425,25 @@ class TraitType(BaseDescriptor):
             if self.name is not None:
                 obj._trait_values[self.name] = v
 
+    def get(self, obj, cls):
+        try:
+            value = obj._trait_values[self.name]
+        except KeyError:
+            # Check for a dynamic initializer.
+            dynamic_default = self._dynamic_default_callable(obj)
+            if dynamic_default is None:
+                raise TraitError("No default value found for %s trait of %r"
+                                 % (self.name, obj))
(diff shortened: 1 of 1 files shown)


minrk:
seems sensible. This is so `get` may be more easily overridden, I presume? Can you link a use case for that?

rmorshea:
@minrk, I added an example to the description.

minrk:
Thanks.

PR #88, opened 2015-09-23
Allow @validate to take multiple trait names

SylvainCorlay:
I don't think there is a reason why `@observe` can take multiple names and not `@validate`.


minrk:
Makes sense.

PR #89, opened 2015-09-23
(Do not merge - discussion) On the decorator API for default value generation

SylvainCorlay:
I was giving a try to an alternative implementation of #76, i.e. to create a decorator-based API for default value generation (`@default`) similar to the one used for observers (`@observe`) and cross-validators (`@validate`).

___
The case of default value generation differs from the "validate" and "observe" cases in that the current implementation with magic methods only looks for such magic methods in subclasses of `trait_type_instance.this_class`, by looking through the `mro` the HasTraits type... 

In other words, the goal was to ensure the following behavior.

```Python
class Foo(HasTraits):
    bar = Int()
    def _bar_default(self):
        return 4

class FooD(Foo)
    bar = Int()  # Int is overloaded in FooD, so that FooD.bar.this_class is FooD
                 # and _bar_default is not used because not defined in a subclass
                # of FooD...
```
___ 
Now, in implementing the decorator API for default value generation, I used the same trick as I introduced for observe, which is to have a decorator replacing the method with an instance of `DefaultHandler`  which inherits from `BaseDescriptor`. Later, this descriptor is picked up by the metaclass, which runs instance_init, and populates the `_trait_default_generator` dictionary. (See implementation in this PR) 

However, there is not guarantee that all the `DefaultHandler` instances will have run their instance_init when TraitType.instance_init is run. Hence, we may miss some default handlers.

Example:

```Python
class D(HasTraits):
    a = Int()
    b = Int()
    
    @default('a')
    def i_uoi(self):
        return 1
    
    def _b_default(self):
        return 1

class E(HasTraits):
    z = Int()
    b = Int()
    
    @default('z')
    def i_uoi(self):
        return 1
    
    def _b_default(self):
        return 1

print (D().a, D().b)   # prints 0, 1
print (E().z, E().b)   # prints 1, 1
```

I don't see a good solution to this problem and therefore, I think that keeping the current magic method mechanism is probably the most reasonable option.

### DIFF
--- a/traitlets/tests/test_traitlets.py.orig
+++ b/traitlets/tests/test_traitlets.py.orig
@@ -17,11 +17,19 @@ import nose.tools as nt
 from nose import SkipTest
 
 from traitlets import (
+<<<<<<< HEAD
     HasTraits, MetaHasDescriptors, TraitType, Any, Bool, CBytes, Dict, Enum,
     Int, Long, Integer, Float, Complex, Bytes, Unicode, TraitError,
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
-    ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default
+    ForwardDeclaredType, ForwardDeclaredInstance, validate, observe
+=======
+    HasTraits, MetaHasTraits, TraitType, Any, Bool, CBytes, Dict, Enum,
+    Int, Long, Integer, Float, Complex, Bytes, Unicode, TraitError, Union,
+    Undefined, Type, This, Instance, TCPAddress, List, Tuple, ObjectName,
+    DottedObjectName, CRegExp, link, directional_link, ForwardDeclaredType,
+    ForwardDeclaredInstance, validate, observe, default
+>>>>>>> default generators
 )
 from ipython_genutils import py3compat
 from ipython_genutils.testing.decorators import skipif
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
Just curious, but why can't you do the initialization in `HasTraits.__init__`?

SylvainCorlay:
@NeilGirdhar this is about default values generation when None is provided in the TraitType instance or to `__init__`.

rmorshea:
I probably should have done it this way in #76 since it's more intuitive (maybe a little less efficient), but you could do it in essentially the same way as with the bound default generator methods wherein you look for a `DefaultHandler` instance whose `_name` is the same as the trait's in the objects mro. This wouldn't require the fanciness where I register the handler to the `TraitType` instance.

rmorshea:
@SylvainCorlay The sort of logic I mentioned in the above comment, slows down Matplotlib to an absurd degree because you have to search through all the names in `dir(cls)` to find `EventHandlers` which is more than just "a little less efficient".

SylvainCorlay:
Marked as WIP. A solution to that issue would be to make the `@default` descriptor create a `_**_default` method, or add the default generator to a class-level attribute, which is equivalent.

rmorshea:
@SylvainCorlay If you did that you'd need to make sure the method was popped off the state in `__getstate__` to maintain pickleability. That harkens back to this notion of `getstate` and `setstate` methods on descriptors proposed by @NeilGirdhar in #94.

And making the default handler dict a class attribute won't work without creating it in the metaclass because it isn't something that should be shared with subclasses. That would then cause problems in the cases where `HasDescriptors` was not the base class.

SylvainCorlay:
@rmorshea good point.

PR #91, opened 2015-09-25
fix dynamic reprs for some trait types

minrk:
these are needed when generating config files, and had been broken by some deferred default generation changes.


SylvainCorlay:
:+1:

PR #92, opened 2015-09-26
simplify deprecation of MetaHasTraits

minrk:
rather than using a generic mixin deprecated class factory that's only used once


SylvainCorlay:
:+1:

SylvainCorlay:
@minrk actually it could make sense to make `MetaHasDescriptors` inherit from the deprecated `MetaHasTraits` rather than the opposite.
In `nbconvert/exporters/exporter.py`, we have a `isinstance(preprocessor, MetaHasTraits)`, which will not work in the current state of things.

minrk:
@SylvainCorlay good point, done. I also opened jupyter/nbconvert#134, which removes the check for MetaHasTraits, which was probably never the right thing to do, anyway.

SylvainCorlay:
Merging (!!)

PR #79, opened 2015-09-11
Correct Pickling Test and Edit to Pass

rmorshea:
```python
class Pickleable(HasTraits):

    i = Int()
    @observe('i')
    def _i_changed(self, change): pass
    @validate('i')
    def _i_validate(self, commit):
        return commit['value']
```
pass tests, but would have failed to pickle

PR #96, opened 2015-10-02
Revert change from #79 due to memory leak

rmorshea:
Popping `_notify_trait` off `__dict__` after `hold_trait_notifers` must occur inside the context manager to prevent memory leaks. This reverts a change which was made in #79, where popping was done in `__getstate__`, and edits the comment to reflect this potential memory leak. Previously it'd been marked for removal after dropping support for Python < 3.4.


SylvainCorlay:
Should this be closed?

rmorshea:
@minrk objections to this?

PR #97, opened 2015-10-02
Plug memory leak due to method use in observe and validate

rmorshea:
The storage of method types on the `HasTraits` instance by the `@observe` and `@valdiate` decorators causes memory leaks. This changes that, and instead stores `EventHandler` instances to access change and validation callbacks.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -503,7 +503,7 @@ class TraitType(BaseDescriptor):
     def _cross_validate(self, obj, value):
         if self.name in obj._trait_validators:
             proposal = {'trait': self, 'value': value, 'owner': obj}
-            value = obj._trait_validators[self.name](proposal)
+            value = obj._trait_validators[self.name](obj, proposal)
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -864,9 +864,12 @@ def _notify_trait(self, name, old_value, new_value):
```
minrk: the above offset calculation is no longer used

rmorshea: The way I've written it, that's not needed anymore.

minrk: That's what I mean. Since it's not used, it should be removed.

rmorshea: Is this line note in the wrong spot? I think I've fixed what you're referring to, but it wasn't here.

minrk: This line seems to have wandered off, and definitely isn't where it originated. I see that the line where it was originally is removed, thanks.

PR #100, opened 2015-10-05
Re-add notify_trait

SylvainCorlay:
ipywidgets overload `_notify_trait` and rely on it being called on `__set__` . We can probably remove it completely once widgets require the version of traitlets to come and overload `_notify_change` instead.


SylvainCorlay:
@minrk

SylvainCorlay:
Thanks

PR #101, opened 2015-10-09
Make HasTraits._notify_change public

SylvainCorlay:
This simply makes `HasTraits.notify_change` public. Indeed, ipywidgets currently makes direct use of the deprecated (and private) `_notify_trait`, and should rather call the new `_notify_change`. 

This PR acknowledges that this is public API.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -176,7 +176,7 @@ class Configurable(HasTraits):
         # merge new config
         self.config.merge(config)
         # unconditionally notify trait change, which triggers load of new config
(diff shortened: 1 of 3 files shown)


minrk:
Is it true that ipywidgets should actually be doing this in an overridden `_notify_trait`? Or should ipywidgets have been using a trait listener all along, and never overridden `_notify_trait`?

SylvainCorlay:
Good point. I will look at it and see if there is a good reason for widgets to overload it.

SylvainCorlay:
Closing, this should not be needed.

PR #108, opened 2015-10-15
Default Decorator and Generate All Default Values "On-Get"

rmorshea:
New implementation of default decorators as compared to #76. To accomplish this, the logic for creating default values was consolidated into `TraitType.get` meaning that invalid defaults will only be caught, and valid ones will only be set, in the `HasTraits` instance when there's an attempt to access the value.

```python
class A(HasTraits):
    i = Int('bad default')

a = A() # no longer fails when instance is created
a.i # instead fails when trying to access the value
```

brief default decorator example

```python
class A(HasTraits):
    i = Int()
    
    @default('x')
    def _x_default(self):
        return 1

A().i
# result : `1`
```

closes #76, closes #89


SylvainCorlay:
Will discuss with Bryan when he is here about the two solutions.

SylvainCorlay:
Closing. #114 added the default decorator.

PR #109, opened 2015-10-16
@observe and @validate docstrings

SylvainCorlay:
- The `@validate` and `@observe` decorators docstrings were missing the documentation of the signature of the change and validation handlers.
- I added an important note in the docstring of the `@validate` decorator, which is a recommendation for authors of cross validator functions.

Let me explain in more details here.
- the owner is passed to the cross validator function, because we want the validation / coercion of the proposed value to be able to read other attributes of the owner, for example, to enforce min <= value <= max in a slider (flooring and capping proposed value with `min` and `max`). 
- as a consequence, the author of such cross validator could also potentially make changes to other attributes (pushing `min` and `max` around when changing the value).

The note is there to explain that one should **never** do the latter. The reason is that such changes may not commute and are applied in arbitrary orders when existing the `hold_trait_notification` context manager.

In short:
 - coercing the proposed value is **OK**

    (flooring and capping proposed value with `min` and `max`).
 - changing other attributes of the owner is **NOT OK**
 
   (pushing `min` and `max` around with the proposed value).

This would be easy to enforce if Python had a `const` keyword.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -572,10 +572,10 @@ class TraitType(BaseDescriptor):
 
 class _CallbackWrapper(object):
     """An object adapting a on_trait_change callback into an observe callback.
-    
+
     The comparison operator __eq__ is implemented to enable removal of wrapped
     callbacks.
-    """ 
+    """
 
     def __init__(self, cb):
         self.cb = cb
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@ellisonbg @minrk

minrk:
Sensible, thanks for the thorough docstring. Python being what it is, we can't (and shouldn't) *force* the validator functions to be well behaved, but a docstring indicating what makes sense is great.

rmorshea:
How important is the rollback that occurs after a `TraitError`? That's why this isn't resolvable.

SylvainCorlay:
This is not so much related with the rollback as it is that `modification1 + modification2 != modification2 + modification1`.

PR #110, opened 2015-10-18
Remove Cross Validation Lock

rmorshea:
**Depends on #108**

I'm not against including `_cross_validation_lock`, I just can't think of any immediate use cases (especially if #108 is merged). However at the moment it's being used for things within traitlets beyond just holding cross validation and that muddles its purpose.

The three major results of this change:
+ `hold_trait_notifications` now uses `_holding_trait_notifications` to control nested contexts
+ cross validation will always occur if there's a validator to do it,
+ the following validation heuristic would no longer be possible

```python
class A(HasTraits):
    foo = Int()
    bar = Int()

    @validate('bar')
    def _bar_validate(self, change):
        self.foo = change['value']
        return change['value']

    @validate('foo')
    def _foo_validate(self, change):
        self.bar = change['value']
        return change['value']

A(foo=1)
A(bar=1)
```

The above code is only made possible due to the fact that cross validation is suspended inside the `hold_trait_notifications` context. In all other cases this would result in infinite recursion. The fact that this is possible at all seems very unintuitive.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -21,7 +21,7 @@ from traitlets import (
     Int, Long, Integer, Float, Complex, Bytes, Unicode, TraitError,
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
-    ForwardDeclaredType, ForwardDeclaredInstance, validate, observe
+    ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default
 )
 from ipython_genutils import py3compat
 from ipython_genutils.testing.decorators import skipif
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
I am :-1:  on cross validating when the holding traits: 
If you have multiple trait values (e.g. `parity`, `value`) that are cross-dependent, you cannot change them one at a time while remaining in a consistent state. This is the whole point of the `hold_trait_notification`.

rmorshea:
So you're saying that the cross validation lock allows validation operations to commute?

SylvainCorlay:
Hum I don't see what you mean? I am not talking about thing commuting here.

rmorshea:
I guess I'm not quite sure, that the changes I made effect `hold_trait_notifications` in the way that you're worried about. Is the above validation heuristic something that might be advantageous?

SylvainCorlay:
The cross-validation lock serves a purpose, which is to prevent cross-validation within the `hold_trait_validation` so that you can set multiple attributes without being concerned about inconsistency at intermediary stages, and cross-validate the final result.

rmorshea:
Sorry, in my mind that's what I meant by commute. So I'll put it back. In any case, does it make sense to stop using `_cross_validation_lock` as the nested context control?

SylvainCorlay:
`hold_trait_notifications` holds cross validation and notifications. We don't have a usecase for holding one and not the other, there is only one context manager.

rmorshea:
Ok, makes sense. At the moment there's nothing that tests for that so I can write one up.

PR #114, opened 2015-11-03
Add the `@default` decorator

SylvainCorlay:
Another shot at implementing the `@default` decorator. The reasoning is the following.

-----
 - Unlike observers and validators (which are registered at the instance-level), default value generators are class-level things. Hence `_trait_default_generators ` attribute is class-level.

    - `_trait_values`
    - `_trait_notifiers`
    - `_trait_decorators`  

    all belong to the trait **instance**.

    - `_trait_default_generators`

    belongs are **class-level** attributes of the class derived from HasTraits.

----
The `_trait_default_generators` must be created *before* any `instance_init` is called (otherwise, a TraitType may miss its default initializer).

Therefore, populating `_trait_default_generator` cannot be done at the `instance_init` but before.

This means that the meta-class magic must register all the default value generators in `_trait_default_generators` *before* any `instance_init` is run. 

This is why, in the same way we have an `instance_init` containing the part of the initialization of trait types which depends on the HasTraits instance, I added a `class_init` trait type method containing the part of the initialization depending on the HasTraits class.

-----
The expected behavior wrt inheritance is the following: 

*Default generators should only be called if they are registered in subclasses of `trait.this_type`.*

```Python
class A(HasTraits):
    bar = Int()

    @default('bar')
    def get_bar_default(self):
        return 11


class B(A):
    bar = Float()  # This should ignore the default generator defined in the base class A


class C(B):

    @default('bar')
    def some_other_default(self):  # This default generator should not be ignored since 
        return 3.0                 # it is defined in a class derived from B.a.this_class.
```

Btw, we should put this in the documentation.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -298,7 +298,7 @@ class BaseDescriptor(object):
 
     Notes
     -----
-    This implements Python's descriptor prototol.  
+    This implements Python's descriptor prototol.
 
     This class is the base class for all such descriptors.  The
     only magic we use is a custom metaclass for the main :class:`HasTraits`
@@ -317,19 +317,17 @@ class BaseDescriptor(object):
 
     def class_init(self, obj):
         """Part of the initialization which may depend on the underlying
-        HasTraits class.
+        HasDescriptors class.
 
         It is typically overloaded for specific types.
 
-        This method is called by :meth:`HasTraits.__new__` and in the
-        :meth:`BaseDescriptor.instance_init` method of descriptors holding
-        other descriptors.
+        This method is called by :meth:`MetaHasDescriptors.__init__`.
         """
         pass
 
     def instance_init(self, obj):
         """Part of the initialization which may depend on the underlying
-        HasTraits instance.
+        HasDescriptors instance.
 
         It is typically overloaded for specific types.
 
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -315,7 +315,7 @@ class BaseDescriptor(object):
     name = None
     this_class = None
 
-    def class_init(self, obj):
+    def class_init(self, cls):
         """Part of the initialization which may depend on the underlying
         HasDescriptors class.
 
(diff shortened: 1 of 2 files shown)


rmorshea:
I remember trying this at some point, but I dropped the idea. I'm trying to figure out why exactly that is, but I think it had to do with something in matplotlib where HasTraits wasn't the base class. I'll let you know if I find anything.

SylvainCorlay:
Thanks @rmorshead indeed the case of mixins is really annoying indeed that
is why I check for the presence of the _trai_default_generator attribute.
Otherwise it would not be needed.

S.

rmorshea:
I see. Though I am getting a new image comparison failure in the mpl test suite. It's something I've seen before, but I'll have to check what's causing it because I don't remember why it came up in the past.

rmorshea:
I see.

rmorshea:
There should probably be a test to ensure that mixins work.

SylvainCorlay:
I just cleaned up things a bit more. @ellisonbg do you want to have a look at this?

rmorshea:
the phrasing `install_instance` and `install_class` makes it sound like you're installing that class onto something rather than installing something onto the class.

Maybe `install_`(`to`/`on`)`_`(`class`/`instance`)?

rmorshea:
The phrasing `install_instance` and `install_class` makes it sound like you're installing the class onto something rather than installing something onto the class.

Maybe `install_(to/on)_(class/instance)`?

SylvainCorlay:
`[class/instance]_setup` ?

SylvainCorlay:
ping @minrk

ellisonbg:
I looked briefly at the code and it looks clean. But I haven't taken the time to dive into the micro-details of the mro stuff in the implementation.

But I think the behavior is the right one, especially wrt to inheritance and re-definitions of traits in subclasses.

SylvainCorlay:
It would be nice to get a beta out after this.

traitlets/traitlets.py
```diff
@@ -315,9 +315,19 @@ class that does the following:
     name = None
     this_class = None
 
+    def class_init(self, obj):
```
SylvainCorlay: review with Bryan, obj-> cls

traitlets/traitlets.py
```diff
@@ -775,13 +831,13 @@ def install_descriptors(self, cls):
                     value.instance_init(self)
 
 
-class HasTraits(HasDescriptors):
+class HasTraits(py3compat.with_metaclass(MetaHasTraits, HasDescriptors)):
 
-    def install_descriptors(self, cls):
+    def install_instance(self, cls):
```
SylvainCorlay: take out `cls`from `install_instance`.

SylvainCorlay:
Brian is +1 on `setup_instance` and `setup_class`.

rmorshea:
Sounds good. Generic is probably a good thing.

ellisonbg:
Looks great, merging. More seriously, @SylvainCorlay and I did some in person review and his recent changes address everything. Great work!

PR #116, opened 2015-11-06
CaselessStrEnum should cast to unicode

minrk:
We should never have any `str`s anywhere except as explicit `Bytes` or where bad py2 APIs require `str`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1910,8 +1910,14 @@ class Enum(TraitType):
 
 class CaselessStrEnum(Enum):
     """An enum of strings where the case should be ignored."""
-
+    
+    def __init__(self, values, default_value=Undefined, **metadata):
+        values = [py3compat.cast_unicode_py2(value) for value in values]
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -1910,8 +1910,14 @@ def info(self):
...
     """An enum of strings where the case should be ignored."""
-
+    
+    def __init__(self, values, default_value=Undefined, **metadata):
+        values = [py3compat.cast_unicode_py2(value) for value in values]
+        super(CaselessStrEnum, self).__init__(values, default_value=default_value, **metadata)
+    
     def validate(self, obj, value):
+        if not py3compat.PY3 and isinstance(value, str):
+            value = value.decode('utf8')
```
SylvainCorlay: you could also use `cast_unicode_py2` here right?

minrk: Yes - me being inconsistent about whether or not to use py3compat from one minute to the next. Should be more consistent now. It does leave the check for `str` in case of allow_none.

PR #117, opened 2015-11-06
WIP Remove uses of ipython_genutils.py3compat  in favor of six

SylvainCorlay:


jdfreder:
+1 :star:

traitlets/config/application.py
```diff
@@ -27,7 +27,9 @@
 from ipython_genutils.importstring import import_item
 from ipython_genutils.text import indent, wrap_paragraphs, dedent
 from ipython_genutils import py3compat
-from ipython_genutils.py3compat import string_types, iteritems
+
+import six
+from six import iteritems
```
minrk: I would actually just get rid of all of our uses of iteritems. I don't think we ever use it in a case where it's beneficial.

SylvainCorlay: so just `.items`?

minrk: Yup. We aren't operating on any giant dicts here. We can afford to let Python 2 be the tiniest bit slower on what we are dealing with. No need to call in six for this one.

Carreau:
Hum, this will need rebase.

SylvainCorlay:
Yeah it is stale, I have to get back to this one.

PR #121, opened 2015-11-08
Adding transform attribute to directional link

SylvainCorlay:
@ellisonbg

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1605,6 +1605,28 @@ class TestDirectionalLink(TestCase):
         b.value = 6
         self.assertEqual(a.value, 5)
 
+    def test_tranform(self):
+        """Test transform link."""
+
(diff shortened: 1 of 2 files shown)


ellisonbg:
+1

minrk:
@SylvainCorlay @ellisonbg  do you want this in 4.1? I think we can cut the beta today.

SylvainCorlay:
I would rather not postpone the release so it can certainly go without it!
Are you making a 4.1 or a 5.0 beta?

minrk:
...now I don't remember what we decided, and when I look over the conversation again it seems inconclusive. 4.1?

SylvainCorlay:
I don't think that we decided anything.

ellisonbg:
I prefer to get this into 4.1

SylvainCorlay:
Added a test.

minrk:
AOK

PR #119, opened 2015-11-06
warn on misstyped config options

Carreau:
That have annoyed me for too long: 

```
$ ipython notebook --NotebookApp.ignoreMinifiedJs=True
[W 10:15:45.741 NotebookApp] Config option `ignoreMinifiedJs` not recognized by `NotebookApp`, do you mean : `ignore_minified_js`
[W 10:15:45.742 NotebookApp] Unrecognized JSON config file version, assuming version 1
[W 10:15:45.743 NotebookApp] Config option `ignoreMinifiedJs` not recognized by `NotebookApp`, do you mean : `ignore_minified_js`
[W 10:15:45.744 NotebookApp] Config option `ignoreMinifiedJs` not recognized by `NotebookApp`, do you mean : `ignore_minified_js`
[I 10:15:45.787 NotebookApp] The port 8888 is already in use, trying another random port.
[I 10:15:45.793 NotebookApp] Serving notebooks from local directory: /Users/bussonniermatthias/dev/notebook
[I 10:15:45.793 NotebookApp] 0 active kernels
```

```
$ ipython --InteractiveShell.color=linux
[TerminalIPythonApp] WARNING | Config option `color` not recognized by `TerminalInteractiveShell`, do you mean one of : `colors ,color_info`
Python 3.5.0 |Continuum Analytics, Inc.| (default, Oct 20 2015, 14:39:26)
Type "copyright", "credits" or "license" for more information.

IPython 4.1.0-dev -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]:
```

It does warn only if if find an option which is **close** by standard definition of **close** from difflib.

This **does** make  `SingletonConfigurable` a `LoggingConfigurable`.

I'd like @fperez input on this one, as there might be historical reasons not to do it.

PR #123, opened 2015-11-09
update_config only loads *new* config

minrk:
rather than trigger `_config_changed`, which reloads *all* config, even what's already been loaded.

Should address the repeated warning discussed in #119


Carreau:
works for me.

Do you want to address the todo now? or later ?

minrk:
No, the TODO should wait for the dict-change events, which don't exist.

SylvainCorlay:
This needs a rebase.

minrk:
Rebased

PR #124, opened 2015-11-11
Call contained trait types' `class_init` and set `this_class` in `class_init`

SylvainCorlay:
... rather than in `instance_init`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1600,9 +1600,14 @@ class Union(TraitType):
         self.default_value = self.trait_types[0].default_value
         super(Union, self).__init__(**metadata)
 
-    def instance_init(self, obj):
+    def class_init(self, cls):
         for trait_type in self.trait_types:
             trait_type.this_class = self.this_class
+            trait_type.class_init(cls)
+        super(Union, self).class_init(cls)
+
+    def instance_init(self, obj):
+        for trait_type in self.trait_types:
             trait_type.instance_init(obj)
         super(Union, self).instance_init(obj)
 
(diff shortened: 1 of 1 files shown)


minrk:
Can you show an example of the difference in behavior this causes (in a test, for example)?

SylvainCorlay:
The gist of it is that

- if `trait.this_type` is set in `instance_init`, it is done every time the `HasTraits` class is instantiated, which is redundant.
- if `trait.this_type` is set in `class_init`, it is done every time the `MetaHasTraits` class is instantiated, which is only on declaration of new trait types.

SylvainCorlay:
Oh ok, I did not expect this would go in 4.1

minrk:
Good point. Probably shoudn't have merged it, but I think it'll be okay.

minrk:
We can revert and re-merge after the actual release if you'd be more comfortable.

SylvainCorlay:
I am fine with the new state - was just expressing surprise...

rmorshea:
At the moment, because `setup_class` iterates over `cls.__dict__`, it wouldn't be possible to add attributes to the class inside `class_init` because `cls.__dict__` will change size and thus an error will be thrown. This forces users to explicitly define new attributes in `setup_class` rather than dynamically in `class_init`, is this a desireable behavior?

rmorshea:
...the alternative being to iterate over `cls.__dict__.copy()`

SylvainCorlay:
@rmorshea good point. 

It seems that it is not an issue at the moment since class_init does not add things to the class. I don't think that `class_init` should edit the class though but better safe than sorry.

I would be ok to copy the dict first in `setup_class`.

rmorshea:
Should I make a pr for that then?

rmorshea:
Also, is there any reason that `this_class` can't be set inside `class_init` on the descriptor, or at the very least inside `setup_class`? It would mean that not all descriptors would have this attribute when `class_init` is called, however since it seems like the philosophy so far has been that descriptors should be agnostic to the presence of other descriptors, that shouldn't be too much of an issue.

SylvainCorlay:
Regarding the setting of `this_class` and `name`, I thought of putting it inside `class_init` but did not do it in this PR.

rmorshea:
In #129 I decided not to set `this_class` and `name` in `class_init` just to keep `BaseDescriptor` as simple as possible and instead put it in `setup_class`.

PR #126, opened 2015-11-11
only pass a change dict to `_notify_change`

rmorshea:
Changes the calls signature for `_notify_change` from `(name, type, change)` to `(change)` since both the trait name and the change type are data that should already be given in the change dict.


rmorshea:
@SylvainCorlay if this seems reasonable, it might be good to have in 4.1.

SylvainCorlay:
Yes, it makes sense.

PR #127, opened 2015-11-12
make trait name references in event handlers explicit

rmorshea:
Currently the attribute `name` on an event handler would conflict with the one assigned during setup in the metaclass. The workaround has just been to use `_name`, so to reduce confusion I've just extended that attribute to be `trait_name` (and `trait_names` plural).

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -804,33 +804,33 @@ class ObserveHandler(EventHandler):
 
     def __init__(self, names, type):
         if names is All:
-            self.names = [All]
+            self.trait_names = [All]
         else:
-            self.names = names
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@rmorshea but `names` does not conflict right?

rmorshea:
No. It just seemed confusing to have one be a "private" attribute while the other one wasn't even though really are referring to the same sort of thing.

rmorshea:
Just one of those things that's been bugging me.

SylvainCorlay:
@rmorshea I am :+1: on this. We should merge it before 4.1. (Otherwise it would strictly speaking be a change to the public API of the event handlers.)

PR #129, opened 2015-11-15
Streamline Class Setup Logic in MetaHasDescriptors

rmorshea:
Reduces the number of times the metaclass must iterate over the classdict, and uses a copy of `cls.__dict__` when iterating in `setup_class` so that changes in the size of the dict, which could potentially be caused by the `class_init` methods of descriptors, won't throw errors.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -650,43 +650,35 @@ class MetaHasDescriptors(type):
     """
 
     def __new__(mcls, name, bases, classdict):
-        """Create the HasDescriptors class.
-
-        This sets :attr:`name` attribute of each descriptor in the class dict.
-        """
+        """Create the HasDescriptors class."""
         for k, v in iteritems(classdict):
-            if isinstance(v, BaseDescriptor):
-                v.name = k
-
             # ----------------------------------------------------------------
             # Support of deprecated behavior allowing for TraitType types
             # to be used instead of TraitType instances.
-            elif inspect.isclass(v):
-                if issubclass(v, TraitType):
-                    warn("Traits should be given as instances, not types (for example, `Int()`, not `Int`)",
-                         DeprecationWarning, stacklevel=2)
-                    vinst = v()
-                    vinst.name = k
-                    classdict[k] = vinst
+            if inspect.isclass(v) and issubclass(v, TraitType):
+                warn("Traits should be given as instances, not types (for example, `Int()`, not `Int`)",
+                     DeprecationWarning, stacklevel=2)
+                classdict[k] = v()
             # ----------------------------------------------------------------
 
         return super(MetaHasDescriptors, mcls).__new__(mcls, name, bases, classdict)
 
     def __init__(cls, name, bases, classdict):
-        """Finish initializing the HasDescriptors class.
-
-        This sets the :attr:`this_class` attribute of each BaseDescriptor in the
-        class dict to the newly created class ``cls``.
-        """
-        for k, v in iteritems(classdict):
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -650,43 +650,35 @@ class MetaHasDescriptors(type):
...
-        This sets the :attr:`this_class` attribute of each BaseDescriptor in the
-        class dict to the newly created class ``cls``.
-        """
-        for k, v in iteritems(classdict):
-            if isinstance(v, BaseDescriptor):
-                v.this_class = cls
-        cls.setup_class()
+        """Finish initializing the HasDescriptors class."""
         super(MetaHasDescriptors, cls).__init__(name, bases, classdict)
+        cls.setup_class()
```
SylvainCorlay: Would this work if setup_class got the above `classdict` argument instead of copying `cls.__dict__`?

rmorshea: As long as it's not iterating over `cls.__dict__` it should work. Decided to keep the signature clean since I didn't think a shallow copy of the dict would be terribly intensive (could be wrong though).

SylvainCorlay: Another thing that could be done is to set `this_class` and `name` from within `class_init`, but it would require passing the `name` argument to the `class_init` method.

SylvainCorlay:
@ellisonbg @minrk any take on this? (Re: setting `name` and `this_class` in `class_init`)

traitlets/traitlets.py
rmorshea: Initially I was for the idea, but I decided that it was better to keep the `BaseDescriptor`'s methods clean so anyone inheriting from it wouldn't have to worry about using super. It's definitely an option, but I think I'm in favor of keeping that out of `class_init`.

minrk:
I confess that I don't have a complete grasp on the relevant bits, but setting name and this_class in class_init seems reasonable on its face.

SylvainCorlay:
OK, if we want to do this, we need to do it before 4.1 because it would be a backward incompatible change later.

ellisonbg:
How is it not BW incompat now then?

On Sat, Nov 28, 2015 at 3:21 PM, Sylvain Corlay <notifications@github.com>
wrote:

> OK, if we want to do this, we need to do it before 4.1 because it would be
> a backward incompatible change later.
>
> —
> <https://github.com/ipython/traitlets/pull/129#issuecomment-160345327>.
>



-- 
Brian E. Granger

SylvainCorlay:
- This PR is backward compatible IMO.
- `TraitType.class_init` is a new method in 4.1.
- The proposal in the discussion to set `this_class` and `name` in
`class_init` requires changing the signature, so doing after 4.1 would be
backward incompatible.

S.

minrk:
+1 on adding this to 4.1

SylvainCorlay:
The code looks good to me, including the the change of signature of `class_init` that was just pushed by @rmorshea .

takluyver:
I can no longer start the notebook, and git bisect points to commit 7a0e20c1b7b1d3904e2a4c23d11689341c1c7c24 in this PR.

SylvainCorlay:
@takluyver I am looking at it.

takluyver:
Thanks @SylvainCorlay . I was trying to create a failing test, but so far it's not failing.

SylvainCorlay:
OK, I understand what is going on. 

- We used to never set the name of the element traits.
- Now that we do, calling instance_init on element_traits sets their default value to the corresponding trait attribute in the parent class.

PR #131, opened 2015-11-19
dynamically generate HasTraits class documentation

### LINKED ISSUE #130: Trait introspection
A class with multiple traits usually has an extremely useless call signature of `__init__`, and is often lacking docstrings. This is understandable, since traits themselves can come with descriptions.

On the other hand, it would be useful to automatically document the traits present in the class in the docstring by e.g. using a decorator, or even fiddle with the `__init__` call signature, since that would enable tools like `sphinx` to produce more complete documentation, and would allow introspection by third party software (unlike the related ipython/ipython#9000).

rmorshea:
A first go at generating `TraitType` documentation inside the `HasTraits.__init__` docstring.

```python
class A(HasTraits):
    i = Int().tag(allow_none=True, default_value=1)
    j = Any().tag(help='some help text for j', data={'x':1})

    def __init__(self, *args, **kwargs):
        """The constructor for the class
        
        More details about the constructor"""
        pass
    
    @default('j')
    def h(self): pass
    
    @validate('j')
    def g(self, commit):
        return commit['value']
    
    @observe('i', 'j')
    def f(self, change):
        pass
    
help(A.__init__)
```
has a docstring that reads:
```text
:Traits of :class:`A` Instances:
--------------------------------
i : an int
    - trait metadata
        * ``allow_none = True``
        * ``default_value = 1``
    - event handlers
        * :method:`A.f` - observes changes

j : any value
    :help: some help text for j

    - trait metadata
        * ``data = {'x': 1}``
    - event handlers
        * :method:`A.f` - observes changes
        * :method:`A.g` - validates values
        * :method:`A.h` - sets the default
```

closes #130

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1401,7 +1401,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
             return trait.metadata.get(key, default)
 
     @classmethod
-    def class_own_events(cls, name):
+    def class_own_trait_events(cls, name):
         """Get a dict of all event handlers defined on this class, not a parent.
 
         Works like ``event_handlers``, except for excluding traits from parents.
@@ -1411,7 +1411,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
                 if getattr(sup, n, None) is not e}
 
     @classmethod
-    def events(cls, name=None):
+    def trait_events(cls, name=None):
         """Get a ``dict`` of all the event handlers of this class.
 
         Parameters
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -58,6 +58,7 @@ import six
 from .utils.getargspec import getargspec
 from .utils.importstring import import_item
 from .utils.sentinel import Sentinel
+from .utils.docgen import traitlet_documentation
 
 SequenceTypes = (list, tuple, set, frozenset)
 
@@ -161,7 +162,6 @@ def is_trait(t):
     return (isinstance(t, TraitType) or
             (isinstance(t, type) and issubclass(t, TraitType)))
 
-
 def parse_notifier_name(names):
     """Convert the name argument to a list of names.
 
@@ -185,7 +185,6 @@ def parse_notifier_name(names):
         for n in names:
             assert isinstance(n, six.string_types), "names must be strings"
         return names
-
 
 class _SimpleTest:
     def __init__ ( self, value ): self.value = value
@@ -814,14 +813,13 @@ def validate(*names):
     """
     return ValidateHandler(names)
 
-
-def default(name):
+def default(*names):
     """ A decorator which assigns a dynamic default for a Trait on a HasTraits object.
 
     Parameters
     ----------
-    name
-        The str name of the Trait on the object whose default should be generated.
+    *names:
+        The str names of the Traits on the object whose default should be generated.
 
     Notes
     -----
@@ -853,10 +851,12 @@ def default(name):
                 return 3.0                 # ignored since it is defined in a
                                            # class derived from B.a.this_class.
     """
-    return DefaultHandler(name)
+    return DefaultHandler(names)
 
 
 class EventHandler(BaseDescriptor):
+
+    info_text = None
 
     def _init_call(self, func):
         self.func = func
@@ -874,8 +874,14 @@ class EventHandler(BaseDescriptor):
             return self
         return types.MethodType(self.func, inst)
 
+    def info(self):
+        """Returns helpful info"""
+        return self.info_text
+
 
 class ObserveHandler(EventHandler):
+
+    info_text = "observes changes"
 
     def __init__(self, names, type):
         self.trait_names = names
@@ -887,6 +893,8 @@ class ObserveHandler(EventHandler):
 
 class ValidateHandler(EventHandler):
 
+    info_text = "validates values"
+
     def __init__(self, names):
         self.trait_names = names
 
@@ -896,12 +904,15 @@ class ValidateHandler(EventHandler):
 
 class DefaultHandler(EventHandler):
 
-    def __init__(self, name):
-        self.trait_name = name
+    info_text = "sets the default"
+
+    def __init__(self, names):
+        self.trait_names = names
 
     def class_init(self, cls, name):
         super(DefaultHandler, self).class_init(cls, name)
-        cls._trait_default_generators[self.trait_name] = self
+        for name in self.trait_names:
+            cls._trait_default_generators[name] = self
 
 
 class HasDescriptors(six.with_metaclass(MetaHasDescriptors, object)):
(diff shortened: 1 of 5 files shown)


rmorshea:
@akhmerov I've never used `sphinx`, but hopefully this is something like what you were looking to get out of #130. I didn't really see an easy way to use a descriptor to get the job done since it needs access to the class, and from what I was able to learn, manipulating call signatures really only seems feasible with [PEP 0362](https://www.python.org/dev/peps/pep-0362/), but that's for Python>3.3

akhmerov:
It should be quite sufficient for `sphinx`, and yes this is roughly what I thought about. How do you treat other trait arguments, (`default_value`, `allow_none`, `read_only`, `help`)?

Not sure what the proper thing is to do with the formatting; Perhaps the numpy docstring standard:
```
Traits:
-------
<trait_name> : <trait_type>[, optional]
    <help>[ (default <default_value>)].
```

You can possibly omit class name from the appended docstring.

Should read-only traits be omitted from the doc? @minrk mentioned that there's a nontrivial decision to be made regarding subclassing.

akhmerov:
Signature manipulation: perhaps then do it optionally, checking for the interpreter version?

rmorshea:
I only looked into signature manipulation briefly, so I'll try to confirm, but I think the issue prior to 3.3 is that functions compile differently depending on signature. Given that, `eval` would really be the only way to fundamentally alter the signature. The other option would be to wrap `__init__` (but again, I don't think a decorator would work since you need access to the class).

akhmerov:
Oh, I just mean "rewrite the signature on 3.3, do nothing < 3.3". I cannot think of negative consequences of a less informative signature.

rmorshea:
Gotcha. In terms of doc formatting in `__init__` would it be confusing that the `trait_type` instance only exists on the class? I that with the numpy style, someone would expect `hastraits.name` would be an instance of the `trait_type` rather than something which satisfies the validation conditions for a `trait_type` instance present on the class.

akhmerov:
Sorry, didn't get that. What exactly do you mean? What sort of confusion do you expect?

Also a disclaimer I should have made earlier: I'm not a dev, just another guy with an opinion, and I'm not even sure if devs consider this reasonable to implement.

rmorshea:
So with traitlets having a class like this:

```python
class A(HasTraits):
    i = Int()

a = A()
out = a.i
```

`out

rmorshea:
Based on my experience, I think having documentation in some capacity that describes what traits exist on the class would be helpful. At the moment, the only real way to understand what's going on is to look at the source.

I'd have to think about how to implement it, but it might even be possible to include information on what static notifiers are associated with each trait providing more context for what the class does.

akhmerov:
Ah, my apologies, I was unclear. By `<trait_type>` in the doc I really meant also what you have written, so `int` for an `Int` trait, and `any` for `Any`. I don't think there's a reason why the actual trait types must be mentioned in the docstring.

rmorshea:
I think I'll wait to work through signature manipulation until there's consensus on whether this is needed. For now though, I think this pretty much covers documentation of `TraitTypes` and their notifiers.

akhmerov:
Is there anything missing in this PR? It's quite important in making `ipywidgets` interface manageable (because otherwise there are essentially no docstrings).

minrk:
Sorry, this fell of my radar. This needs a rebase, and it doesn't work at all on Python 3, due to the way it detects the `__init__` method, but I think the basic reasoning is sound, and super valuable. @rmorshea will you have a chance to rebase and finish this up?

traitlets/traitlets.py
```diff
@@ -872,6 +885,112 @@ def setup_instance(self):
...
         super(HasTraits, self).setup_instance()
 
+    @classmethod
+    def setup_class(cls):
+        cls._write_docs_to_class()
+        MetaHasTraits.setup_class(cls)
+
+    @classmethod
+    def _write_docs_to_class(cls):
+        if hasattr(cls.__init__, 'im_func'):
```
minrk: There's no im_func on init on Python 3

traitlets/traitlets.py
```diff
@@ -872,6 +885,112 @@ def setup_instance(self):
...
+            docs = cls._rst_trait_help() or ""
+            if docs:
+                # copy the image function
+                im = cls.__init__.im_func
+                n, c, d, g = (im.__name__, im.__code__,
+                              im.__defaults__, im.__globals__)
+                f = types.FunctionType(c, g, n, d, im.__closure__)
```
minrk: On Python 3, you should be able to set `method.__doc__ = 'foo'` without this copying.

rmorshea:
Working on this right now...

traitlets/traitlets.py
rmorshea: I did this for two reasons:

1. I edit `im_func` because, as it turns out, in Python 2, `method.__doc__` isn't writable
2. I copy the image function because all subclass reference `cls.__init__`. Thus, because of 1, all subclasses would essentially be trying to edit the parent docstring.

So instead of contend with all that I've now decided to assign trait/event docs to `cls.__doc__`.

rmorshea:
@minrk, this now depends on #209

rmorshea:
@minrk now rebased and tests pass.

In 97897ba, the dynamic trait documentation is written to `cls.__init__.__doc__`. However in 01adb49, it's in `cls.__doc__` - this has a simpler implementation because `__doc__` operates the same in Python2 and Python3, and `cls.__doc__` isn't reference by subclasses.

So depending on where we prefer the documentation to be I can either revert to 01adb49, or squash the commits. I think putting them in `cls.__init__.__doc__` makes more sense, because attributes are writable via `**kwargs` in `__init__`, but ultimately I think it's subjective.

minrk:
Sorry, I think this needs a rebase due to the revert of class_traits changes.

traitlets/utils/docgen.py
```diff
@@ -0,0 +1,155 @@
...
+    docs = (t*2 + '* ').join(bullets)
+
+    return docs
+
+# - - - - - - - - - - - - - - - -
+# Public Documentation Generator
+# - - - - - - - - - - - - - - - -
+
+def trait_documentation(cls, name=None):
+    """Generate documentaiton for traits and event handlers on the class
```
minrk: documentation

rmorshea:
@minrk I'm giving this [an update](https://gist.github.com/rmorshea/acd061cfedb701469989154675aeb73e). However it's still a work in progress. In general I'd like the doc generation to be more modular. This will make it easier to write `DocLog` classes for new vanilla traits or custom traits defined by applications that use traitlets. I've still got a lot to do in terms of testing to make it read for a PR, but for right now this is functionality exists:

```python

class HasInt(HasTraits):
    
    i = Int()

class HasRangedInt(HasInt):
    i = Int(min=0, max=10,
            default_value=None,
            help='nothing interesting'
           ).tag(tag='value')

logger_type = trait_documentation(HasInt)
log0 = logger_type(HasInt)
log1 = logger_type(HasRangedInt)
```

PR #132, opened 2015-11-22
Populating the readme and some doc fixes

SylvainCorlay:


README.md
```diff
@@ -1,6 +1,126 @@
...
+return the default value. For example:
+
+```Python
+import getpass
+from traitlets import HasTraits, Unicode, observe
+
+class Identity(HasTraits):
+    username = Unicode()
+
+    @default('username')
```
Carreau: I'm going to be annoying, I know, but reading that I'm wondering if for a future iteration it wouldn't be nice to (also) be able to use `@username.default` as a decorator. It would be closer to how property works. Agreed it's not doable on subclass....

README.md
```diff
@@ -1,6 +1,126 @@
...
+[pip installed](https://pip.readthedocs.org/en/stable/installing/) and run:
+
+```
+pip install traitlets
+```
+
+For a development installation:
+
+* Clone this repository and `cd` into it
+* `pip install --pre -e .`
```
Carreau: `--pre` is likely not needed with `-e` :-P

Carreau:
+1

README.md
rmorshea: @Carreau I think that might get a little messy. Since the problem of subclassing is similar with `property`, people will likely attempt to use the syntax `@Parent.username.default` which, without a management system inside the descriptor, would break `Parent`.

SylvainCorlay: Another thing that we could do in the future was `@observe('foo.bar')` but I am not sure we have a use case for this for now.

rmorshea: @Carreau I deleted the last comment I made because what I said was totally wrong. The real problem with doing decoration similar to `property` is that the `name` attr of the descriptors isn't defined until after the decorator is run. So ironically, the only place this would work easily is on subclasses since the traits on the parent class would already have their `name` attrs set.

I made a mockup of what it might look like here: #134

PR #137, opened 2015-11-26
Make the JsonFileLoader a context manager

Carreau:
On exit of context manager write file back to disk.
That makes it easy to update config.

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -385,7 +385,16 @@ class FileConfigLoader(ConfigLoader):
         self.full_filename = filefind(self.filename, self.path)
 
 class JSONFileConfigLoader(FileConfigLoader):
-    """A JSON file loader for config"""
+    """A JSON file loader for config
+
+    Can also act as a context manager that rewrite the configuration file to disk on exit.
+
+    Example::
+
+        with JSONFileConfigLoader('myapp.json','/home/jupyter/configurations/') as c:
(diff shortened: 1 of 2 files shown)


minrk:
I like it! We can merge once we ship 4.1.

Carreau:
Todo, maybe default the config file to 0x600

Carreau:
> Todo, maybe default the config file to 0x600

Ok, I don't know why I wrote that there, it's unrelated to this PR.

traitlets/config/loader.py
```diff
@@ -414,6 +423,21 @@ def _convert_to_config(self, dictionary):
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """
+        Exit the context manager but do not handle any errors.
+
+        In case of any error, we do not want to write the potentially broken
+        configuration to disk.
+        """
+        with open(self.full_filename, 'w') as f:
+            f.write(json.dumps(self.config, indent=2))
```
minrk: If there is an error here, the file will be left empty due to the `open(...'w')`. Perhaps dump the config prior to opening the file, so that it raises before deleting existing contents.

minrk: ...A failure mode you can test!

Carreau: Did you take care of that after merging ? Or should I open an issue and do it (sorry I was unavailable to work on that yesterday)

PR #138, opened 2015-11-30
add Path trait

minrk:
allows specifying list-of-text paths as path-separated strings: 'foo:bar:baz'

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -9,6 +9,7 @@
 
 import pickle
 import re
+import os
 import sys
 from unittest import TestCase
 from ._warnings import expected_warnings
@@ -18,7 +19,7 @@ from nose import SkipTest
 
 from traitlets import (
     HasTraits, MetaHasTraits, TraitType, Any, Bool, CBytes, Dict, Enum,
-    Int, Long, Integer, Float, Complex, Bytes, Unicode, TraitError,
+    Int, Long, Integer, Float, Complex, Bytes, Unicode, TraitError, Path,
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
     ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
Is this specific to paths?

minrk:
Not technically, but I think we should treat it as such. If we want this syntax for specifying `List(Unicode)` generally, I would do it at a different level (in config parsing).

SylvainCorlay:
Is this for `4.2`?

minrk:
4.2, I think we can call 4.1 bugfix-only frozen at this point.

NeilGirdhar:
Why doesn't this use `pathlib` instead of `os.path…`?  `pathlib` is a much nicer interface.

minrk:
Because pathlib's new in 3.4 and offers ~nothing of relevance to this that I'm aware of.

NeilGirdhar:
@minrk well, the paths are now platform-independent tuples rather than strings…  Anyway, it doesn't matter to me, I was just wondering.

Carreau:
Needs rebase.

minrk:
Probably not worth having this in core traitlets.

TomAugspurger:
Any thoughts on revisiting this in light of [PEP 519](https://www.python.org/dev/peps/pep-0519/) (File system protocol)? I ran into a problem today where handling various path-like objects uniformly would be nice.

Actually, reading through the comments the original proposal was for a list of paths, is that correct? Would a `Path`  type that coerces a path-like things to either unicode or bytes be a welcome addition? It's not too hard to implement on my own if you think it's not worth including.

minrk:
That's what I was originally after, yes, but the discussion wandered toward single path objects, which is fine, too. A `PathLike` trait would be interesting. I'm not sure how to deal with it in terms of backported APIs from 3.6. I'd start with implementing the trait in your own package, and see how it goes, then we can look at merging upstream if you like it. How does that sound? I do this [in JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/traitlets.py) for a few simple things.

TomAugspurger:
Yeah, that sounds good to me. Thinking about it more, by defining my own is probably better since I'm able to store it as a `pathlib.Path` object, but traitlets couldn't since that's not in the stdlib for older pythons.

Zsailer:
I created a lightweight traitlets module for pathlib `Path` objects [here](https://github.com/Zsailer/traitlets_paths), in case anyone finds it useful.

I know this thread is old/closed, but I thought it was worth mentioning.

PR #139, opened 2015-11-30
stop setting `this_class`

rmorshea:
These changes follow from those relating to `this_class` in #129.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1601,7 +1601,6 @@ class Union(TraitType):
 
     def class_init(self, cls, name):
         for trait_type in self.trait_types:
-            trait_type.this_class = self.this_class
             trait_type.class_init(cls, name)
         super(Union, self).class_init(cls, name)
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I agree for the change related to `this_class` but the `_validate` takes `allow_none` into account where `validate` does not.

traitlets/traitlets.py
```diff
@@ -1598,10 +1598,15 @@ def __init__(self, trait_types, **metadata):
         self.info_text = " or ".join([tt.info_text for tt in self.trait_types])
         self.default_value = self.trait_types[0].default_value
         super(Union, self).__init__(**metadata)
+        # since we call `validate` instead of `_validate` on the
+        # specified trait types, we need to validate values of
+        # `None` at the level of the `Union` instance.
+        for tt in self.trait_types:
+            if tt.allow_none:
+                self.allow_none = True
```
rmorshea: @SylvainCorlay this should take care of the issue with `allow_none`

SylvainCorlay:
- regarding the setting of `this_class` I am :+1: . I would this part right away.
- I am not sure what is the correct logic for the `Union`. The current implementation tries to validate each trait type including the cross-validation. Could we enable the cross-validation lock before this loop?

rmorshea:
+ I'll make that proposition in a separate pr so this can be merged.
+ Also, I think using the cross-validation lock is the right way to go.

PR #140, opened 2015-12-01
engage cross-validaiton lock for trait types of a union

rmorshea:
When a `Union` validates a value, it calls `_validate` on each of the traits it depends on. However, this means that cross-validation is performed multiple times - once for each `TraitType` the `Union` is wrapping, and then again for the `Union` itself. It doesn't make sense for this to be the case since the `Union` is passed to the cross-validator and would allow the above behavior to be replicated using its `trait_types` attribute without calling the method multiple times.


rmorshea:
Sorry, force pushed before I committed on accident.

PR #141, opened 2015-12-01
Fix class_init for Container and Union elements

SylvainCorlay:
This fixes the breakage that @takluyver reported introduced in #129 . 

The gist of it is that we used to never set the `name` attribute of the element TraitTypes for containers and Union. If we do it, their `instance_init` is going to run and replace the default value.

The commit fixes the issue. I am looking at a more elegant solution.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1601,7 +1601,7 @@ class Union(TraitType):
 
     def class_init(self, cls, name):
         for trait_type in self.trait_types:
-            trait_type.class_init(cls, name)
+            trait_type.class_init(cls, None)
         super(Union, self).class_init(cls, name)
 
     def instance_init(self, obj):
@@ -2034,7 +2034,7 @@ class Container(Instance):
 
     def class_init(self, cls, name):
         if isinstance(self._trait, TraitType):
-            self._trait.class_init(cls, name)
+            self._trait.class_init(cls, None)
         super(Container, self).class_init(cls, name)
 
     def instance_init(self, obj):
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
cc @rmorshea

rmorshea:
should name be `''` or `None`?

SylvainCorlay:
The real question is if we should call `instance_init` at all for the child traits.

SylvainCorlay:
@rmorshea now setting to `None` since it was the previous behavior.

rmorshea:
Hmmm... the reason I would say calling `instance_init` might be a problem is because the preparations done by one trait might conflict with those from another. I think the beginings of this issue can be seen through the fact that these traits with `name = None` set `_trait_values[self.name]`.

SylvainCorlay:
@minrk I would like to get this in since it resumes the old behavior and fixes the issue.

I have a branch with the change in instance_init logic but I realized that in the case of ForwardDeclaredInstance, it is a bit more involved. Hence I would go with this for now and probably for 4.1.

takluyver:
I think this is right. The name attribute of a trait is used as "what is this trait called as an attribute of the HasTraits instance it belongs to". The trait instances for typechecking the elements in a container are not an attribute on a HasTraits instance, so it doesn't make sense for `name` to be set on them.

If I was redesigning traitlets from scratch, I would make a clearer separation between the type checking parts and handling trait values (i.e. defaults, change notifications). Using traits for the elements in a list only involves the former part.

PR #144, opened 2015-12-08
faster trait collection via metadata

rmorshea:
reduces the number of iterations being performed in `Hastraits.class_traits` and `Hastraits.traits`


rmorshea:
Is there a that we have both `class_traits` and `traits`? I understand why the unbound method exists, but it seems redundant to have the bound one as well if it's simply going to use `self.__class__` to perform the operation. Why not remove the bound method and call `class_traits` `traits`?

minrk:
I'm guessing historical reasons. I think deprecating `class_traits` and making `traits` a class method seems sensible, at least before I've read much into the code.

PR #146, opened 2015-12-12
_notify_change should be public API

SylvainCorlay:
1.  Changes a 'trait_change' to 'change' which was left out in #107 .

2. Rename method `_notify_change` to `notify_change`, making it public API.
    The reason why `notify_change` should be public API becomes clear in ipython/ipywidgets/pull/270:
     - ipywidgets directly uses the private method `_notify_traits` which is replaced by the more general `_notify_change`.
     - If, instead of overloading `_notify_trait` or `notify_change`, we used `observe` to register the notification to the front-end we could not guarantee that these would be triggered before other trait notifications. It is important that they are because if another trait notifications raises an exception, the notification of the front-end is never done and we end up in an inconsistent state. Therefore, ipywidgets needs to overload `notify_change` which should be public API.

cc @jdfreder @minrk

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -186,18 +186,18 @@ class Configurable(HasTraits):
         # merge new config
         self.config.merge(config)
         # unconditionally notify trait change, which triggers load of new config
-        self._notify_change({
+        self.notify_change({
             'name': 'config',
             'old': oldconfig,
             'new': self.config,
             'owner': self,
-            'type': 'trait_change', 
+            'type': 'change',
         })
 
     @classmethod
     def class_get_help(cls, inst=None):
         """Get the help string for this class in ReST format.
-        
+
         If `inst` is given, it's current trait values will be used in place of
         class defaults.
         """
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -37,7 +37,7 @@ def change_dict(*ordered_values):
 
 class HasTraitsStub(HasTraits):
 
-    def _notify_change(self, change):
+    def notify_change(self, change):
         self._notify_name = change['name']
         self._notify_old = change['old']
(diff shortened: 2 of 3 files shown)


jdfreder:
cc @rmorshea

SylvainCorlay:
If we could merge this I would love to get 4.1 out right after.

rmorshea:
This makes sense as I've implemented a `force_callbacks` function in my matplotlib pr that I can probably be removed if this is made public. So I think it's a good idea.

SylvainCorlay:
@rmorshea  I don't understand what you mean. I am not forcing or muting any notification here.

rmorshea:
It's not that you are in the pr. It's just that it allows someone to do that (and I presume that's what it's used for in ipywidgets). In mpl, I'm using my `force_callbacks` function in tandem with muting in a case where (to my knowledge) a context manager can't be used, to accomplish something with similar results:

```python
class Base(HasTraits):

    a = Something()
    
    def __init__(self, *force_notifiy):
        for c in force_notify:
            self.notify_change(c)
        # do something dependant
        # on post-notified state

class Mixin(object):

    def __init__(self, value):
        change = {'name': 'a', 'owner': self,
                  'old': self.a, 'type': 'change'}
        with self.mute_trait_notificaitons():
            self.a = value
        change['new'] = self.a
        super(Mixin, self).__init__(change)
        
def class_factory(klass):
    return type('FromFactory',
                (klass, Base))

inst = class_factory(Mixin)(#something)
```

rmorshea:
Merging now is reasonable. I just think it might reopen a discussion as to whether "muting" is useful.

SylvainCorlay:
ok, ipywidgets overloads the method to send notifications to the front-end *before* any trait notification fires.

SylvainCorlay:
I will merge tomorrow morning unless someones reacts before.

PR #149, opened 2015-12-17
use observe in traitlets.config

minrk:
instead of deprecated APIs

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -22,7 +22,7 @@ from traitlets.config.loader import (
 )
 
 from traitlets.traitlets import (
-    Unicode, List, Enum, Dict, Instance, TraitError
+    Unicode, List, Enum, Dict, Instance, TraitError, observe, default
 )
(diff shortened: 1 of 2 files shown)


minrk:
This has already broken the IPython kernel. I need to make sure to update this such that `_config_changed` accepts both the `@observe` *and* pre-`@observe` signatures.

SylvainCorlay:
It is clear that if someone registers `@observe('_foo_changed')` and you override the method in derived classes, the overridden method will not be called. Do you want to give priority to the old API in the implementation so that whenever we have a `_foo_changed` *method*, it is called with priority over the registered one?

minrk:
For this to be valid as 4.1, I need Configurable subclasses that have something like:

```python
def _config_changed(self, name, old, new):
    ...
    super()._config_changed(name, old, new)
```

to work *exactly* as before. This should also work in such a way that the base Configurable class doesn't produce any DeprecationWarnings for having `_config_changed` defined.

PR #151, opened 2015-12-17
Work on deprecation warnings

minrk:
- handle super offset to get correct stacklevel for `init(**metadata)` warning
- use `warn_explicit` to point to deprecated magic method definitions,
  instead of the event-triggering code, which isn't relevant.

(includes #149)

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -22,7 +22,7 @@ from traitlets.config.loader import (
 )
 
 from traitlets.traitlets import (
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
wow, I did not know one could do this with `stacklevel`. This looks good to me. Is it ready in your opinion?

minrk:
Yup, I think this one's ready. There's more thinking to do on #149, since the code is not currently backward-compatible.

PR #152, opened 2015-12-18
Don't inspect observer signatures

minrk:
The main point of the new API is that we don't need to do different things with different signatures.

Listeners must always take one positional arg. Zero-arg listeners are not allowed.


SylvainCorlay:
I am :+1: on this. This is simplifying the headache of supporting the old API for magic methods.

PR #153, opened 2015-12-19
getting bound methods requires handler name

rmorshea:
Retrieving bound methods from event handlers is only valid if their `name` attribute is not `None` (i.e. they were registered with `@observe`). Without checking that `name` is not `None`, EventHandler instances registered using `HasTraits.observe` will fail.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1046,7 +1046,7 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, HasDescriptors)):
 
             if isinstance(c, _CallbackWrapper):
                 c = c.__call__
-            elif isinstance(c, EventHandler):
+            elif isinstance(c, EventHandler) and c.name is not None:
                 c = getattr(self, c.name)
             
             c(change)


rmorshea:
ping @minrk

minrk:
@rmorshea is there a test we can add?

rmorshea:
added a test and an optional argument to `ObserveHandler.__init__` in which you can pass the handler's function.

SylvainCorlay:
@rmorshea in which case would we use an EventHandler that is not used as a descriptor?

rmorshea:
In #111 I was wrapping everything in event handlers because I needed to keep track of tagging information, however I just recently changed things so doing that's probably unnecessary.

In any case, I think you're right - adding a test as if registering unbound event handlers is a feature doesn't seem right. However the error that arises, if you were to use unbound event handlers, isn't indicative of what went wrong, so checking that `name` is not `None` still seems like a good idea, either to raise an informative error, or simply let it slide.

rmorshea:
#111 involves more than just observing tags, but for what it does, everything's got to be wrapped.

minrk:
Do we want to merge this (with a confirming test) before 4.1? I think I've done the testing I need for the downstream verifications that we are actually backwards-compatible, so we should be able to release this week. I'm just not sure which side of the release this sits on.

SylvainCorlay:
I am fine with this, but I don't think that this should be required.

PR #154, opened 2015-12-20
use logging.shutdown()

Carreau:
As mention in [the docs](https://docs.python.org/3/library/logging.html?highlight=logging#logging.shutdown)

Though they don't say what does/does not happen if you don't.


minrk:
I believe it.

PR #155, opened 2015-12-21
Add `@observe_compat` decorator

minrk:
shim for *explicit* backward-compatible adoption of new APIs.

Using the shim allows classes to update to the new API without breaking subclasses that may be overriding observers with the old API.

```python
class Parent(HasTraits):
    @observe('name')
    @observe_compat
    def _name_changed(self, change)

class Child(parent):
    def _name_changed(self, name, old, new):
        super()._name_changed(name, old, new)
```

I'm not sure if there's an easy way to do this without the explicit additional shim,
but it's possible. I'll spend a little more time investigating that.

If we go with this, we should probably add a similar one for validate.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -21,7 +21,7 @@ from traitlets.config.loader import (
 )
 
 from traitlets.traitlets import (
-    Unicode, List, Enum, Dict, Instance, TraitError, observe, default
+    Unicode, List, Enum, Dict, Instance, TraitError, observe, observe_compat, default,
 )
 from ipython_genutils.importstring import import_item
 from ipython_genutils.text import indent, wrap_paragraphs, dedent
@@ -82,6 +82,7 @@ def catch_config_error(method, app, *args, **kwargs):
 
 class ApplicationError(Exception):
     pass
+
 
 class LevelFormatter(logging.Formatter):
     """Formatter with additional `highlevel` record
@@ -152,8 +153,12 @@ class Application(SingletonConfigurable):
     log_level = Enum((0,10,20,30,40,50,'DEBUG','INFO','WARN','ERROR','CRITICAL'),
                     default_value=logging.WARN,
                     help="Set the log level by value or name.").tag(config=True)
-    def _log_level_changed(self, name, old, new):
+
+    @observe('log_level')
+    @observe_compat
+    def _log_level_changed(self, change):
         """Adjust the log level when log_level is set."""
+        new = change['new']
         if isinstance(new, string_types):
             new = getattr(logging, new)
             self.log_level = new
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
I like the solution. I had another idea too but I am travelling today. I might have some time to open a PR for it tonight (fr time) and try out this one.

SylvainCorlay:
I am not sure this is enough, because the new `@observe` will not work as expected with inheritance (independently of the question of the signature of the callback). In the following code, the handler in the derived class will not be called:

```Python
from traitlets import *

class Foo(HasTraits):
    
    bar = Int()
    
    @observe('bar')
    def handle_bar(self, change):
        print('base')
        
class Baz(Foo):
    
    def handler_bar(self, change):
        print('derived')

baz = Baz()
baz.bar = 1
```

While if you also decorate the derived class' method, you will get both 'base' and 'derived' printed... Hence, there is no natural way to reproduce the behavior of the `_bar_changed` handler.

minrk:
I'm more concerned with not breaking *existing* code (using magic method names):

```python
from traitlets import *

class Foo(HasTraits):

    bar = Int()

    @observe('bar')
    def _bar_changed(self, change):
        print('base')

class Baz(Foo):

    def _bar_changed(self, name, old, new):
        print('derived')
        super()._bar_changed(name, old, new)

b = Baz()
b.bar = 5
```

Having to register `@observe('bar')` on subclasses in order for override to work (it already does) doesn't seem too problematic to me (though frankly, I don't find the new API to be clearer or simpler than the old, since it requires more code to use).

minrk:
I added tests for adding/overriding observers in subclasses, and the behavior seems okay. The one case where behavior seems undefined is overriding a listener method *without* registering the overridden method. Currently, it appears that this just disables the original. That is surprising to me, but doesn't seem so problematic, if we just define in docs:

> to override an event-watcher, you must register the new method with `@observe`:

```python
class Subclass(Parent):
    @observe('foo'):
    def handle_foo(self, change):
        # overrides Parent.handle_foo
```

Here's a question, though:

Should The `@observe` decorator register an event-watching *object*, or should it just register the name, so that overrides in subclasses work? If it just registered the name, and then the object were looked up, overriding methods in subclasses would work in a much more predictable manner.

SylvainCorlay:
That is a good question.
As of now, `@observe` only wraps the call to the `register` method which does not take inheritance into account. 
We could also register a wrapper that does what you describe by calling

`getattr(change['owner'], self.name)(change)`.

rmorshea:
If we combine the `getattr(change['owner'], self.name)(change)` behavior with a use of `class_init` there could be a distinction between notifiers registered with `@observe` and those made with `HasTraits.observe`. Ultimately this would move `@observe` in the direction of `@default` in the sense that you could search through the `mro` to solve inheritance problems. I'm not sure that this would solve the problems with signatures though.

minrk:
I think it's appropriate for the signature-compat bit to be an *explicit* choice, rather than an automatic one. We don't need *new* classes or traits to be declared in a backward-compatible way. I think it's enough to be possible for existing classes to adopt the new API without breaking subclasses, and for that adoption to be done explicitly.

SylvainCorlay:
Another possibility would have been to overload `__call__` in the `ObserveHandler` object to also accept the old signature - but I am ok with this solution.
The tests that you just added seem to fail.

minrk:
@SylvainCorlay the test failure seems like it reveals an actual, but, but I'm not sure. In the following situation, what would you expect/intend to happen:

```python
from traitlets import HasTraits, Integer, observe

class A(HasTraits):
    b = Integer()
    @observe('b')
    def observer(self, change):
        print("A.observer", change)

class C(A):
    def observer(self, change):
        print("C.observer", change)


c = C()
c.b = 5
```

Currently, overriding a method that is an observer and not indicating that the new method is an observer *disables* the original method. If this is intended, I'll update the test to reflect that. If it should just be "Behavior undefined", I can remove it. Otherwise, there's a bug to be fixed.

SylvainCorlay:
Knowing how it works, I expect the observer to not be registered for `C`. This actually really suggests that we should ideally do something with static observers at the `class_init` stage.

minrk:
Okay, I updated the test to expect overriding a registered observer to disable the original.

PR #158, opened 2015-12-31
default initialization of configurable traits from os.environ

rmorshea:
Configurables hooks into a call from the metaclass to `setup_class` in order to setup default values from `os.environ`. This in turn calls `TraitType.set_default_from_envvar` to apply that default value to the TraitType if the key `'envvar'` is in its metadata.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -6,11 +6,14 @@
 
 from __future__ import print_function
 
+import os
 from copy import deepcopy
 import warnings
 
 from .loader import Config, LazyConfigValue, _is_section_key
-from traitlets.traitlets import HasTraits, Instance, observe, observe_compat, default
+from traitlets.traitlets import (HasTraits, BaseDescriptor, TraitType, Instance,
+                                observe, observe_compat, default)
+
 from ipython_genutils.text import indent, dedent, wrap_paragraphs
 from ipython_genutils.py3compat import iteritems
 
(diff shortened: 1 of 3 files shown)


traitlets/utils/index_table.py
```diff
@@ -0,0 +1,88 @@
+"""A dict-like table mapping keys to values based on indexing"""
```
minrk: Looks like this file was added by mistake.

traitlets/config/configurable.py
```diff
@@ -32,9 +33,47 @@ class MultipleInstanceError(ConfigurableError):
...
+                "{type} trait with name {name} expected "
+                "os.environ[{envvar!r}] to be set.".format(
+                    type=type(trait).__name__,
+                    envvar=vname,
+                    # repring here to add quotes in the formatted error message.
+                    name=repr(trait.name) if trait.name else "<anonymous>",
+                )
+            )
+        try:
+            value = eval(string, {}, cls.envvar_eval_context)
```
minrk: It seems like the string=> value conversion should depend on the trait type, and not use eval. `from_env` should probably be a method on each TraitType that supports coming from an env (e.g. only Unicode, numbers).

rmorshea:
I've included `TraitType.set_default_from_envvar` to apply default values from environment variables, but I have two questions about how I've implemented that hook:

+ Should I check a trait's metadata for `'envvar'`(as I am now) or an attribute `envvar` to determine whether it needs a default value to be set from an environment variable?

Having `envvar` be an attribute, and an explicit entry in `TraitType.__init__` makes it clear that environment variable defaults are a feature, however it might introduce confusion, because it is only a feature of traits attached to configurables.

+ Should I apply environment variable defaults through `setup_class` (as I am now) or `setup_instance`?

Using `setup_class` makes sense to me, and would be slightly more efficient, but I'm not really familiar with conventions surrounding how applications are typically configured (i.e. whether configurations should exist at the class level or their instances).

Carreau:
This seem to be stale a bit, and start to look like over engineering what traitlets can (and should do). 
This seem also like it's pretty easy to get by using `@default` function. 

So I'm curious if
  - we really want that ? 
  - is it worked on ? 

Because that seem like quite a bit of maintenance for a potentialy really low use case, and we are spread thin, so I'm leaning toward "no new feature" at that point.

rmorshea:
@Carreau, should probably bring up whether this is needed in #157. As far as `@default`, it really doesn't add anything different than `_<name>_default`. However what it does allow for is an `@envvar_default(name, envvar)` decorator in `configurable.py`. The decorator could set trait metadata, and pass the envvar string into the default generator.

rmorshea:
I'll close this one and make a new pr for `@envvar_default`.

PR #166, opened 2016-02-01
MAINT: Systematize access to TraitType._cross_validation_lock.

ssanderson:
Use a standard contextmanager for setting it everywhere (except
hold_trait_notifications, which is complex enough that I'm nervous about
refactoring it here without more input.).

Set _cross_validation_lock in TraitType.setup_instance, which is called
before __init__. Previously it was being implicitly initialized in
different places.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -486,15 +486,14 @@ class TraitType(BaseDescriptor):
         return getattr(self, 'make_dynamic_default', None)
 
     def instance_init(self, obj):
-        obj._cross_validation_lock = True
         # If no dynamic initialiser is present, and the trait implementation or
         # use provides a static default, transfer that to obj._trait_values.
-        if (self._dynamic_default_callable(obj) is None) \
-                and (self.default_value is not Undefined):
-            v = self._validate(obj, self.default_value)
-            if self.name is not None:
-                obj._trait_values[self.name] = v
-        obj._cross_validation_lock = False
+        with obj.cross_validation_lock:
+            if (self._dynamic_default_callable(obj) is None) \
+                    and (self.default_value is not Undefined):
+                v = self._validate(obj, self.default_value)
+                if self.name is not None:
+                    obj._trait_values[self.name] = v
 
     def get(self, obj, cls):
         try:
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -486,15 +486,14 @@ def _dynamic_default_callable(self, obj):
         return getattr(self, 'make_dynamic_default', None)
 
     def instance_init(self, obj):
-        obj._cross_validation_lock = True
```
ssanderson: This is the block that made me think this change was worth making.  If something inside the block fails, obj._cross_validation_lock will be stuck on `True`.  

I think this happens to be harmless here, because we're assuming that we're being called from `obj.__new__`, which means an uncaught exception will prevent a live instance to the invalid object from ever existing, but repeating this pattern elsewhere would result in hard-to-find bugs where the relative ordering of validators would cause very different behaviors.

ssanderson:
The other benefit here is that it provides a public API for blocking cross-validation.  For example, in serializable-traitlets, I have a notion of "example values" that are validated in a way similarly to default values, which requires holding the cross validation lock to ensure that we don't fire cross-validators when checking that an example is valid.  That's essentially the same thing that's happening in https://github.com/ipython/traitlets/pull/166/files#diff-8dc419d0669d108df4ac215bf816cc43R491, but it's that's a tenuous implementation detail to rely on in an external project.

SylvainCorlay:
Holding trait notifications also holds cross validation. In which case
would you want to notify and not cross-validate? It seems that it could
lead to inconsistent state.

ssanderson:
@SylvainCorlay I assume you're asking about the use-case I mentioned w/r/t validating example values?

In that case, the example isn't ever set on the instances itself: I just want to validate that it would pass TraitType checks **if it were set**.  The most straightforward way to do that, I think, is to call `TraitType._validate` while holding the cross validation lock.  Note that that's not equivalent to calling `TraitType.validate`, since `_validate` is what handles skipping validation of `None` if `allow_none` is True.

SylvainCorlay:
Looking back at this one 
 -in the context of nested `hold_trait_notifications` and `cross_validation_lock`.
 - about making the lock a public API
@ssanderson do what is your take on the second point?

PR #169, opened 2016-02-14
Add support for Enum classes (py34: enum.Enum)

jenisys:
Add support for Enum classes (py34: enum.Enum) as model for the data type description.

EXAMPLE:
```python
# -- SINCE: Python 3.4 (or install backport: pip install enum34)
import enum
from traitlets import HasTraits, UseEnum

class Color(enum.Enum):
    red = 1         # -- IMPLICIT: default_value
    blue = 2
    green = 3

class MyEntity(HasTraits):
    color = UseEnum(Color, default_value=Color.blue)

entity = MyEntity(color=Color.red)
entity.color = Color.green    # USE: Enum-value (preferred)
entity.color = "green"        # USE: name (as string)
entity.color = "Color.green"  # USE: scoped-name (as string)
entity.color = 3              # USE: number (as int)
assert entity.color is Color.green
```


Carreau:
Hi, thanks!

I have no objection about Enum, we already have one so maybe merge with it, instead of having 2 enums class ?

Testing and coverage are already taken care of by `.travis.yml` and codecov.io, from experience -1 on adding a `tox.ini` (and `.coverage` file ?) as they will get out of sync with the `.travis.yml` one.
But in any case, changing the testing and coverage behavior is better in a separate PR.

Also don't change copyright, the copyright date is the **first year** where something is copyrighted, 
is Basically meaning `2015-now`.

docs/source/conf.py
```diff
@@ -29,6 +29,12 @@
 _release = {}
 exec(compile(open('../../traitlets/_version.py').read(), '../../traitlets/_version.py', 'exec'),_release)
 
+# -- ENSURE: Build docs with current development version of traitlets.
+# REQUIRED-BY: autodoc (autoclass, etc.)
+HERE = os.path.dirname(__file__)
+TOPDIR = os.path.join(HERE, "..", "..")
+sys.path.insert(0, os.path.abspath(TOPDIR))
```
Carreau: Why is that ? It seem to currently work [on RTD](http://traitlets.readthedocs.org/en/stable/trait_types.html?highlight=caselessstrenum#traitlets.CaselessStrEnum)

docs/source/conf.py
```diff
@@ -69,7 +75,7 @@
 
 # General information about the project.
 project = u'traitlets'
-copyright = u'2015, The IPython Development Team'
+copyright = u'2016, The IPython Development Team'
```
Carreau: As said in other comment, please leave 2015.

setup.py
```diff
@@ -68,6 +68,8 @@
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
```
Carreau: Might be a good idea to just keep `Python::3`, we've started to do that other project.  Not sure though, don't really care.

minrk:
It looks like several unrelated files are being added here that shouldn't be (tox, sphinx-config changes, coverage).

docs/source/conf.py
jenisys: No, it doesn't. I got an error regarding the "UseEnum" autoclass. Therefore, I assume that it works with the installed traitlets version instead of using the current development version.

It may work for you, if you run "pip install -e ." (develop mode). If you use the recipe from above, it works on your workspace (even without installing the traitlets package in development mode).

docs/source/conf.py
jenisys: I will revert the change.

jenisys:
I added the ".coveragerc", "tox.ini" because they allowed me to test the pull-request without needing to use travis CI.

No problems, I can/will remove these files again.

jenisys:
1. REMOVED AGAIN: tox.ini, .coveragerc (development-support)
2. REVERTED: docs/source/conf.py (as before; but note that you need to install traitlets to build docs)
3. MINIMIZED CHANGES: setup.py

Remaining noise is caused by failed attempt to rebase the branch to the latest changes of master branch.

jenisys:
REPLACED-BY: pull #200 (clean pull-request)

PR #171, opened 2016-02-17
pass args and kwargs to setup_instance

rmorshea:
There is no way to initialize the attributes of a `HasDescriptors` instance, which descriptors are dependant on, if those attributes are determined by the args and kwargs passed to the `HasDescriptors` constructor.

The following attempt to initialize `HasDescriptors` attributes that descriptors are dependant on raises:

```python
class FooDescriptor(BaseDescriptor):

		foo = inst.foo # <-- this will raise an AttributeError
		
class HasFooDescriptors(HasDescriptors):

	fd = FooDescriptor()

	def __new__(cls, *args, **kwargs):
		inst = super(HasFooDescriptors, cls).__new__(cls, *args, **kwargs)
		inst.foo = kwargs.get('foo', None)
		
hfd = HasFooDescriptors(foo='bar')
# raises AttributeError because foo does not exist
# on the instance by the time instance_init is called
```

Now, because args and kwargs are passed to `setup_isntance`, this can be accomplished:

```python
class FooDescriptor(BaseDescriptor):

		foo = inst.foo # no error is raised
		
class HasFooDescriptors(HasDescriptors):

	fd = FooDescriptor()

	def setup_instance(self, *args, **kwargs):
		self.foo = kwargs.get('foo', None)
		super(HasFooDescriptors, self).setup_instance(*args, **kwargs)
		
hfd = HasFooDescriptors(foo='bar')
```

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -22,7 +22,7 @@ from traitlets import (
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
     ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default,
-    observe_compat,
+    observe_compat, BaseDescriptor, HasDescriptors,
 )
 from ipython_genutils import py3compat
 from ipython_genutils.testing.decorators import skipif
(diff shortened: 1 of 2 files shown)


minrk:
:+1:. Needs rebase because I've merged your other PRs. Can you add a test to exercise this?

rmorshea:
@minrk rebased

rmorshea:
@minrk rebased and tested.

minrk:
Thanks!

PR #173, opened 2016-02-17
pass *args to __new__

rmorshea:
Positional arguments are left out of the call to `super(...).__new__` here.

* I also change `kw` to `kwargs` in a couple places to keep things consistent.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -898,14 +898,14 @@ class HasDescriptors(py3compat.with_metaclass(MetaHasDescriptors, object)):
     """The base class for all classes that have descriptors.
     """
 
-    def __new__(cls, *args, **kw):
+    def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -898,14 +898,14 @@ class HasDescriptors(py3compat.with_metaclass(MetaHasDescriptors, object)):
...
-    def __new__(cls, *args, **kw):
+    def __new__(cls, *args, **kwargs):
         # This is needed because object.__new__ only accepts
         # the cls argument.
         new_meth = super(HasDescriptors, cls).__new__
         if new_meth is object.__new__:
             inst = new_meth(cls)
         else:
-            inst = new_meth(cls, **kw)
+            inst = new_meth(cls, *args, **kwargs)
```
rmorshea: Was there a reason *args was left out?

PR #183, opened 2016-03-06
The hold_trait_notifications context manager should use set_trait

SylvainCorlay:
The `hold_trait_notifications` context manager should use `set_trait` instead of `setattr` to be able to rollback changes on read-only attributes.


rmorshea:
Seems good to me.

This solves the case where `set_trait` (instead of `setattr`) is used inside `hold_trait_notifications` right?

SylvainCorlay:
That is right.

PR #185, opened 2016-03-10
Allow Applications to pick if they should allow config-file failure

minrk:
Some applications (most?) should halt on failing to load config, this lets them choose by setting `raise_config_file_errors = True`.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -21,7 +21,7 @@ from traitlets.config.loader import (
 )
 
 from traitlets.traitlets import (
-    Unicode, List, Enum, Dict, Instance, TraitError, observe, observe_compat, default,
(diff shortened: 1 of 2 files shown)


traitlets/config/application.py
```diff
@@ -149,6 +149,9 @@ def _classes_inc_parents(self):
     # the argv used to initialize the application
     argv = List()
 
+    # Whether failing to load config files should prevent startup
+    raise_config_file_errors = Bool(False)
```
SylvainCorlay: no need to specify False.

minrk: I think it's clearer to be explicit, sometimes.

SylvainCorlay:
LGTM

PR #188, opened 2016-03-14
demote super().__init__ argument warning to deprecation

minrk:
because stable IPython (4.1) shows *a lot* of these (on Python 3), which is probably annoying. It is deprecated behavior, after all.


SylvainCorlay:
:+1:

minrk:
Thanks!

PR #190, opened 2016-03-18
Allow more than 'change' notification types

tjragan:
Currently the notification type 'change' is hard-coded into _notify_trait, and the default is to set observers to listen for notifications of type 'change'.  Although all notification types in the IPython codebase seem to be type 'change', this means that other types can't currently be used.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1084,13 +1084,13 @@ class HasTraits(py3compat.with_metaclass(MetaHasTraits, HasDescriptors)):
                     for change in changes:
                         self.notify_change(change)
 
-    def _notify_trait(self, name, old_value, new_value):
+    def _notify_trait(self, name, old_value, new_value, type='change'):
         self.notify_change({
             'name': name,
             'old': old_value,
             'new': new_value,
             'owner': self,
-            'type': 'change',
+            'type': type,
         })
 
     def notify_change(self, change):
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@tjragan this is because people should not use `_notify_trait` but `_notify_change`. `_notify_trait` should be completely removed.

SylvainCorlay:
The only reason why `_notify_trait` is still around is because some libraries were effectively overloading it even though it was private API.

tjragan:
Ok, sounds good.  What about setting the observe type default to All instead of 'change'?

SylvainCorlay:
@tjragan yeah, this is what we did originally when we introduced this new API but we eventually decided to default to change for backward compatibility.

Otherwise, old handlers registered with `on_trait_change` listening to all trait attributes would get events they don't know how to handle when we add trait attributes that emit other event types.

tjragan:
Right, so is that a permanent decision?  If so, can we make the string 'All' work like the object All?  Otherwise I'll have to import the All object throughout my code.

SylvainCorlay:
I have not thought of the 'all' stting but is it an issue to import the
`All` class?

Defaulting to 'change' is likely to be permanent. Is that a problem?

The only way to change this behavior in a backward compatible way would be
to start requiring a value (and only warning the user when they don't), but
given the amount of code in Jupyter that depends on traitlets, deprecation
cycles are going to be very long.

Sylvain

SylvainCorlay:
I have not thought of the 'all' stting but is it an issue to import the `All` class?

The defaulting to 'change' is likely to be permanent. Is that a problem?

The only way to change this behavior in a backward compatible way would be to start requiring a value (and only warning the user when they don't), but given the amount of code in Jupyter that depends on traitlets, deprecation cycles are going to be very long.

tjragan:
Obviously I don't want to break any legacy code, and **definitely** don't want to mess with Jupyter!  It just feels odd to import the All object in core parts of my system that don't import any other libraries.  It also feels odd that the default is not to observe everything, BUT, I think you decision is the right one.

In my case, I think I'll try subclassing HasTraits to change the behaviour and not mess with the rest of the community.

Thanks for the help!

-TJ

PR #182, opened 2016-03-05
Remove depedency towards ipython_genutils.

lucasg:
Hi,

**ipython_genutils** is marked for deprecation, you might as well incorporate only the functionalities needed and remove the import.  Traitlets is much better as a stand-alone library anyway.
The 9 commits incrementally add external utilities from ipython_genutils into the traitlets package (every commits pass the nosetest tests on my machine). 
Note : I've copied py3compat in its entirety, but only a portion is useful. You might need to trim it down, or abstracting away using six.

Cheers,

lucasg

### Results from Travis 

https://travis-ci.org/lucasg/traitlets/builds/113952990

### Results from tests (on my machine)
```
`(ipython_env) [lucasg traitlets]$ python --version
Python 3.5.1

(ipython_env) [lucas traitlets]$ pip freeze
coverage==4.0.3
coveralls==1.1
decorator==4.0.9
docopt==0.6.2
ipython==5.0.0.dev0
ipython-genutils==0.2.0.dev0
nose==1.3.7
path.py==8.1.2
pexpect==4.0.1
pickleshare==0.6
prompt-toolkit==0.59
ptyprocess==0.5.1
Pygments==2.1.3
requests==2.9.1
simplegeneric==0.8.1
six==1.10.0
traitlets==4.2.0.dev0
wcwidth==0.1.6
wheel==0.29.0

(ipython_env) [lucas traitlets]$ python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': 'eb2f497',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': REDACTED,
 'ipython_version': '5.0.0.dev',
 'os_name': 'posix',
 'platform': 'Linux-4.1.18-2-MANJARO-x86_64-with-glibc2.3.4',
 'sys_executable': REDACTED,
 'sys_platform': 'linux',
 'sys_version': '3.5.1 (default, Dec  7 2015, 12:58:09) \n[GCC 5.2.0]'}




(ipython_env) [lucasg traitlets]$ nosetests --exe --with-coverage --cover-package traitlets traitlets
...................................................................................................................................................................................S................S............................................................................................
Name                               Stmts   Miss  Cover   Missing
----------------------------------------------------------------
traitlets.py                           3      0   100%   
traitlets/_version.py                  2      0   100%   
traitlets/config.py                    3      0   100%   
[...]

PR #194, opened 2016-03-23
Switch over nose2

lucasg:
Hi,

As discussed in PR #182, it might be a better idea to migrate the nose testing to `nose2`, since nose does not seem to be maintained any more. This  pull request does not add any new functionalities or tests, it is just simple refactoring. As a bonus, `traitlets` does not need to depend on `ipython_genutils.skipif` since this method is already implemented in `unittest`.

Results from Travis : https://travis-ci.org/lucasg/traitlets/builds/118092234

NB : nose2 output warnings for the test_loader tests, since `TestKeyValueCL` and `TestArgParseKVCL` consider unknwon input parameters as potentially erroneous (see `loader.py:419`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
     - 3.3
 sudo: false
 install:
+    - pip install . nose2
     - pip install . cov-core
 script:
     - nose2 --with-coverage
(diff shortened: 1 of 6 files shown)


minrk:
Great, thanks! For the record, I would also be fine with pytest.

PR #200, opened 2016-03-26
Provide UseEnum as Traitlets class to support enum34 Enum classes 

jenisys:
CLEAN-REPLACEMENT FOR: pull #169

Add support for Enum classes (py34: enum.Enum) as model for the data type description (instead of string-based enums).

EXAMPLE:
```python
# -- SINCE: Python 3.4 (or install backport: pip install enum34)
import enum
from traitlets import HasTraits, UseEnum

class Color(enum.Enum):
    red = 1         # -- IMPLICIT: default_value
    blue = 2
    green = 3

class MyEntity(HasTraits):
    color = UseEnum(Color, default_value=Color.blue)

entity = MyEntity(color=Color.red)
entity.color = Color.green    # USE: Enum-value (preferred)
entity.color = "green"        # USE: name (as string)
entity.color = "Color.green"  # USE: scoped-name (as string)
entity.color = 3              # USE: number (as int)
assert entity.color is Color.green
```

### DIFF
--- a/docs/source/trait_types.rst
+++ b/docs/source/trait_types.rst
@@ -101,6 +101,8 @@ Miscellaneous
 
 .. autoclass:: CaselessStrEnum
 
+.. autoclass:: UseEnum
+
 .. autoclass:: TCPAddress
 
 .. autoclass:: CRegExp
--- a/setup.py
+++ b/setup.py
@@ -83,6 +83,10 @@ install_requires = setuptools_args['install_requires'] = [
 
 extras_require = setuptools_args['extras_require'] = {
     'test': ['nose2'],
+    # -- SUPPORT UNIFORM-WHEELS: Extra packages for Python 2.7, 3.3
+    # SEE: https://bitbucket.org/pypa/wheel/ , CHANGES.txt (v0.24.0)
+    ':python_version=="2.7"': ["enum34"],
+    ':python_version=="3.3"': ["enum34"],
 }
 
 if 'setuptools' in sys.modules:
--- a/traitlets/tests/test_traitlets_enum.py
+++ b/traitlets/tests/test_traitlets_enum.py
@@ -1,0 +1,181 @@
+# -*- coding: UTF-8 -*-
+# pylint: disable=missing-docstring, too-few-public-methods
+"""
+Test the trait-type ``UseEnum``.
+"""
+
+import unittest
+import enum
+from ipython_genutils.py3compat import string_types
+from traitlets import HasTraits, TraitError, UseEnum
+
+
+# -----------------------------------------------------------------------------
+# TEST SUPPORT:
+# -----------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


minrk:
Sorry for neglecting this one. Do you think it makes sense to add enum support to the existing `Enum` traitlets, rather than making a new, apparently equivalent traitlet?

jenisys:
I looked at it at first, but I think both solve a slightly different problem. Therefore, I think they cannot be easily merged (without causing problems with existing code). But if you have a good idea how to do it ... (maybe I overlooked something).

jenisys:
I looked at both implementations again. You should be able to merge the two implementations. This would require to always use an enum.Enum class internally. In addition, the implementation of most methods would look like in the current ``UseEnum`` class (which is am approximate superset of the current ``traitlets.Enum`` class). If values (as list<string>) are used in the constructor, an "anonymous Enum class" would be created (instead of using the current implementation). 

I can prepare such a merged ``traitlets.Enum`` class. 
Should I provide an own pull request or use the current one ?

minrk:
@jenisys I think it would be great to always use enum internally, as long as the existing calls to Enum produce enums that still work as expected.

jenisys:
OK, I provide a pull-request that satisfies both worlds.

jenisys:
I stumbled over a problem or at least a risk when the enum implementation is merged.
The following uses of ``traitlets.Enum`` are at least problematic (or no longer valid):

```python
# -- FILE: traitlets/config/application.py:156
# CASE: Mixed numbers and string
# BUT: Easily solvable (in traitlets), but may break compatibility of existing use cases.
  # OLD: 
  log_level = Enum((0,10,20,30,40,50,'DEBUG','INFO','WARN','ERROR','CRITICAL'),
  # NEW:
  log_level = Enum([('NOTSET', 0), ('DEBUG', 10), ('INFO', 20), ('WARN', 30), ('ERROR', 40), ('CRITICAL', 50)], ...)

# -- FILE: IPython/core/interactiveshell.py:182
# CASE: Only numbers
autocall = Enum((0,1,2), default_value=0, config=True, help="...")

```

minrk:
We can't do something that would break existing code, so if we can't at least get a shim in that will continue to work, let's keep the new Enum traitlet separate, as you initially proposed, and we can start directing people on the old implementation to the new one if it's better (or keep both). Alternately, if it can be folded into the existing Enum but not force it to be *always* an enum, preserving the existing behavior, that would work, too.

jenisys:
After exploring it, I think two separate classes are the better solution (with less risk). By using the merged approach (even with raw values), especially the constructors gets more complicated. Therefore, I would vote for the first solution: Two (slightly) different concepts, each with its class and with a dedicated scope.

minrk:
Makes sense, thanks for exploring the options!

PR #209, opened 2016-04-20
deprecate `class_traits` and `class_trait_names` + add event searchers

rmorshea:
+ creates deprecation warnings to `class_traits` and `class_trait_names` and shifts them to depend on the now class based `traits` and `trait_names`.
+ adds `events` and `class_own_events` as a way find to the event handlers associated with a particular trait or simply the event handlers for the class in general.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1401,7 +1401,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
             return trait.metadata.get(key, default)
 
     @classmethod
-    def class_own_events(cls, name):
+    def class_own_trait_events(cls, name):
         """Get a dict of all event handlers defined on this class, not a parent.
 
         Works like ``event_handlers``, except for excluding traits from parents.
@@ -1411,7 +1411,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
                 if getattr(sup, n, None) is not e}
 
     @classmethod
-    def events(cls, name=None):
+    def trait_events(cls, name=None):
         """Get a ``dict`` of all the event handlers of this class.
 
         Parameters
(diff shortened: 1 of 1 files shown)


minrk:
What was the difference between `.traits()` on an instance and `.class_traits()` on its class? Would they ever have different results?

rmorshea:
Nope. The only difference being that `class_traits` is unbound.

traitlets/traitlets.py
```diff
@@ -1362,49 +1369,71 @@ def traits(self, **metadata):
...
-        else:
-            getattr(cls, name).set(self, value)
+        Works like ``event_handlers``, except for excluding traits from parents.
+        """
+        sup = super(cls, cls)
+        return {n: e for (n, e) in cls.events(name).items()
+                if getattr(sup, n, None) is not e}
+
+    @classmethod
+    def events(cls, name=None):
```
minrk: Since these methods get defined on all HasTraits derivatives, we should be careful about adding too many generic top-level names. Perhaps `trait_events` (and `class_own_trait_events` above)?

rmorshea: I guess the main reason I didn't do that is because it seemed unnecessary to call it `trait_events` as opposed to just `events`, but that's reasonable if you think that might be too generic.

minrk: I think it would be better, since we have to think of how it affects users - if we define an `events` method on the base, that means no HasTraits can ever define its own `events` method or attribute. We should be very conservative about what names we are reserving on a base class, and try to make them trait-specific where we can. While `trait_events` may seem redundant on HasTraits, it isn't on `InteractiveShell` or other HasTraits descendants, where it really matters.

rmorshea:
@minrk names changed to `trait_events` and `class_own_trait_events`

minrk:
Thanks!

Carreau:
This is wrong you changed an API, traits suddenly goes from a instance method to a class method you can't do that.

PR #213, opened 2016-04-27
absolute_import in configurable

### LINKED ISSUE #212: ironpython - absolute_import
The import from __future__ in config/configurable.py needs to add , absolute_import

It's causing an error when loading from jupyter_client under IronPython (2.7.5).  I think this may be the only change needed in this lib, somebody want to add it for me and save me making a one-word pull request?

rmorshea:
closes #212 - import from future in config/configurable.py needs to add absolute_import


minrk:
Merging, but registering a comment that this shouldn't actually have any effect because there are no imports in this module that should be affected by `absolute_import`.

PR #226, opened 2016-05-17
Warn if known constructor args are passed as metadata to .tag()

Carreau:


traitlets/traitlets.py
```diff
@@ -637,6 +637,11 @@ def tag(self, **metadata):
 
         >>> Int(0).tag(config=True, sync=True)
         """
+        maybe_constructor_keywords = set(metadata.keys()).intersection({'help','allow_none', 'read_only_default_value'});
+        if maybe_constructor_keywords:
+            warn('The following attributes are set in using `tag`, but seem to be constructor keywords arguments: %s ',
+                    maybe_constructor_keywords, UserWarning, stacklevel=2);
```
minrk: too much js: semicolons

Carreau: OOps.

Carreau: (hopefully it's not too much matlab), I was also missing a `%` for string formatting.

PR #228, opened 2016-05-18
Allow to configure whether Application abort on bad config.

Carreau:
Does that through an env variable:
TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR

Might want to expose that at higher level so that application can
make this setting on a per-application bassis.

Related to https://github.com/ipython/ipython/issues/8927

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -62,6 +62,17 @@ subcommand 'cmd', do: `{app} cmd -h`.
 # Application class
 #-----------------------------------------------------------------------------
 
+
+
+_envvar = os.environ.get('TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR','')
+if _envvar.lower() in {'1','true'}:
+    TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR = True
+elif _envvar.lower() in {'0','false',''} :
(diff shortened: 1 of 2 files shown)


Carreau:
Needs documentation (if accepeted)

traitlets/config/application.py
```diff
@@ -151,7 +151,7 @@ def _classes_inc_parents(self):
     argv = List()
 
     # Whether failing to load config files should prevent startup
-    raise_config_file_errors = Bool(False)
+    raise_config_file_errors = Bool(os.environ.get('TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR', '').lower() == 'true')
```
ssanderson: Any argument for making this a dynamic default so that the default value gets read at application construction time rather than at import time?  I imagine that would make this easier to test, if nothing else.

Carreau: The others ENV variables are module wise and `_load_config_files` is a class method, so it is use without instance (like in IPython/application.py) so I'm unsure.

Carreau:
TODO: match the API with the IPython equivalent PR.

Carreau:
Updated to match IPtyhon API, and documented.

minrk:
Thanks, @Carreau!

Issue #221, opened 2016-05-17
observe should raise with better error if wrong argument passed. 


Carreau:
Do you see the mistake : 

```python
    readline_parse_and_bind = List(default_value=[
            'tab: complete',
            '"\C-l": clear-screen',
            'set show-all-if-ambiguous on',
            '"\C-o": tab-insert',
            '"\C-r": reverse-search-history',
            '"\C-s": forward-search-history',
            '"\C-p": history-search-backward',
            '"\C-n": history-search-forward',
            '"\e[A": history-search-backward',
            '"\e[B": history-search-forward',
            '"\C-k": kill-line',
            '"\C-u": unix-line-discard',
        ]).tag(config=True)
    
    _custom_readline_config = False
   
    @observe(readline_parse_and_bind)
    def _readline_parse_and_bind_changed(self, change):
        # notice that readline config is customized
        # indicates that it should have higher priority than inputrc
        self._custom_readline_config = True
```

it took me 1/2 an hour to figure out I was missing a quote, because it took me 1/2 an hour to figure out which observe was wrong:

```python
Traceback (most recent call last):
  File "/Users/bussonniermatthias/anaconda/bin/ipython", line 9, in <module>
    load_entry_point('ipython', 'console_scripts', 'ipython')()
  File "/Users/bussonniermatthias/dev/ipython/IPython/__init__.py", line 119, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "/Users/bussonniermatthias/anaconda/lib/python3.5/site-packages/traitlets/config/application.py", line 595, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-111>", line 2, in initialize
  File "/Users/bussonniermatthias/anaconda/lib/python3.5/site-packages/traitlets/config/application.py", line 74, in catch_config_error
    return method(app, *args, **kwargs)
  File "/Users/bussonniermatthias/dev/ipython/IPython/terminal/ipapp.py", line 305, in initialize
    self.init_shell()
  File "/Users/bussonniermatthias/dev/ipython/IPython/terminal/ipapp.py", line 321, in init_shell
    ipython_dir=self.ipython_dir, user_ns=self.user_ns)
  File "/Users/bussonniermatthias/anaconda/lib/python3.5/site-packages/traitlets/config/configurable.py", line 405, in instance
    inst = cls(*args, **kwargs)
[...]

Issue #215, opened 2016-05-16
`observe` should allow any iterable containers. 


Carreau:
Not just `list`, and `tuple`,  thinks like `set`... should work. 

More generally you should fix `parse_notifier_name`. 

And technically it should allow any container  that implement [`__contains__`](https://docs.python.org/3.6/reference/datamodel.html#object.__contains__) I guess. And `parse_notifier_name` should not blindly return nothing (ie `None`) if the type of argument is not one of the accepted argument. It shoudl raise a **meaningful** error.

PR #230, opened 2016-05-23
register TraitEventHandlers with tags

### LINKED ISSUE #219: don't use `assert`.
Assert is meant optimised out by `-O` Python, which means that it is not equivalent to raise. 
Should be replaced by actual exceptions raising. 

Was introduced in #200, slipped through review.

rmorshea:
Allows the use of metadata as a way to register trait events

``` python
class A(HasTraits):
    foo = Int().tag(bar=True)
    @observe(tags={'bar':True})
    def _bar_observer(self, change):
        print 'change from bar tagged trait'

a = A()

a.foo = 1
```

The same pattern also works for `@validate` and `@default`.
## General:
- More elegant than #111.
- Closes #219, #215, #221 
## Under The Hood:
1. `default`, `validate`, `observe` (method and decorator), `unobserve`, and `unobserve_all` have a new `'tags'` keyword that provides metadata for gathering groups of traits for observation or for removing observers (depending on the method).
   - The metadata provided via `'tags'` does not create a dynamic reference to traits - e.g. if the metadata of a trait is modified, the tagged registration of observers to that trait will not be recomputed. Thus, in edge cases where metadata might be modified, or new traits are added, the timing of a call to one of the above methods should be closely considered.
3. New functionality for `unobserve_all`
   - Previously `unobserve_all` either deleted all notifiers, or all notifiers of a given name.
   - Now, it's taken a feature previously belonging to `HasTraits.unobserve` when its `'handler'` argument was `None`. Thus `unobserve_all` can remove the handlers of a given name and type.
   - Additionally, `unobserve_all` can now remove all handlers of a specified type across all traits.
   - Calls to `unobserve` where `'handler'` is `None` are redirected to `unobserve_all` with a warning.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -988,7 +988,7 @@ class ObserveHandler(TraitEventHandler):
 
     caches_instance_resources = True
 
-    def __init__(self, names, tags, type):
+    def __init__(self, names, type, tags):
         super(ObserveHandler, self).__init__(names, tags)
         self.type = type
 
@@ -1376,7 +1376,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
         else:
             self.observe(_callback_wrapper(handler), names=name)
 
-    def observe(self, handler, names=None, tags=None, type='change'):
+    def observe(self, handler, names=None, type='change', tags=None):
         """Setup a handler to be called when a trait changes.
 
         This is used to setup dynamic notifications of trait changes.
@@ -1411,10 +1411,10 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
         if isinstance(handler, ObserveHandler):
             self._add_notifiers(handler, names, type)
         else:
-            event = ObserveHandler(names, tags, type)
+            event = ObserveHandler(names, type, tags)
             event(handler).register(self)
 
-    def unobserve(self, handler, names=None, tags=None, type='change'):
+    def unobserve(self, handler, names=None, type='change', tags=None):
         """Remove a trait change handler.
 
         This is used to unregister handlers to trait change notifications.
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -233,7 +233,7 @@ def parse_notifier_names(owner, names=None, tags=None):
         names = []
     elif names is All or isinstance(names, six.string_types):
         names = [names]
-    elif not names or All in names:
+    elif (not names and not tags) or All in names:
         names = [All]
     else:
         for n in names:
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -934,30 +934,14 @@ class EventHandler(BaseDescriptor):
         else:
             return other == self.func
 
-    def __hash__(self):
-        """This hash is ONLY safe for hash comparisons to other EventHandlers!"""
-        if self.func is None:
-            return id(self)
-        else:
-            return hash(self.func)
 
-
-class TraitEventHandler(EventHandler):
-
-    metadata = {}
-    trait_names = []
-
-    def __init__(self, names, tags):
-        if (not names and names is not None) and not tags:
-            raise TypeError("Please specify at least one trait name or tag.")
-        self.trait_names = parse_notifier_name(names)
-        self.metadata = tags or {}
-
-
-class ObserveHandler(TraitEventHandler):
+class ObserveHandler(EventHandler):
 
     def __init__(self, names, type, tags):
-        super(ObserveHandler, self).__init__(names, tags)
+        if not names and not tags:
+            raise TypeError("Please specify at least one trait name or tag.")
+        self.trait_names = names
+        self.metadata = tags
         self.type = type
 
     def instance_init(self, obj):
@@ -965,14 +949,26 @@ class ObserveHandler(TraitEventHandler):
         obj.observe(self, self.trait_names, self.type, self.metadata)
 
 
-class ValidateHandler(TraitEventHandler):
+class ValidateHandler(EventHandler):
+
+    def __init__(self, names, tags):
+        if not names and not tags:
+            raise TypeError("Please specify at least one trait name or tag.")
+        self.trait_names = names
+        self.metadata = tags
 
     def instance_init(self, obj):
         super(ValidateHandler, self).instance_init(obj)
         obj._register_validator(self, self.trait_names, self.metadata)
 
 
-class DefaultHandler(TraitEventHandler):
+class DefaultHandler(EventHandler):
+
+    def __init__(self, names, tags):
+        if not names and not tags:
+            raise TypeError("Please specify at least one trait name or tag.")
+        self.trait_names = names
+        self.metadata = tags
 
     def class_init(self, cls, name):
         super(DefaultHandler, self).class_init(cls, name)
@@ -1218,7 +1214,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
 
             if isinstance(c, _CallbackWrapper):
                 c = c.__call__
-            elif isinstance(c, TraitEventHandler) and c.name is not None:
+            elif isinstance(c, EventHandler) and c.name is not None:
                 c = getattr(self, c.name)
 
             c(change)
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -451,7 +451,7 @@ class TraitType(BaseDescriptor):
     def subclass_init(self, cls):
         if '_%s_default' % self.name in cls.__dict__:
             method = getattr(cls, '_%s_default' % self.name)
-            cls._trait_default_generators[self.name] = method
+            cls._register_default_generators(method, self.name)
 
     def __init__(self, default_value=Undefined, allow_none=False, read_only=None, help=None,
         config=None, **kwargs):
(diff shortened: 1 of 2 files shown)


traitlets/traitlets.py
```diff
@@ -1199,7 +1327,7 @@ def on_trait_change(self, handler=None, name=None, remove=False):
         else:
             self.observe(_callback_wrapper(handler), names=name)
 
-    def observe(self, handler, names=All, type='change'):
+    def observe(self, handler, names=None, tags=None, type='change'):
```
minrk: The insertion of an argument makes this a backward-incompatible change.

rmorshea: @minrk I've inserted a warning, so it could be phased in. The alternative is to use `**kwargs`.

rmorshea:
@minrk waiting on #240

traitlets/traitlets.py
minrk: Yeah, I'd get in the habit of putting new args at the end. Most of these should be treated (in documentation, etc.) as keyword-only, even if they *can* be positional.

traitlets/traitlets.py
```diff
@@ -883,33 +907,131 @@ def __get__(self, inst, cls=None):
...
+            return other == self.func
+
+
+class ObserveHandler(TraitEventHandler):
 
-    def __init__(self, names, type):
-        self.trait_names = names
+    caches_instance_resources = True
+
+    def __init__(self, names, tags, type):
```
Carreau: Change of API also, types moves from 3rd to 4th.

traitlets/traitlets.py
```diff
@@ -883,33 +907,131 @@ def __get__(self, inst, cls=None):
...
+                names = list(matched)
+        else:
+            names = self.event_of(inst)
+        inst._register_validator(self, names)
 
 
-class DefaultHandler(EventHandler):
+class DefaultHandler(TraitEventHandler):
 
-    def __init__(self, name):
```
Carreau: change of API once more. from `self, name`, to `self, names, tags`.

traitlets/traitlets.py
```diff
@@ -1231,11 +1359,19 @@ def observe(self, handler, names=All, type='change'):
...
+        if names is None and tags is None:
+            names = All
+        if isinstance(handler, ObserveHandler):
+            self._add_notifiers(handler, names, type)
+        else:
+            event = ObserveHandler(names, tags, type)
+            event(handler).register(self)
 
-    def unobserve(self, handler, names=All, type='change'):
+    def unobserve(self, handler, names=None, tags=None, type='change'):
```
Carreau: Insertion of `tags` here too is a change of API.

traitlets/traitlets.py
rmorshea: This was a conscious change in order to keep the use of trait names and tags in `TraitEventHandlers` and their decorators uniform. It wouldn't be difficult to limit `@default` to one trait name, but I see no reason to limit it, especially if the framework to allow it exists right under it.

And though `DefaultHandler.__init__` may be different `@default`, the "public" constructor, is still backward compatible. This change is only a breakage for people who have overridden `DefaultHandler.__init__`, not `@default`.

Carreau: `DefaultHandler` is not marked as private, it is importable from top level packages: `from traitlets import default, DefaultHandler` works, 
and despite being a positional argument  you can pass it as a kwarg as for any Python functions not written in C. 

Sidenote, **positional only** argument in Python **do** exist, they are marked by a slash in the function signature:

```
In [3]: input?
Signature: input(prompt=None, /)
```

This is possible only with C Python api, so we can consider that all arguments name in functions signature in python are part of a function api. 

We can check that:  `DefaultHandler(name='foo')` currently work on traitlets stable, and break on with this PR.

Thus regardless of the **intent** of whether this was public or not it **is** a change of API. and it **does** break backward compatibility.

Carreau:
@rmorshea Sorry if I can seem annoying, I understand that I might be looking like I'm bikeshedding. 
Though, I have been confronted to this kind of things before. These are things that I learned the hard way, and these are difficult things to get use to.  

Also @rmorshea You are doing a fantastic job, and understand that if we reach this kind of level of review in a pull request is because you are doing really great, so we are pushing you into even harder things !

I hope you are not taking anything personally, and if you find there is too much pressure, or that we are too hard, feel free to ask for help. I know it can be hard to have a PR languishing for a long time.

traitlets/traitlets.py
rmorshea: @Carreau, I used "" to imply that it `DefaultHandler` was sudo-private (sorry for the confusion). In other words, most users won't encounter `DefaultHandler`, just `@default`. I was just trying to make the argument that this limits the extent of the damage the change causes.

The reason I say that this is a backward compatible change though is that any valid use of `DefaultHandler.__init__` that could have been made prior to the change, is still valid now. This is because `parse_notifer_name` is used to handle values of `names = 'something'` in `TraitEventHandler.event_of`, which is I use to determine which traits a handler should be registered to.

rmorshea:
@Carreau I totally understand. The fact that I'm not working heavily with traitlets myself means I might not have the greatest foresight for the problems that will crop up in the future. So I value whatever input you or anyone else has.

rmorshea:
... adding tests for `@validate` and `@default` via tags

traitlets/traitlets.py
rmorshea: This has been changed, though the diff is still showing.

Carreau:
Don't worry that's something you'll learn through time !

traitlets/traitlets.py
```diff
@@ -820,10 +846,10 @@ def validate(*names):
     exiting the ``hold_trait_notifications` context, and such changes may not
     commute.
     """
-    return ValidateHandler(names)
+    return ValidateHandler(names, tags=kwargs.get('tags', {}))
```
minrk: Should probably raise on unrecognized kwargs

minrk:
> should HasTraits.trait_events return these static handlers in addition to the methods it already does?

I don't think there is a reason to distinguish between `@observe` and `observe`. What prompted you to maintain this distinction? When would it be relevant?

rmorshea:
@minrk rebased. Not sure how trivial the hit to patch coverage is yet...

rmorshea:
@minrk, any chance we can fast-track a review of this PR? I'm hoping to integrate this functionality in my work for matplotlib soon.

ankostis:
In general it would be nice to start having some 5.x.x pre-releases, of possible.

traitlets/traitlets.py
```diff
@@ -820,7 +840,7 @@ def compatible_observer(self, change_or_name, old=Undefined, new=Undefined):
     return compatible_observer
 
 
-def validate(*names):
+def validate(*names, **kwargs):
```
minrk: Why `**kwargs` instead of explict `tags=None`?

traitlets/traitlets.py
```diff
@@ -1293,20 +1374,52 @@ def unobserve(self, handler, names=All, type='change'):
             The type of notification to filter by. If All, the specified handler
             is uninstalled from the list of notifiers corresponding to all types.
         """
-        names = parse_notifier_name(names)
-        for n in names:
-            self._remove_notifiers(handler, n, type)
+        if isinstance(tags, six.string_types):
+            warn("new argument 'tags' introduced: use 'type' as keyword",
```
minrk: since `tags` didn't get inserted before `type`, I'm not sure why this warning.

traitlets/traitlets.py
```diff
@@ -995,6 +1066,7 @@ def setup_instance(*args, **kwargs):
         self._trait_values = {}
         self._trait_notifiers = {}
         self._trait_validators = {}
+        self._static_trait_notifiers = []
```
minrk: What is the new `_static_trait_notifiers` for?

traitlets/traitlets.py
```diff
@@ -1342,6 +1455,14 @@ def add_traits(self, **traits):
         """Dynamically add trait attributes to the HasTraits instance."""
         self.__class__ = type(self.__class__.__name__, (self.__class__,),
                               traits)
+
+        events = self._static_trait_notifiers
+        events.extend(self.trait_events().values())
+
+        for e in events:
+            if e.caches_instance_resources:
```
minrk: What does `e.caches_instance_resources` mean?

traitlets/traitlets.py
rmorshea: Tags are actually allowed for validators.

traitlets/traitlets.py
rmorshea: I'm working on a rewrite to remove this. It has to do with complications associated with re-registering tagged notifiers to new traits created via `add_traits`.

traitlets/traitlets.py
rmorshea: Yes, this should be removed.

traitlets/traitlets.py
rmorshea: Similarly to `_static_trait_notifiers` I need to rethink this and make the implementation cleaner.

rmorshea:
@minrk, I've drastically simplified things. Still need to review the accuracy of the docstrings, and edit this PR's description to reflect changes.

rmorshea:
After much simplifying, this should be ready to be merged. 

cc: @minrk

rmorshea:
I think we can push this back to 5.1 since it's not a breaking change.

Carreau:
Closing as state for 3 years. Labeled accordingly, feel free to reopen.

PR #233, opened 2016-05-26
ENH: Use own logger for the package rather than the root logger.

rkern:
Libraries should not attempt to configure the root logger. The standard best practice is to use a namespaced logger for the package and to add a `NullHandler` to it.


rkern:
Thanks!

PR #236, opened 2016-05-27
add attr-access to change dicts

### LINKED ISSUE #217: observe's handler change should likely be an object.
It always have the same field, and is annoying to access through keys.
Maybe a NameTuple for immutability ? not sure if you can access a named tuple with square brackets.

minrk:
it's still a dict, so everything should be 100% backward-compatible.

include yet-another implementation of bunch in utils.bunch

closes #217

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -173,7 +173,7 @@ class Application(SingletonConfigurable):
     @observe_compat
     def _log_level_changed(self, change):
         """Adjust the log level when log_level is set."""
-        new = change['new']
+        new = change.new
         if isinstance(new, six.string_types):
             new = getattr(logging, new)
             self.log_level = new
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
:+1: this is almost as cool as JavaScript :stuck_out_tongue:

SylvainCorlay:
It would be nice to do the same thing with the `proposal` dictionary for cross validation.

SylvainCorlay:
@minrk lgtm except for the comment above.

minrk:
@SylvainCorlay good call, done. Are there any other dict-events at this point?

SylvainCorlay:
I don't think so.

SylvainCorlay:
Awesome, this is a significant improvement IMO.

jasongrout:
Nice! Thanks!

Carreau:
Thanks ! That's much more readable, and I'm sure completers will allow to complete that easily.

jasongrout:
In order for completers to work, we should add a `__dir__` to the Bunch: https://docs.python.org/2/library/functions.html#dir

minrk:
@jasongrout added keys to dir in #238. It doesn't mess up completions of dict methods since the keys are just added.

PR #237, opened 2016-05-31
limit common deprecation messages to once per package

minrk:
since `_default` and `config=True` deprecations are likely to occur numerous times in packages, limit their display to once per package by default.

Setting env TRAITLETS_ALL_DEPRECATIONS=1 will go back to showing all the deprecation warnings.

Not sure if this is the best thing to do, but it should reduce some of the annoyance while packages are in transition.

### DIFF
--- a/traitlets/config/tests/test_configurable.py
+++ b/traitlets/config/tests/test_configurable.py
@@ -19,6 +19,7 @@ from traitlets.config.configurable import (
 
 from traitlets.traitlets import (
     Integer, Float, Unicode, List, Dict, Set,
+    _deprecations_shown,
 )
 
 from traitlets.config.loader import Config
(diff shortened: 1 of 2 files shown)


Carreau:
Can't we just add a warningsfilter `once` ?

minrk:
We could do that, it would certainly be simpler. The behavior wouldn't be quite the same. This PR limits it to once per package, while a `once` filter limits it to once per module, so it can still be a lot of repetition for a package like IPython.

What do you think? More warnings, but less code of our own, or more code of our own and fewer warnings? Of course, these are all deprecations that are not shown to anyone by default, so it's probably not worth spending a huge amount of time on.

Carreau:
Up to you, you already did the work here and I have no objection to merge that in. 

I'm not convinced having too many deprecation warning during tests is that harmful.

Carreau:
Ok let's stop bikeshedding, 

Thanks a lot for the work you've put in that. That's a good improvement !

PR #238, opened 2016-06-02
add keys to bunch.__dir__

minrk:
so they are exposed to attribute completions

### DIFF
--- a/traitlets/utils/bunch.py
+++ b/traitlets/utils/bunch.py
@@ -18,7 +18,8 @@ class Bunch(dict):
         self.__setitem__(key, value)
     
     def __dir__(self):
-        names = super(Bunch, self).__dir__()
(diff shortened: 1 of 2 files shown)


traitlets/utils/bunch.py
```diff
@@ -16,3 +16,9 @@ def __getattr__(self, key):
     
     def __setattr__(self, key, value):
         self.__setitem__(key, value)
+    
+    def __dir__(self):
+        names = super(Bunch, self).__dir__()
+        names.extend(self.keys())
+        return names
```
SylvainCorlay: Cool I did not know one could do that.

minrk: Turns out, you can't on Python 2. I did something a bit worse so that it works on both reasonable Python and Python 2.

Carreau:
Thanks !

PR #239, opened 2016-06-02
run tests with py.test

minrk:
instead of nose2.

And use codecov instead of coveralls while we're at it, since that's what we do now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ python:
     - 3.3
 sudo: false
 install:
-    - pip install . nose2
-    - pip install . cov-core
+    - pip install .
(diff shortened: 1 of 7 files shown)


minrk:
nose2 was recently introduced as a solution to nose being mostly abandoned, but pytest seems a bit more common, and what we use on most of our projects that have moved away from nose. Not a lot of changes to make.

Carreau:
+1

rmorshea:
Should change the docs for running the tests.

minrk:
@rmorshea good catch, done.

Carreau:
Thanks @minrk  !

PR #242, opened 2016-06-10
[WIP] Add conf.d config-loading

### LINKED ISSUE #192: conf.d-style loading of a bag of files
One of the things discussed with the [downstream](https://github.com/jupyter/notebook/pull/879) is a [conf.d](https://lists.debian.org/debian-devel/2010/04/msg00352.html)-style loading of a directory of files, probably put there by a package manager (at least, not put there by an application itself).

It could live here. This would have the benefit of making all ConfigManager downstream implementations able to use this feature. There would also be the question of whether one could silently add this.

If it lived here, it looks like it would have to be reflected in:
- a FileConfigLoader subclass
- a BaseJSONConfigManager

> **Question** Why do both of these even exist?

For example, in the notebook, this would entail loading and merging the following files:
- `[sys.prefix | /usr/local | ~)/jupyter/
  - `jupyter_notebook_config.json.d/`
    0. some_thing.json
    1. some_other_thing.json
  - `jupyter_notebook_config.json`

When writing in both those files, when a config is _written_, it can only write to the "non-.d" location. Further, it shouldn't write out the whole shooting match (of all the `.d/*` files), but only the novel, unique data introduced by the user. so:

```
conda install nbpresent
jupyter nbextension disble --py some_other_extension
```

shouldn't store the nbpresent data in the config. This suggests we may need some kind of "taint checking" to know whether a value should be persisted or not.

minrk:
Wherever a `config_file.{py|json}` would be loaded, also load any/all config in `config_file.d/*.{py|json}`.

This is a prototype for discussion in https://github.com/jupyter/notebook/issues/1508

closes #192

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -581,18 +581,31 @@ class Application(SingletonConfigurable):
         yield each config object in turn.
         """
 
+        def new_loader(name, path):
+            if name.endswith('.py'):
+                return cls.python_config_loader_class(name, path=path, log=log)
+            elif name.endswith('.json'):
+                return cls.json_config_loader_class(name, path=path, log=log)
+
         if not isinstance(path, list):
             path = [path]
         for path in path[::-1]:
             # path list is in descending priority order, so load files backwards:
-            pyloader = cls.python_config_loader_class(basefilename+'.py', path=path, log=log)
             if log:
                 log.debug("Looking for %s in %s", basefilename, path or os.getcwd())
-            jsonloader = cls.json_config_loader_class(basefilename+'.json', path=path, log=log)
+            pyloader = new_loader(basefilename + '.py', path=path)
+            jsonloader = new_loader(basefilename + '.json', path=path)
+            loaders = [pyloader, jsonloader]
+            # load conf.d/config files in lorder
+            conf_d = os.path.join(path, basefilename + '.d')
+            if os.path.isdir(conf_d):
+                for filename in sorted(os.listdir(conf_d)):
+                    if filename.endswith(('.py', '.json')):
+                        loaders.append(new_loader(filename, path=conf_d))
             config = None
             loaded = []
             filenames = []
-            for loader in [pyloader, jsonloader]:
+            for loader in loaders:
                 try:
                     config = loader.load_config()
                 except ConfigFileNotFound:


traitlets/config/application.py
```diff
@@ -535,13 +535,25 @@ def _load_config_files(cls, basefilename, path=None, log=None, raise_config_file
         if not isinstance(path, list):
             path = [path]
         for path in path[::-1]:
-            # path list is in descending priority order, so load files backwards:
-            pyloader = cls.python_config_loader_class(basefilename+'.py', path=path, log=log)
             if log:
                 log.debug("Looking for %s in %s", basefilename, path)
-            jsonloader = cls.json_config_loader_class(basefilename+'.json', path=path, log=log)
+            # path list is in descending priority order, so load files backwards:
+            def new_loader(name, path):
```
rmorshea: No reason to define `new_loader` inside the for-loop over `path[::-1]` on line 537

Madhu94:
@minrk  This change hasn't been merged yet ?

minrk:
I think we still do want this in general, but it is no longer an urgent need thanks to that PR.

bollwyvl:
Raised some questions about [py/json vs env/home precendence](https://github.com/ipython/traitlets/pull/452#issuecomment-350051700). Would love to see this finally put to rest!

traitlets/config/application.py
```diff
@@ -581,18 +581,31 @@ def _load_config_files(cls, basefilename, path=None, log=None, raise_config_file
...
             path = [path]
         for path in path[::-1]:
             # path list is in descending priority order, so load files backwards:
-            pyloader = cls.python_config_loader_class(basefilename+'.py', path=path, log=log)
             if log:
                 log.debug("Looking for %s in %s", basefilename, path or os.getcwd())
-            jsonloader = cls.json_config_loader_class(basefilename+'.json', path=path, log=log)
+            pyloader = new_loader(basefilename + '.py', path=path)
+            jsonloader = new_loader(basefilename + '.json', path=path)
+            loaders = [pyloader, jsonloader]
```
ankostis: I have a question not strictly related to the core functionality of this PR
Why did you changed the separate `cls.XXX_config_loader_class()` calls with `new_loader()`?

I also need to change the abov 3 lines,
but for adding a new loader (i.e. YAML) without overriding and reimplementing `_load_config_files()`,
like that:
```diff
-            pyloader = cls.python_config_loader_class(basefilename+'.py', path=path, log=log)
             if log:
                 log.debug("Looking for %s in %s", basefilename, path or os.getcwd())
-            jsonloader = cls.json_config_loader_class(basefilename+'.json', path=path, log=log)
+           loaders = cls._make_loaders(basefilename, path=path, log=log)
+           log.debug("Looking for %s in %s", basefilename, path or os.getcwd())
```

You may implement `_make_loaders()` either with `new_loader()` or `cls.XXX_config_loader_class()` calls.
 Then adding a new loader requires only to overrd this trivial method.

Carreau:

PR #245, opened 2016-06-17
[WIP] allow casting integer values to integer traits

### LINKED ISSUE #243: Add CInteger trait type
I think there is a need for a CInteger trait: A trait that converts anything you throw at it into a python integer if possible, without enforcing either `int` or `long` on python 2. In essence, it should simply call `int(value)` for coercion, which under python 2 only produces a `long` where needed. A typical situation is when numpy array scalars are assigned to a trait, which will be rejected by the non-coercing traits.

minrk:
Rational, Decimal, Complex, etc. values that are real-integer-valued can be used as integers without using casting trait variants.

In general, the goal should be that anything where `int(x) == x` should be accepted for integer traits, etc. This should be the difference between Integer and CInt, because CInt allows anything that can be cast to an int, regardless of mutation, whereas Integer should be able to accept anything integer-valued.

closes #243

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -7,6 +7,8 @@
 # Adapted from enthought.traits, Copyright (c) Enthought, Inc.,
 # also under the terms of the Modified BSD License.
 
+import decimal
+import fractions
 import pickle
 import re
 import sys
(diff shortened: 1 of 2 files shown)


rmorshea:
if/when https://github.com/ipython/traitlets/pull/403 is merged, this should be reworked using the `TypeCast` api for casting values.

Carreau:

Issue #210, opened 2016-04-25
autodoc plugin for sphinx


minrk:
Perhaps related to #131, we should consider an extension to sphinx autodoc that would let you include trait info, such as types and help strings, in autodoc output.

This doesn't have to start in this repo, it can be a standalone extension, but I think it would be useful for things like generating good docs on what's configurable in a project.

rmorshea:
So far #131 pulls help text from `TraitType.info()`, lists the initial metadata, and shows a trait's associated event handlers. Is there more information that would be good to automatically generate?

minrk:
I'm mainly thinking of generating a nice rst summary with type info and help strings. The mechanism will probably be similar, but the output will probably need to be a bit different, since it should do things like `.. autoattribute::`, etc., and indicate whether they are configurable, etc.

PR #246, opened 2016-06-22
add envvar default handler

rmorshea:
substitute for #158 
closes #157 

see [test](https://github.com/rmorshea/traitlets/blob/dcd9e1db4717e1fac978d3fba999b174afd3fb69/traitlets/tests/test_traitlets.py#L179) / [doc](https://github.com/rmorshea/traitlets/blob/dcd9e1db4717e1fac978d3fba999b174afd3fb69/traitlets/traitlets.py#L868) for usage

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -22,7 +22,7 @@ from traitlets import (
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
     ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default,
-    observe_compat, BaseDescriptor, HasDescriptors,
+    observe_compat, BaseDescriptor, HasDescriptors, envvar_default
 )
 
 import six
@@ -175,6 +175,28 @@ class TestTraitType(TestCase):
         self.assertEqual(a._trait_values, {})
         self.assertEqual(a.x, 11)
         self.assertEqual(a._trait_values, {'x': 11})
+
+    def test_envvar_dynamic_initializer(self):
+        import os
+
+        os.environ['MY_ENVVAR'] = 'true'
+
+        class A(HasTraits):
+            b = Bool()
+
+            @envvar_default('b', 'MY_ENVVAR')
+            def _b_from_envvar(self, init):
+                if init.value is None:
+                    m = "The trait '%s' expected os.environ[%r] to be set"
+                    raise TraitError(m % (init.trait.name, init.envvar))
+                else:
+                    return bool(init.value)
+
+        a = A()
+        self.assertTrue(a.b)
+        del os.environ['MY_ENVVAR']
+        del a._trait_values['b']
+        self.assertRaises(TraitError, getattr, a, 'b')
 
     def test_tag_metadata(self):
         class MyIntTT(TraitType):
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -43,6 +43,7 @@ Inheritance diagram:
 import contextlib
 import inspect
 import re
+import os
 import sys
 import types
 import enum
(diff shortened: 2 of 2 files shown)


traitlets/tests/test_traitlets.py
```diff
@@ -176,6 +176,28 @@ def _default_x(self):
...
+    def test_envvar_dynamic_initializer(self):
+        import os
+
+        os.environ['MY_ENVVAR'] = 'true'
+
+        class A(HasTraits):
+            b = Bool()
+
+            @envvar_default('b', 'MY_ENVVAR')
+            def _b_from_envvar(self, init):
```
minrk: What is `init` here? Shouldn't it be the value of the environment variable?

rmorshea: I was working from the idea that the handlers should generally accept a 'bunch' object, or nothing, so I chose to include the trait in the 'init' bunch. I couldn't really think of a better name than 'init' so if there's something better I'm open to it.

minrk:
While the decorator gives you the most control, needing to define a method means this doesn't really improve the situation over the status quo, where today you might have:

```python
class C(Configurable):
    a = Unicode().tag(config=True)
    @default('a')
    def _a_default(self):
        return os.environ.get('A_ENV', '')
```

whereas with this, it becomes:

```python
class C(Configurable):
    a = Unicode().tag(config=True)
    @envvar_default('a', 'A_ENV')
    def _a_default(self, init):
        return init.value or ''
```

which requires a new import, understanding a new construct, etc.

I'm not sure this is worth adding, unless it can be done via *just* a tag, at least sometimes, e.g.:

```python
class C(Configurable):
    a = Unicode().tag(config=True, envvar='A_ENV')
```

Then, *optionally*, a validate method could be used to handle any validation/casting/etc.

Carreau:
Thanks @rmorshea !

I think one of the biggest point that could help get that in would be help autogeneration. 
It's relatively hard to keep docs in Sync, or even just write the docs when you use envvars.
Here I guess we could pull that out automatically !

I'm also wondering if it **has to** be on core. It looks here like we could make it a separate package, which would allow to : 
  - iterate on it
  - Bring functionality older traitlets versions. 

Though before diving into that, and want to make sure we want to move forward with this;
I tend to agree with @minrk, and the usecase seem to be really narrow, and I'm not convinced this will be that much used. Though I might not be the best to judge. 

I think we should need to be extra careful for the precedence rule of ENV var vs config, vs command line argument as well, and make sure we agree with the order, as well as the signification of having an env variable unset vs having an empty env variable. I can see also issue with casting as most env var are strings. 

I think @fperez  is often of help in these case. Can you look at this as well as related issue ? We tend to come back and forth on that and pull in many direction.

rmorshea:
@minrk I think the fact that envvar defaults are only applicable to traits of `Configurables` makes `trait.tag(envvar='x')` very awkward to implement as a part of `TraitType` since it's available to all `HasTraits` subclasses. I'd argue the feature isn't not worth implementing if we need to do it with `tag`.

rmorshea:
@Carreau I'll have to familiarize myself more with `sphinx` before I can cover something like #210, but I can make that my next project with `traitlets` since I've already put work into autodocs with #131.

As for where the pr belongs in traitlets? I'm not really too sure. Perhaps @fperez can help with that, or the original group involved with #157.

Carreau:
> @Carreau I'll have to familiarize myself more with sphinx

I was more thinking of `--help-all` which woudl likely require to only patch a [single method in `configurable.py`](https://github.com/ipython/traitlets/blob/9988734c6e65f73e87aeabdcd689711c2aead188/traitlets/config/configurable.py#L205-L218)  to be aware of the env varaible, and basically just happend a templated paragraph. Sphinx is quite a complex beast.

rmorshea:
@Carreau, given that. I think #131, makes [strong progress](https://github.com/rmorshea/traitlets/blob/doc_gen/traitlets/utils/docgen.py#L100-L124) towards that goal.

PR #249, opened 2016-06-24
copy self.config in update_config

### LINKED ISSUE #248: Command line flag do not take precedence over config.
Cf https://github.com/ipython/ipython/pull/9655#issuecomment-228182348

Marking as critical as this was release in 4.1.

minrk:
protects IPython's CLI priority, which accidentally relies upon self.config being replaced during load_config_files.

for backport to 4.2.2

closes #248

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -187,9 +187,12 @@ class Configurable(HasTraits):
 
     def update_config(self, config):
         """Update config and load the new values"""
-        # make a copy of our config prior to loading to protect IPython,
-        # which relies on self.config being immutable
-        # to preserve CLI config priority
+        # traitlets prior to 4.2 created a copy of self.config in order to trigger change events.
+        # Some projects (IPython < 5) relied upon one side effect of this,
+        # that self.config prior to update_config was not modified in-place.
+        # For backward-compatibility, we must ensure that self.config
+        # is a new object and not modified in-place,
+        # but config consumers should not rely on this behavior.
         self.config = deepcopy(self.config)
         # load config
         self._load_config(config)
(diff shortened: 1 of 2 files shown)


traitlets/config/configurable.py
```diff
@@ -187,6 +187,10 @@ def _config_changed(self, change):
 
     def update_config(self, config):
         """Update config and load the new values"""
```
fperez: Given these specific semantics (deepcopy first before update), I think it would be a good idea to extend a little bit the docstring with a note that indicates that's what happens, so it becomes clearly official that such semantics are the intended ones.

traitlets/config/configurable.py
```diff
@@ -187,6 +187,10 @@ def _config_changed(self, change):
...
+        # make a copy of our config prior to loading to protect IPython,
```
fperez: Rather than this being an IPython-specific hack, which is what the comments seems to imply, we should indicate whether we mean these semantics to be how the system is designed to work, and in that case IPython would be making a proper, supported assumption.

fperez:
Other than my small and easy to address comments, +1 for merging.

traitlets/config/configurable.py
minrk: I've updated the comment to make it clear that IPython is **not** making a proper, supported assumption. I left it out of the docstring, because I don't want to commit to the behavior. This is specifically reintroducing undesirable behavior for backward compatibility in support of IPython in particular due to an accidental assumption in IPython of unspecified, undocumented behavior.

minrk:
Comment updated, let me know if I can do anything else to make the situation clearer.

fperez:
Much better, thanks! I think that's clear enough, let's move forward.

PR #250, opened 2016-06-24
always override config files with CLI options

minrk:
This includes #249, so will need a rebase after that's merged

This is done in all IPython and Jupyter traitlets Applications, but not implemented in the base Application class until now.

- save copy of cli_config in Application.cli_config during parse_command_line
- apply cli_config to config from files prior to loading,
  so that any lower-priority changes in config files are never loaded by the application,
  rather than requiring load and re-load to preserve priority.

avoids bugs like #248 by applying CLI config priority in load_config_file instead of relying on applications to implement it.

If we want to preserve prior behavior of load_config_file clobbering CLI config, we can put this behavior beind an optional flag in load_config_file or an attribute on Application.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -6,6 +6,7 @@
 
 from __future__ import print_function
 
+from copy import deepcopy
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


fperez:
+1, logic looks good, and the small changes I suggested to #249 should propagate cleanly upon rebase.

PR #251, opened 2016-07-01
Add an EventReport class

rmorshea:
The `EventReport` class provides a base class for all `FrozenBunch` instances being passed to event callbacks. At minimum, a report has a `'type'` key/attribute indicating the type of event being reported. Outside of this, it's up to subclasses to add more information.

Benefits of an `EventReport` class:
1. Uniformity can be enforced across traitlets callbacks
2. Subclasses can set which keys/attributes they contain
3. The report instances themselves are "frozen" bunches

*note: we implement a `FrozenBunch` rather than a `Bunch` class (still inherits from `dict`)

This will break any usages of Traitlets that attempt to modify event reports. Ultimately this shouldn't ever happen, but it may be a significant enough problem that the reports will need to be mutable. Making them mutable would be unfortunate though, since an error caused by a notification that accidently modified a report would be hard to trace.

At the very least it would be beneficial to implement an `EventReport` that has a `'type'`:

``` python
class EventReport(Bunch):
    def __init__(self, type, **kwargs):
        super(Bunch, self).__init__(type=type, **kwargs)
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -86,30 +86,29 @@ NoDefaultSpecified = Undefined
 class TraitError(Exception):
     pass
 
-class EventReport(FrozenBunch):
-    """An object containing callback information for particular events"""
 
-    _event_type = None
+class EventReport(FrozenBunch):
+    """An object containing callback information for particular events
+
+    Event report classes provide a base class for all FrozenBunch
+    instances being passed to event callbacks. At minimum, a report
+    has a 'type' key/attribute indicating the type of event being
+    reported. Outside of this, it's up to subclasses to add more
+    information.
+    """
+
+    _etype = None
 
     def __init__(self, *args, **kwargs):
         super(EventReport, self).__init__(*args, **kwargs)
         if "type" not in self:
-            dict.__setitem__(self, "type", self._event_type)
-
-    def copy(self, **updates):
-        """Get a new instance with the given updates"""
-        if 'type' in updates:
-            raise KeyError("'type' is a reserved key")
-        else:
-            new = super(EventReport, self).copy()
-            dict.update(new, updates)
-            return new
+            dict.__setitem__(self, "type", self._etype)
 
 
 class DefaultEvent(EventReport):
     """An object holding information about a trait's generated default"""
 
-    _event_type = 'default'
+    _etype = 'default'
 
     def __init__(self, name=None, value=Undefined, owner=None):
         super(DefaultEvent, self).__init__(name=name, value=value, owner=owner)
@@ -118,7 +117,16 @@ class DefaultEvent(EventReport):
 class ChangeEvent(EventReport):
     """An object holding information for trait change callbacks"""
 
-    _event_type = 'change'
+    _etype = 'change'
+
+    def rollback(self):
+        if self.old is not Undefined:
+            self.owner.set_trait(self.name, self.old)
+        else:
+            del self.owner._trait_values[self.name]
+
+    def compress(self, other):
+        return self.copy(new=other.new)
 
     def __init__(self, name=None, new=Undefined, old=Undefined, owner=None):
         super(ChangeEvent, self).__init__(name=name, new=new, old=old, owner=owner)
@@ -126,7 +134,7 @@ class ChangeEvent(EventReport):
 class ValidationEvent(EventReport):
     """An object holding information for trait validation callbacks"""
 
-    _event_type = 'proposal'
+    _etype = 'validation'
 
     def __init__(self, trait=None, value=Undefined, owner=None):
         super(ValidationEvent, self).__init__(trait=trait, value=value, owner=owner)
@@ -1140,21 +1148,19 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
             cache = {}
             notify_change = self.notify_change
 
-            def compress(past_changes, change):
-                """Merges the provided change with the last if possible."""
-                if past_changes is None:
-                    return [change]
-                else:
-                    if past_changes[-1]['type'] == 'change' and change.type == 'change':
-                        past_changes[-1] = EventReport(past_changes[-1], new=change.new)
-                    else:
-                        # In case of changes other than 'change', append the notification.
-                        past_changes.append(change)
-                    return past_changes
-
             def hold(change):
-                name = change.name
-                cache[name] = compress(cache.get(name), change)
+                name = change.get("name")
+                if name in cache:
+                    past = cache[name]
+                    if "compress" in change:
+                        past[-1].type
+                        if change.type == past[-1].type:
+                            past[-1] = past[-1].compress(change)
+                    else:
+                        past.append(change)
+                else:
+                    past = [change]
+                    cache[name] = past
 
             try:
                 # Replace notify_change with `hold`, caching and compressing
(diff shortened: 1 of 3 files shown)


traitlets/traitlets.py
```diff
@@ -1129,37 +1182,38 @@ def hold(change):
...
-        name, type = change.name, change.type
+            change = EventReport(change)
 
         callables = []
-        callables.extend(self._trait_notifiers.get(name, {}).get(type, []))
-        callables.extend(self._trait_notifiers.get(name, {}).get(All, []))
-        callables.extend(self._trait_notifiers.get(All, {}).get(type, []))
-        callables.extend(self._trait_notifiers.get(All, {}).get(All, []))
+        if 'name' in change:
+            # name is not gaurenteed to be in the event report
```
minrk: Why/how is name not guaranteed? It's assumed two lines above here in master, so this method would error if name were ever left undefined.

minrk:
👍 on the idea. Since this is technically a backward-incompatible change, even though a small one, I think we should start marking traitlets master as 5.0, and skip 4.3. *Any* such change should trigger a major release, but that doesn't mean major releases should be big or break lots of stuff.

traitlets/traitlets.py
rmorshea: In the docs for 'observe' and '@observe' on master the "type" is the only guarenteed info in a "change dict", so the if-statement is to accomodate that.

minrk: Maybe add to the docs that `'name'` is required instead of handling the never-true case here, since there are currently no cases where it can be absent, and there probably shouldn't ever be any.

rmorshea: `'name'` issue resolved by https://github.com/ipython/traitlets/pull/285

rmorshea:
@minrk rebased.

rmorshea:
@SylvainCorlay add rollback and compression to `ChangeEvent`

traitlets/traitlets.py
```diff
@@ -1154,18 +1194,18 @@ def hold(change):
...
-            type='change',
         ))
 
     def notify_change(self, change):
-        if not isinstance(change, Bunch):
+        if not isinstance(change, EventReport):
             # cast to bunch if given a dict
-            change = Bunch(change)
+            change = EventReport(change)
+            raise
```
rmorshea: Need to remove this debug statement...

traitlets/utils/bunch.py
```diff
@@ -6,8 +6,30 @@
...
+            m = "type object '%s' " % cls.__name__
+        raise self.etype(m + self.msg)
+
+
 class Bunch(dict):
-    """A dict with attribute-access"""
+    """A dict with attribute access"""
+
+    def __contains__(self, key):
+        return hasattr(self, key)
```
minrk: This shouldn't be added. Methods shouldn't report as being `in` the Bunch.

traitlets/traitlets.py
```diff
@@ -88,6 +87,59 @@
...
+
+    _etype = 'change'
+
+    def rollback(self):
+        if self.old is not Undefined:
+            self.owner.set_trait(self.name, self.old)
+        else:
+            del self.owner._trait_values[self.name]
+
+    def compress(self, other):
```
minrk: Adding these methods on event reports is exposing them to observer/validate handlers. Is that what we want? It doesn't seem so. Or are we adding a `change.rollback()` public API to be called in any observer?

traitlets/traitlets.py
```diff
@@ -783,13 +830,12 @@ def setup_class(cls, classdict):
...
     """A decorator which can be used to observe Traits on a class.
 
-    The handler passed to the decorator will be called with one ``change``
-    dict argument. The change dictionary at least holds a 'type' key and a
-    'name' key, corresponding respectively to the type of notification and the
-    name of the attribute that triggered the notification.
+    The handler for the decorator must accept one ``change`` argument, into which is
+    passed a :class:`EventReport` instance that holds (at minumum) a 'type' key/attribute
+    corresponding to the type of event being reported. Other keys/attributes may be
+    included depending on the value of 'type'.
```
minrk: This has removed the 'name' field, which is still guaranteed. It should also be clear that the object is still a considered a dictionary with attribute-access, rather than making any reference to EventReport, which is not a user-facing API.

traitlets/traitlets.py
```diff
@@ -1113,21 +1158,19 @@ def hold_trait_notifications(self):
...
-                        past_changes.append(change)
-                    return past_changes
-
             def hold(change):
-                name = change.name
-                cache[name] = compress(cache.get(name), change)
+                name = change.get("name")
+                if name in cache:
+                    past = cache[name]
+                    if "compress" in change:
```
minrk: Don't use dict-access to check for methods. This should always return False. If we're checking types, use `isinstance`.

traitlets/traitlets.py
```diff
@@ -1146,11 +1189,8 @@ def hold(change):
                 for name, changes in cache.items():
                     for change in changes[::-1]:
                         # TODO: Separate in a rollback function per notification type.
-                        if change.type == 'change':
-                            if change.old is not Undefined:
-                                self.set_trait(name, change.old)
-                            else:
-                                self._trait_values.pop(name)
+                        if "rollback" in change:
```
minrk: Same here - methods shouldn't be present in the change dict.

traitlets/traitlets.py
```diff
@@ -1275,12 +1315,11 @@ def observe(self, handler, names=All, type='change'):
         Parameters
         ----------
         handler : callable
-            A callable that is called when a trait changes. Its
-            signature should be ``handler(change)``, where ``change`` is a
-            dictionary. The change dictionary at least holds a 'type' key.
-            * ``type``: the type of notification.
-            Other keys may be passed depending on the value of 'type'. In the
-            case where type is 'change', we also have the following keys:
+            A callable that accepts one ``change`` argument, into which is
```
minrk: > where `change` is a dictionary with attribute-access (Bunch)

traitlets/traitlets.py
rmorshea: @minrk, these methods should definitely be private.

traitlets/utils/bunch.py
rmorshea: Yup!

traitlets/traitlets.py
rmorshea: Yup!

traitlets/utils/bunch.py
```diff
@@ -19,7 +41,42 @@ def __setattr__(self, key, value):
...
+        classdict['__delattr__'] = EmptyMethodDescriptor( '__delattr__',
+                        TypeError, 'does not support attribute deletion')
+        classdict['__setitem__'] = EmptyMethodDescriptor( '__setitem__',
+                        TypeError, 'does not support item assignment')
+        classdict['__delitem__'] = EmptyMethodDescriptor('__delitem__',
+                        TypeError, 'does not support item deletion')
+        return super(MetaFrozenBunch, mcls).__new__(mcls, name, bases, classdict)
+
+
+class FrozenBunch(with_metaclass(MetaFrozenBunch, Bunch)):
```
rmorshea: @minrk, are you saying that we don't need to freeze the bunch, or that there must be a better way to freeze it.

minrk: Either way. I don't think we need to freeze it, but if you really want to, I think metaclasses isn't the right way to do it.

traitlets/traitlets.py
```diff
@@ -1146,11 +1189,8 @@ def hold(change):
                 for name, changes in cache.items():
                     for change in changes[::-1]:
                         # TODO: Separate in a rollback function per notification type.
-                        if change.type == 'change':
-                            if change.old is not Undefined:
-                                self.set_trait(name, change.old)
-                            else:
-                                self._trait_values.pop(name)
+                        if "rollback" in change:
+                            change.rollback()
```
rmorshea: Need to make this logic backward compatible.

Carreau:

PR #254, opened 2016-07-02
enable non-decorator event handler setup

rmorshea:
## Summary

Enable non-decorate setup for event handlers:

+ `EventHandler._init_call(func)` goes to `EventHandler.setup_callback(func, force=False)`

This change makes it possible to register a callback without using `__call__`.

Currently, if an attempt to re-instantiate a callback with `__call__` is made, no useful error gets raise because the new callback will simply be passed to the old one as an argument.

With this change, non-decorator `EventHandler` setup should now use `setup_callback`. By default a `ValueError` is raised if there is an attempt to replace a callback. However this can be bypassed with `force=True`.

## Problem Case

```python
def callback_0(change):
	print('cb0', change)

def callback_1(change):
	print('cb1', change)

# the only public way to register
# a callback is with `__call__`
e = ObserveHandler('a', 'change')(callback_0)

# if we try to register a callback again
# on accident we get unexpected
# behavior, no errors get raised, and
# the new callback is never registered
e(callback_1)
```
```
('cb0', <function callback_1 at 0x10ad147d0>)
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -893,16 +893,28 @@ def default(name):
 
 class EventHandler(BaseDescriptor):
 
-    def _init_call(self, func):
+    func = None
+
+    def setup_callback(self, func, force=False):
+        """Setup a new callback for the event
+
+        Use ``force=True`` to setup the callback even if one already exists
+        """
+        if self.func is not None and not force:
+            raise ValueError("a callback already exist")
         self.func = func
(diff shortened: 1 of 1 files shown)


minrk:
Can you add an example and test illustrating what problem this solves?

rmorshea:
Just renaming (I misspelled the branch name)

PR #255, opened 2016-07-07
allow non-decorator event handler setup

rmorshea:
## Summary

Enable non-decorate setup for event handlers:

+ `EventHandler._init_call(func)` goes to `EventHandler.setup_callback(func, force=False)`

This change makes it possible to register a callback without using `__call__`.

Currently, if an attempt to re-instantiate a callback with `__call__` is made, no useful error gets raise because the new callback will simply be passed to the old one as an argument.

With this change, non-decorator `EventHandler` setup should now use `setup_callback`. By default a `ValueError` is raised if there is an attempt to replace a callback. However this can be bypassed with `force=True`.

## Problem Case

```python
def callback_0(change):
	print('cb0', change)

def callback_1(change):
	print('cb1', change)

# the only public way to register
# a callback is with `__call__`
e = ObserveHandler('a', 'change')(callback_0)

# if we try to register a callback again
# on accident we get unexpected
# behavior, no errors get raised, and
# the new callback is never registered
e(callback_1)
```
```
('cb0', <function callback_1 at 0x10ad147d0>)
```

## Solution

```python
def callback_0(change):
	print('cb0', change)

def callback_1(change):
	print('cb1', change)

# a public callback setup method is available
e = ObserveHandler('a', 'change').setup_callback(callback_0)

# try to register a callback again
# on accident will now raise *if
# `register_callback` is used*
.setup_callback(callback_1)
```

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -22,7 +22,7 @@ from traitlets import (
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
     ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default,
-    observe_compat, BaseDescriptor, HasDescriptors,
+    observe_compat, BaseDescriptor, HasDescriptors, EventHandler
 )
 
 import six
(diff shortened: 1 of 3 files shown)


minrk:
Under what circumstances would ObserveHandler or EventHandler ever be called outside traitlets.py itself?

rmorshea:
@minrk I've been playing with the idea of an API in which you have a decorator that accepts [`TraitEventHandler`](https://github.com/rmorshea/traitlets/blob/b4c913192842f5e6610f8177e88041edd1a93fd6/traitlets/traitlets.py#L914-L948) instances in order to install itself as a sort or generic callback. For example:

```python

class A(HasTraits):

	i = Int()
	j = Int()
	
	@ping(default('i'), observe('j'))
	def do_something(self):
		# do something generic
```

Where the above code would trigger methods decorated with `ping` when the given events occur.

minrk:
Am I right that this would be the first introduction of instantiating EventHandler as a public API for users to instantiate? If so, that's actually something I would prefer to avoid.

rmorshea:
@minrk, yes, that's correct.

PR #257, opened 2016-07-19
[WIP] make This inherit from Instance and add ThisType

rmorshea:
With the new metaclass setup phase `class_init` brought by #114, we can now make `This` properly inherit from `Instance` by setting its `klass` before an instance has a chance to request a default value.

This means that normal dynamic default value heuristics from `Instance`, are now possible in `This`.

This pr also adds a `ThisType` trait which, presumably, was never implemented in the past because it would have been too contrived. However if it's unnecessary, I can revert the commit that added it.

**breakage** : this breaks uses of `This` where the keywords `'args'` or `'kwargs'` were used as metadata

**to do**: add tests for new dynamic default value features

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -22,7 +22,7 @@ from traitlets import (
     Union, All, Undefined, Type, This, Instance, TCPAddress, List, Tuple,
     ObjectName, DottedObjectName, CRegExp, link, directional_link,
     ForwardDeclaredType, ForwardDeclaredInstance, validate, observe, default,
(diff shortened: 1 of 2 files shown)


minrk:
> breakage : this breaks uses of This where the keywords 'args' or 'kwargs' were used as metadata

is there a reason for the breakage that's not fixable?

> ThisType

What are some use cases of ThisType? I can't think of any cases where I've had an attribute on a class that is a class itself, restricted to a subclass of the class itself.

rmorshea:
@minrk, closed in favor of #317

PR #261, opened 2016-08-01
[WIP] Automatic Documentation

### LINKED ISSUE #130: Trait introspection

rmorshea:
Closes #130
Alternatives: #267 #131 

Adds a robust set of tools for rendering the content of an instance or class into a readable document.

This is accomplished with `DocLog`, `Content`, and `Template` objects. While each `DocLog` contains one template, it will hold many content attributes that fill the undefined fields of the template. The following example shows how these objects work together to document the name of a class or instance:

``` python
class GenericDocLog(DocLog):
    template = Template("{name}")
    # pulls the `__name__` attribute
    name = Content("__name__")

# by default input values are converted to classes
assert GenericDocLog(1).document() == "int"
assert GenericDocLog(int).document() == "int"
```

A series of `DocLog` types following this general format allow for detailed `HasTraits` documentation:

``` python
class A(HasTraits):
    i = Int(allow_none=True, default_value=1)
    j = Any(help='some help text for j').tag( data={'x':1}, info="foo")

    def __init__(self, *args, **kwargs):
        """The constructor for the class

        More details about the constructor"""
        pass

    @default('j')
    def h(self): pass

    @validate('j')
    def g(self, commit):
        return commit['value']

    @observe('i', 'j')
    def f(self, change):
        pass

print(help(A.__init__))
```

```
The constructor for the class

More details about the constructor

Traits of ``A`` Instances
-------------------------
i : an int
    :help: None
    :range: [-inf, inf]
    :read only: False
    :allow none: True

    **trait metadata:**
        None

    **event handlers:**
        * :meth:`A.f` : observes changes

j : any value
    :help: some help text for j
    :read only: False
    :allow none: False

    **trait metadata:**
        * info: 'foo'
        * data: {'x': 1}

    **event handlers:**
        * :meth:`A.h` : default value generator
        * :meth:`A.g` : validates values
        * :meth:`A.f` : observes changes
```
## To Do:
- finalize core features
- write tests for them
- more gracefully handle failures
- thoroughly document and comment

### DIFF
--- a/traitlets/doclogs.py
+++ b/traitlets/doclogs.py
@@ -1,0 +1,325 @@
+import six
+
+from .traitlets import *
+from utils.docgen import (DocLog, DocLogMap,
+	Content, content, Template, ThisTemplate)
+
+
+if six.PY3:
+    def copy_func(f, name=None):
+        return types.FunctionType(f.__code__, f.__globals__,
+        	name or f.__name__, f.__defaults__, f.__closure__)
+else:
+    def copy_func(f, name=None):
+        return types.FunctionType(f.func_code, f.func_globals,
+        	name or f.func_name, f.func_defaults, f.func_closure)
+
+def write_docs_to_class(cls, docs):
+    """Write traitlet documentation for this class to __init__"""
+    im_func = copy_func(cls.__init__.im_func)
+    im_func.__doc__ = trim_docstring(im_func.__doc__) + "\n\n" + docs
+    cls.__init__ = types.MethodType(im_func, None, cls)
+
+
+def trim_docstring(docstring):
+	if not docstring:
+		return ''
+	# Convert tabs to spaces (following the normal Python rules)
+	# and split into a list of lines:
+	lines = docstring.expandtabs().splitlines()
+	# Determine minimum indentation (first line doesn't count):
+	indent = sys.maxint
+	for line in lines[1:]:
+		stripped = line.lstrip()
+		if stripped:
+			indent = min(indent, len(line) - len(stripped))
+	# Remove indentation (first line is special):
+	trimmed = [lines[0].strip()]
+	if indent < sys.maxint:
+		for line in lines[1:]:
+			trimmed.append(line[indent:].rstrip())
+	# Strip off trailing and leading blank lines:
+	while trimmed and not trimmed[-1]:
+		trimmed.pop()
+	while trimmed and not trimmed[0]:
+		trimmed.pop(0)
+	# Return a single string:
+	return '\n'.join(trimmed)
+
+def full_type_name(obj):
+	if not inspect.isclass(obj):
+		klass = obj.__class__
+	else:
+		klass = obj
+	m_n = klass.__module__
+	if m_n != "__main__":
+		return m_n + '.' + klass.__name__
+	else:
+		return klass.__name__
+
+
+class BaseDescriptorDocLog(DocLog):
+
+	template = Template()
+
+	logs_type = BaseDescriptor
+
+	_from_instances = True
+	name = Content('name')
+
+class TraitDocLog(BaseDescriptorDocLog):
+
+	logs_type = TraitType
+
+	read_only = Content('read_only')
+	allow_none = Content('allow_none')
+
+	template = ThisTemplate().extend([
+		"{name} : {info}",
+		"    :help: {help}",
+		"    :read only: {read_only}",
+		"    :allow none: {allow_none}",
+		"",
+		"    **trait metadata:**",
+		"        {metadata}",
+		"",
+		"    **event handlers:**",
+		"        {events}"
+		])
+
+	@content
+	def info(self, source):
+		return source.info()
+
+	@content('help')
+	def help(self, trait):
+		return self.metadata.pop('help', None)
+
+	@content('default_value')
+	def default_value(self, value):
+		if 'default_value' in self.metadata:
+			# pop from given metadata if present
+			return self.metadata.pop('default_value')
+		else:
+			return value
+
+	@content("metadata")
+	def metadata(self, metadata):
+		return metadata.copy()
+
+	@metadata.format
+	def metadata(self, value):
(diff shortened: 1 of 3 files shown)


minrk:
Nice!

rmorshea:
@akhmerov I've been working at this for a while on my own, but I'd like to get outside my own head for a bit - what features do you see as being most useful to customizing descriptor documentation?

For instance, to make it easier to extend the `Template` of a parent `DocLog`, I thought defining individually modifiable template sections could be useful. In other words `template.section.append("text")` instead of `template.insert(index, "text")`. While I think this is an interesting feature, I want to know if there's anything else that my time would be better spent on.

cc: @minrk, @SylvainCorlay, @ellisonbg

akhmerov:
Great question. Previously I was thinking about this exclusively from a user perspective, trying to use various parts of jupyter machinery and encountering minimal docstrings/call signatures. From this POV it is already infinitely better, regardless of formatting. Not sure what the devs think, but I would consider it super valuable in current form already.

From a dev point of view, different projects maintain different documentation styles, so more customization would be swell indeed. However perhaps control over content choice would be more important, especially if one wants to distinguish between own traits and inherited traits.  Or maybe I want to turn off trait autodoc for a class. Not sure if it's a cannon against sparrows but I could even imagine jinja2 in this context.

minrk:
@rmorshea one thing that I think would be super useful is working nicely with sphinx autodoc. That would suggest ensuring proper sphinx-napoleon rst docstring formatting. @SylvainCorlay's work with docrepr should also benefit from this in a transparent way.

So for me, the two things that are most useful to look at for measures of success are:

1. `TraitedClass?` in IPython
2. `autodoc` output in sphinx

In both cases, I think it's fine to add code to IPython to improve `?` for trait classes, and for really good sphinx autodoc to require a plugin (that we should also provide in traitlets) if the docstrings don't work seamlessly with autodoc, but I would at least investigate the feasibility of doing a good enough job with docstrings alone.

I don't know how customizable we need the output to be. I would certainly allow classes with user-defined docstrings to take precedence over traitlets (i.e. not unconditionally add info if docstrings are defined). Expressing a few simple options at the class level makes sense, but I would start very small with what's customizable, rather than trying to cover all cases from the start.

fwkoch:
I've just started integrating `traitlets` into a project I am working on and sphinx auto-documentation is a feature I'd love to see fleshed out. Before adopting `traitlets` we used a different properties library that dynamically generated sphinx-style docstrings in the metaclass. An example of how the output looks when rendered in sphinx is [here](https://python.steno3d.com/en/latest/content/api/resources/volume.html).

I took a stab at integrating this sort of documentation into traitlets (on this pull request: https://github.com/ipython/traitlets/pull/267).
All that is included is a small addition to `MetaHasTraits`:
```python
...
class MetaHasTraits(MetaHasDescriptors):
    """A metaclass for HasTraits."""

    def __new__(mcs, name, bases, classdict):
        def sphinx(trait_name, trait):
            return (
                ':param {name}: {doc}\n:type {name}: :class:`{cls}`'.format(
                    name=trait_name,
                    doc=trait.help,
                    cls=trait.__class__.__name__
                )
            )

        doc_str = classdict.get('__doc__', '')
        traits = {key: value for key, value in classdict.items()
                  if isinstance(value, TraitType)}
        if traits:
            doc_str += '\n\nTraits:\n\n' + '\n'.join(
                (sphinx(key, value) for key, value in traits.items())
            )
        classdict['__doc__'] = doc_str.strip()

        return super(MetaHasTraits, mcs).__new__(mcs, name, bases, classdict)

    def setup_class(cls, classdict):
        cls._trait_default_generators = {}
        super(MetaHasTraits, cls).setup_class(classdict)
...
```
The result looks like this:
screen shot 2016-08-05 at 12 22 15 pm

Right now the functionality is quite basic, but it could be expanded pretty easily:
- Rather than getting defined in the metaclass, `sphinx` could be a function of `TraitTypes`. That would allow custom documentation for different subclasses (e.g. `Instance` traits could add their `klass` to the docstring)
- Documentation isn't restricted to TraitTypes. We can introspect anything else in the `HasTraits` subclass (e.g. event handlers, or documentation-specific variables - something like `if classdict['_autodoc'] == False:`, then don't generate any docs...).
- You could also interrogate the environment to dynamically decide what format of docstrings to generate (for example, if you are in the jupyter notebook generate napoleon docs rather than sphinx docs)

Anyway, this is just my alternative solution to the sphinx docs problem. I think the end result is somewhat similar (or at least has the capacity to be similar) to the `DocLog`, but maybe has a few minor advantages. It's at least a different perspective.

rmorshea:
I shifted away from a "functional" documentation system (e.g. #131 or #267) for two main reasons:

1. I suspect that information complexity in various traits and callbacks will compound to a point where more powerful tools to enforce document styles, and inherited information will ultimately be a requirement.
2. Dispersing the parsing of a document parsing across many classes not dedicated to that purpose, might complicate debugging the auto docs, and could clutter an application's source code with unrelated material.

My hope with `DocLogs` is that they will mirror source code inheritance structures, allow `Templates` to be intuitively manipulated in subclasses, and make `Content` easy to style within isolated methods.

rmorshea:
@minrk, there's too much that needs to be done for this to make it into a 5.0 release. I might be able to get it done by 5.1 though.

minrk:
@rmorshea good call.

rmorshea:
I suspect too much has changed for this to be reasonably rebased. Additionally there are probably better and simple ways to do this than what I proposed here.

Issue #259, opened 2016-07-30
Bounded Integer


pbugnion:
The `Int` traitlet supports bounds, but not the `Integer` traitlet:

```python
import traitlets

class A(traitlets.HasTraits):
    bounded_integer = traitlets.Integer(default_value=traitlets.Undefined, min=5, max=100)

class B(traitlets.HasTraits):
    bounded_int = traitlets.Int(default_value=traitlets.Undefined, min=5, max=100)

a = A(bounded_integer=200) # raises a TraitError, as expected
a = B(bounded_int=200) # does not raise
```

This behaviour is a little confusing, since the docs recommend using the Integer traitlet rather than Int.

Is there any reason this was implemented this way (see PR #33)?

I'm happy to put in a PR for a bounded Integer if this would be useful.

minrk:
Yes, bounds should be checked. This is an oversight on Python 2, because `Integer is Int` in Python 3. If you want to add that PR, that would be fine.

pbugnion:
Thanks! I'll put in a PR.

minrk:
Great, thanks!

PR #262, opened 2016-08-02
Bounded numeric traits

pbugnion:
Prior to this PR, the `Integer` and `Long` traitlets with bounds did not actually validate that a value fell within the bounds in Python 2 (issue #259). This does not apply to Python 3, where Long and Integer are just aliases of `Int`.

In its current state, this PR adds bounds checking for `Integer` and `Long`.

*Feedback request*: To avoid repeating code, I have factored out a private top level function called `_validate_bounds`. This would make more sense as a method in a `Number` superclass that `Integer`, `Long` etc. inherit from. Since this is a major-ish change, it would be great to have feedback.

Remaining work:
- [x]  Bounds checking for casting traits, e.g. CLong
- [x]  Use `_validate_bounds` method in existing traitlets, e.g. `Float`.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1174,6 +1174,28 @@ class TestLong(TraitTestBase):
         self.assertEqual(type(self.obj.value), long)
 
 
+class MinBoundLongTrait(HasTraits):
+    value = Long(99 if six.PY3 else long(99), min=5)
+
+class TestMinBoundLong(TraitTestBase):
+    obj = MinBoundLongTrait()
+
+    _default_value = 99 if six.PY3 else long(99)
+    _good_values   = [5, 10L]
+    _bad_values    = [4, -10L]
+
+
+class MaxBoundLongTrait(HasTraits):
+    value = Long(5 if six.PY3 else long(5), max=10)
+
+class TestMaxBoundLong(TraitTestBase):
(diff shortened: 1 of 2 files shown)


minrk:
Looks great to me. The fact that it's a utility function points to us not really having the right inheritance in our numerical traits, but I think it's a fine solution given how things are organized.

pbugnion:
This is now ready for reviewing. The following traitlets now support bounds checking:

 - Int and CInt
 - Integer
 - Long and CLong
 - Float and CFloat

In C{Int,Long}, the max bounds check may be a little surprising:

```python
class SomeClass(HasTraits):
    trait = CInt(5, max=10)

a = SomeClass()
a.trait = 10.5 # does not raise
a.trait == 10 # true
```

I think this is the right behaviour, since the *validated* value of `a` is within bounds, but I'm happy to implement a different behaviour if people don't think this is correct.

pbugnion:
Re. the utility function, I'm happy to try and factor out a superclass or a mixin for numeric traits if people think this would be useful (in a separate PR). I can open an issue with some design ideas.

minrk:
Thanks. Feel free to open a PR for the numerical base class if you are interested. These are extremely slow-moving classes (I think adding bounds check was the first change in a few years), so it may not be worth the effort to refactor.

PR #267, opened 2016-08-05
Alternative automatic documentation in the metaclass

fwkoch:
This is another approach to generating sphinx docs for `HasTraits` subclasses. It is an alternative to https://github.com/ipython/traitlets/pull/261

Documentation is generated at class definition by introspecting the class dictionary in the `MetaHasTraits` metaclass, giving something like this:

screen shot 2016-08-05 at 12 22 15 pm

Right now the functionality is quite basic, but it could be expanded pretty easily:
- Rather than getting defined in the metaclass, `sphinx` could be a function of `TraitTypes`. That would allow custom documentation for different subclasses (e.g. `Instance` traits could add their `klass` to the docstring)
- Documentation isn't restricted to TraitTypes. We can introspect anything else in the `HasTraits` subclass (e.g. event handlers, or documentation-specific variables - something like `if classdict['_autodoc'] == False:`, then don't generate any docs...).
- You could also interrogate the environment to dynamically decide what format of docstrings to generate (for example, if you are in the jupyter notebook generate napoleon docs rather than sphinx docs)

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -763,6 +763,27 @@ class MetaHasDescriptors(type):
 class MetaHasTraits(MetaHasDescriptors):
     """A metaclass for HasTraits."""
 
+    def __new__(mcs, name, bases, classdict):
+        def sphinx(trait_name, trait):
+            return (
+                ':param {name}: {doc}\n:type {name}: :class:`{cls}`'.format(
+                    name=trait_name,
+                    doc=trait.help,
+                    cls=trait.__class__.__name__
+                )
+            )
+
+        doc_str = classdict.get('__doc__', '')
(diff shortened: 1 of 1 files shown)


Carreau:
Tagging as `closed-PR`, and closing as this is seem to have stalled a bit. 
This is just to keep the PR queue a bit shorter and focusing on what is actually worked on.
Feel free to reopen, (or ask for reopening if you don't have enough right to do so). 
If you were not the original author, and want to revive this, please feel free to do so. 

Also, we are doing our best to enable people to learn how to contribute to the project, and get commits in. If you are missing time or have any other issues that prevent you to work on that, please feel free to tell us, we can try to build on top of what you did. 

Thanks a lot and looking forward to see this moving forward again.

PR #269, opened 2016-08-10
Test for sys.executable before trying to use it since it can be None.

infokiller:
See also:
https://docs.python.org/2/library/sys.html#sys.executable

Also, cleanup some unneeded whitespace.


minrk:
Thanks! In the future, it would be great if the unnecessary whitespace changes could be avoided.

infokiller:
Sorry for that!
Will avoid it next time.

Issue #270, opened 2016-08-10
Tutorial?


gvanrossum:
I tried to understand what this library does but all docs pointers seem to focus on advanced usages, and the README focuses on installation. What does simple use of this library actually look like? If the answer is "it's just like traits" that doesn't help me at all, I have no idea what that does either. (Or at least today I'm wearing the hat where I'm pretending not to. ;-)

SylvainCorlay:
@gvanrossum thanks for your interest!

We can certainly do better on narrative documentation. Would you be interested in us expanding on specific points here? (On the subject of PEP 487?)

gvanrossum:
I was just hoping for a tutorial that shows off a bunch of the common (simple) use cases. Is there really nothing you have for that? Not even a blog post somewhere? Then how do people get up to speed on using traitlets?

I'm not sure what you meant by the PEP 487 reference.

minrk:
For a long time, and until very recently, traitlets was just an internal utility as part of IPython, not meant for use outside the project. We split it off as part as a standalone package, but haven't managed to build up the documentation that a standalone package deserves.

For now, the closest approximation of a tutorial is [this documentation](http://traitlets.readthedocs.io/en/stable/using_traitlets.html) which goes through the basics of inheriting from HasTraits, defining a few traits and callbacks.

SylvainCorlay:
I mentioned the PEP 487 because traitlets were mentioned in the context of the discussion of this PEP.

Regarding existing tutorials, we have some material on [readthedocs](http://traitlets.readthedocs.io/en/latest/using_traitlets.html).

In short, traitlets let the user define classes that have
 1. Attributes (traits) with type checking and dynamically computed default values
 2. Traits emit change events when attributes are modified
 3. Traitlets perform some validation and allow coercion of new trait values on assignment. They also allow the user to define custom validation logic for attributes based on the value of other attributes.

Finally, the base `HasDescriptors` metaclass allows users to define other descriptors than trait types, without having to write metaclass code.  (We have used it to make a toy signals and slot library, where signature of slots are described with trait types). This is where traitlets are related with PEP 487 which should allow such usecases without any metaclass. (Besides, the main example provided in the PEP is a traits library.)

### Traitlets Usage

 - Attributes with type checking and default values

```python
import getpass

class Identity(HasTraits):
    username = Unicode()

    @default('username')
    def _default_username(self):
        return getpass.getuser()
```

```python
class Foo(HasTraits):
    bar = Int()

[...]

PR #272, opened 2016-08-15
Split using_traitlets into tutorial and api reference

minrk:
rather than interleaving reference and tutorial, which is off-putting for newcomers.

I think there's a lot more we can put in the using_traitlets tutorial,
but this starts with @SylvainCorlay's comment in #270, largely unmodified.


SylvainCorlay:
Thanks, this is a good start to change the structure into API reference and introduction.

PR #274, opened 2016-08-16
Union metadata switch

### LINKED ISSUE #244: Union trait does not handle trait metadata correctly
If I have a `HasTraits` class, called `Mix`, that has a `Union` trait, called `union`, then since the trait metadata of `union` is a class attribute, all `Mix` instances will have the same metadata for `union`, regardless of the actual trait type of `union`.

This can be harmful if you have a `Union` of traits with different `to_json` for example.

I hope this will be clearer with an example:
```python
from traitlets import Float, Unicode, HasTraits

class Mix(HasTraits):
    union = Float().tag(metadata="I'm a float") | Unicode().tag(metadata="I'm still a float, but with unicode metadata")
    
mix1 = Mix(union=0)
print mix1.union, mix1.trait_metadata('union', 'metadata')
mix2 = Mix(union='')
print mix1.union, mix1.trait_metadata('union', 'metadata')
```
outputs
```python
0.0 I'm a float
0.0 I'm still a float, but with unicode metadata
```

SylvainCorlay:
Fixes #244.

We need metadata to depend on the actual type validated in the case of the Union trait. For example, if we have

```python
class Foo(HasTraits):
    t = Date().tag(to_json=date_serialize) | Float()
```

the `date_serialize` custom serializer should be used only if the value is a date.

In this PR, we make `HasTraits.trait_metadata` return the metadata corresponding to the currently validated trait type in priority, then the metadata assigned to the Union trait itself.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1794,10 +1794,6 @@ class Union(TraitType):
         for trait_type in self.trait_types:
             trait_type.instance_init(obj)
         super(Union, self).instance_init(obj)
-        # Calling get once in instance_init forces the default value to be
-        # generated. This is required for the trait metadata for the initial
-        # value to be initialized
-        self.get(obj)
 
     def validate(self, obj, value):
         with obj.cross_validation_lock:
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -110,7 +110,7 @@ def _should_warn(key):
     env_flag = os.environ.get('TRAITLETS_ALL_DEPRECATIONS')
     if env_flag and env_flag != '0':
         return True
-    
+
     if key not in _deprecations_shown:
         _deprecations_shown.add(key)
         return True
@@ -354,7 +354,7 @@ dlink = directional_link
 
 
 #-----------------------------------------------------------------------------
-# Base Descriptor Class 
+# Base Descriptor Class
 #-----------------------------------------------------------------------------
 
 
@@ -416,7 +416,7 @@ class TraitType(BaseDescriptor):
     read_only = False
     info_text = 'any value'
 
-    def __init__(self, default_value=Undefined, allow_none=None, read_only=None, help=None, **metadata):
+    def __init__(self, default_value=Undefined, allow_none=None, read_only=None, help=None, **kwargs):
         """Declare a traitlet.
 
         If *allow_none* is True, None is a valid value in addition to any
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
cc @rmenegaux

traitlets/traitlets.py
```diff
@@ -1798,13 +1794,19 @@ def instance_init(self, obj):
         for trait_type in self.trait_types:
             trait_type.instance_init(obj)
         super(Union, self).instance_init(obj)
+        # Calling get once in instance_init forces the default value to be
+        # generated. This is required for the trait metadata for the initial
+        # value to be initialized
+        self.get(obj)
```
SylvainCorlay: see comment

traitlets/traitlets.py
```diff
@@ -434,14 +434,7 @@ def __init__(self, default_value=Undefined, allow_none=None, read_only=None, hel
             self.read_only = read_only
         self.help = help if help is not None else ''
 
-        if 'default' in metadata:
-            # Warn the user that they probably meant default_value.
-            warn(
```
minrk: Why have you removed this warning?

minrk:
Minor question about the removal of a warning that appears unrelated to the PR, but 👍 otherwise

traitlets/traitlets.py
SylvainCorlay: It seemed superfluous. Is this a `Traits` compatibility thing?

SylvainCorlay:
The renaming of metadata to kwargs is because since we added the `tag` function, kwargs are not metadata anymore (and anyway they had other things than metadata too)

traitlets/traitlets.py
```diff
@@ -1804,7 +1800,9 @@ def validate(self, obj, value):
             for trait_type in self.trait_types:
                 try:
                     v = trait_type._validate(obj, value)
-                    self.metadata = trait_type.metadata
+                    # In the case of an element trait, the name is None
+                    if self.name is not None:
+                        setattr(obj, '_' + self.name + '_metadata', trait_type.metadata)
```
SylvainCorlay: This is the gist of it.

traitlets/traitlets.py
```diff
@@ -1455,6 +1448,9 @@ def trait_metadata(self, traitname, key, default=None):
         except AttributeError:
             raise TraitError("Class %s does not have a trait named %s" %
                                 (self.__class__.__name__, traitname))
+        metadata_name = '_' + traitname + '_metadata'
+        if hasattr(self, metadata_name) and key in getattr(self, metadata_name):
+            return getattr(self, metadata_name).get(key, default)
```
SylvainCorlay: plus here.

SylvainCorlay:
merge?

minrk:
I asked about the removal of the warning for a likely typo. I don't believe anything about the circumstances of the warning have changed, so I don't see a reason to remove it.

traitlets/traitlets.py
minrk: Now that I see the warning below, double-warnings are probably superfluous.

PR #275, opened 2016-08-16
Remove init

SylvainCorlay:
Was this ever used?


rmorshea:
@SylvainCorlay, probably not. To be sure though, I think we'd have to look back and see if it had a purpose in the past.

SylvainCorlay:
This is something I noticed even before the split. I don't think it was ever used!

minrk:
Hasn't been used since well before traitlets became its own package

PR #276, opened 2016-08-16
Remove element_compare workaround

### LINKED ISSUE #277: Possible move of comparison to a separate function inside set
```python
try:
   silent = bool(old_value == new_value)
except:
   # if there is an error in comparing, default to notify
   silent = False
```
When `==` doesn't work (think `np.array` or `pd.DataFrame`), this is equivalent to failing silently in some sense. We should move this section to a separate compare function that can be overload by a `TraitType` so the entire logic of `set` doesn't need to be copied.

SylvainCorlay:
I guess that this was done for some ad-hoc numpy trait for which `==` does elementwise comparison.

However, it is a buggy behavior because is indicates such elements as changed anytime it is assigned a value even if it is equal to the previous one.

A fix for numpy trait types would be for them to overload `set` so as to use array comparison.

Closes #277

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -544,16 +544,8 @@ class TraitType(BaseDescriptor):
             old_value = obj._trait_values[self.name]
         except KeyError:
             old_value = self.default_value
-
         obj._trait_values[self.name] = new_value
-        try:
-            silent = bool(old_value == new_value)
-        except:
-            # if there is an error in comparing, default to notify
-            silent = False
-        if silent is not True:
-            # we explicitly compare silent to True just in case the equality
-            # comparison above returns something other than True/False
+        if new_value != old_value:
             obj._notify_trait(self.name, old_value, new_value)
 
     def __set__(self, obj, value):


rmorshea:
Would this inadvertantly break numpy traitlet implementations like this [numtraits](https://github.com/astrofrog/numtraits)?

dmadeka:
We should rather do the same (return `==`) with a compare function that can be overloaded right?

SylvainCorlay:
I think that we should rather simplify the traitlets code base and have numtraits and other numpy-based traits overload `set`...

SylvainCorlay:
ping @astrofrog

astrofrog:
This does indeed break ``numtraits`` - what about checking if ``!=`` returns an error and if it does, falling back to ``is not``?

astrofrog:
Otherwise, overloading ``set`` is ok, but brittle because it would involve having to include some of the code that accesses private variables. I like @dmadeka's suggestion of being able to overload specifically that comparison.

SylvainCorlay:
overload of set in bqplot by @dmadeka https://github.com/bloomberg/bqplot/pull/281

SylvainCorlay:
We are most likely not going to merge this now, 

but in any case, numtraits should already overload `set`  because with the current state of things, any assignment triggers a trait change event, even when it is equal...

astrofrog:
The problem with that kind of overload is that traitlets does not need to guarantee that ``_trait_values`` will exist in future and could modify it anytime (since not public).

dmadeka:
Im still +1 on `compare` because its a specific atomic operation that should be delegated to the child, however else `set` changes around it

SylvainCorlay:
All this because numpy 's `==` does not return a boolean.

astrofrog:
So thinking about this moreone alternative would be to overload ``set`` and ``get`` so that they store the array inside an object for which ``__eq__`` does return a boolean (and simply delegate to the original set/get methods)

rmorshea:
@dmadeka are you suggesting a `TraitType.compare` method that returns a boolean? If so, I think that would be preferable to forcing objects to be wrapped in an object that returns a bool on `==` as was suggested by @astrofrog.

cc: @SylvainCorlay

Carreau:

PR #280, opened 2016-08-18
docs: Add detailed API-reference with sphinx-automodule directive.

ankostis:
Made this PR for personal comfort (could not read the sources while travelling).

+ Augmented *API* section on the docs with `automodule` *sphinx* directive for ALL known modules,
listing both public & private members.
+ Fixed 2 minor documentation RsT glitches reported by sphinx on `traitlests` doc-strings.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -125,3 +125,67 @@ For example::
 
 
 
+Detailed API reference
+------------------------
+Module: :mod:`traitlets.traitlets`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.traitlets
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.config.configurable`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.config.configurable
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.config.application`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.config.application
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.config.loader`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.config.loader
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.config.manager`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.config.manager
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.log`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.log
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.utils.bunch`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.utils.bunch
+    :members:
+    :private-members:
+
+
+Module: :mod:`traitlets.utils.getargspec`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.utils.getargspec
+    :members:
+    :private-members:
+
+
+Module: :mod:`traitlets.utils.importstring`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.utils.importstring
+    :members:
+    :private-members:
+
+
+Module: :mod:`traitlets.utils.sentinel`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.utils.sentinel
+    :members:
+    :private-members:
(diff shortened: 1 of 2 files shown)


docs/source/api.rst
```diff
@@ -125,3 +125,67 @@ For example::
 
 
 
+Detailed API reference
+------------------------
+Module: :mod:`traitlets.traitlets`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.traitlets
+    :members:
+    :private-members:
```
minrk: private members should not be included in API docs, since they are private.

docs/source/api.rst
```diff
@@ -125,3 +125,67 @@ For example::
...
+
+Module: :mod:`traitlets.config.configurable`
```
minrk: individual traitlets.config.x submodules should not be included, only the top-level `traitlets.config` should be considered public API

docs/source/api.rst
```diff
@@ -125,3 +125,67 @@ For example::
...
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.log`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: traitlets.log
+    :members:
+    :private-members:
+
+Module: :mod:`traitlets.utils.bunch`
```
minrk: None of the utils should be included in generated API docs. These are private, internal utilities and not public APIs.

docs/source/api.rst
ankostis: I tried this at first.
Unfortunately *automodule* directive does not produce any output at all.

It can only work on a "real" module, not a *package* (i.e. `traitlets.config`).

docs/source/api.rst
ankostis: OK, will remove.

docs/source/api.rst
ankostis: I have to disagree with that.
Marking a method as private, conveys a strong enough signal to the consumer of the API, but we should not limit in addition her capability to understand the library. 

Sometimes reading the documentation (or looking at the sources) is needed to understand the operation of a higher-level function. For instance, I had to read `Configurable._load_config_files(filename, path)` to see that the `path` arg in `Configurable.load_config_file(filename.path)` can be a list.

ankostis:
By the way, is there a reason in `setup.py#L1: from distutils.core import setup` to prefer `distutils` over `setuptools`?

*Setuptools* is now the standard recommendation for python environments [1], and you can build docs using the command below, ensured that python path is set up correctly (whether installed in development mode or not):

    python setup.py build_sphinx

[1] https://packaging.python.org/current/#tool-recommendations

docs/source/api.rst
minrk: > Marking a method as private, conveys a strong enough signal to the consumer of the API, but we should not limit in addition her capability to understand the library.

This is not my experience.

> For instance, I had to read Configurable._load_config_files(filename, path) to see that the path arg in Configurable.load_config_file(filename.path) can be a list.

That is a failure of the documentation of the public API, not an indicator that the private implementation detail should be covered in the public documentation.

There are two *very* different goals of generated API docs:

1. how to use the library
2. how does the library work

Having a single document covering both causes confusion, and inevitably results in people confusing private APIs with public ones.

It is okay (debateable) to have generated documentation of the "internal implementation details" in the docs, but they should not be in the same place (especially not the same page) as the public, "how should I use this library" documentation. Given the general readability of Python, I find that the best balance is to only produce docs of the actual public APIs, and rely on reading source code for the internal implementation details.

docs/source/api.rst
minrk: In that case, we may need to not use automodule, and instead use:

```python
.. module:: traitlets.config

.. autoclass:: Configurable
```

It is important that these be documented under the correct module to import from.

minrk:
> By the way, is there a reason in setup.py#L1: from distutils.core import setup to prefer distutils over setuptools?

For historical reasons, largely having to do with the fact that setuptools does unacceptable egg-based installs by default, and corrupts sys.path as well, causing a decade or so of packaging nightmares. With setuptools (only starting with version 25) and pip, we are considering moving to use setuptools by default, only disabling the implicit building of eggs that should never happen. That said, since `pip install .` is the only way any Python package should ever be installed (`python setup.py install` should never by typed again), whether we import setuptools or not in setup.py doesn't make much difference.

docs/source/api.rst
minrk: In fact, I think this is generally a better approach - to carefully whitelist what APIs are considered public and documented. Liberal `automodule` has caused us many problems in the IPython project, and is specifically why we are trying something else on newer projects like this one. Each public API should be carefully and deliberately considered before committing it to the public API documentation.

ankostis: I see.
This PR were on purpose over-generic, to avoid mnually whitelisting stuff.
You may close it.

docs/source/api.rst
ankostis: This PR goes against exactly what you discribed.
Feel free to close it.

ankostis:
>  ... whether we import setuptools or not in setup.py doesn't make much difference.

Indeed, that is true for what you describe.
Still, if you install the project and its depencencies in develop mode using  `pip install -e .`, 
and try to run `make html`, it will fails to find dependencies. 
While with *setuptools*, you can launch doc-generation with `python setup.py build_sphinx`.

In general, nowdays  *setuptools* is the safe choice.

minrk:
> While with setuptools, you can launch doc-generation with python setup.py build_sphinx.

If this does installation with setuptools instead of pip, that's still a problem, and shouldn't happen. A requirements.txt in the docs directory handles this fine.

minrk:
Thanks for the PR! As you commented, I'll close here, but I agree that there should be more API coverage in our docs, just not quite this way. Feel free to open a new PR if you want to tackle a smaller expansion of the API docs.

ankostis:
Just a reminder: still `Configurables`/`Application` API remains undocumented in [the documents](https://traitlets.readthedocs.io/en/stable/api.html), and need to peek into the code.

PR #283, opened 2016-08-22
Document import_item

juhasch:


docs/source/utils.rst
```diff
@@ -0,0 +1,8 @@
+Utils
+=====
+
+.. module:: traitlets
+
+A simple utility to import something by its string name.
+
+.. autoclass:: import_item
```
minrk: I think this should be `autofunction`

juhasch: Changed it to `autofunction`.

PR #284, opened 2016-08-22
Fix HTML generation

juhasch:
Otherwise HTML generation using `make html` doesn't generate a code block.
(PyCharm says: Blank line is required after a literal block)


docs/source/config.rst
```diff
@@ -112,7 +112,6 @@ The answers to these questions are provided by the various
 :class:`~traitlets.config.Configurable` subclasses that an
 application uses. Let's look at how this would work for a simple configurable
 subclass::
-
```
minrk: I think the right fix here is to remove the `::`, not remove the blank line. The `::` is redundant with the `.. code-block:` below.

PR #288, opened 2016-08-24
Cross validation lock logic change

SylvainCorlay:
Adopting the same logic as in the `hold_trait_notifications` for the flag.

This context manager was introduced into #166

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1052,12 +1052,15 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
         At the end of the block, the lock's value is restored to its value
         prior to entering the block.
         """
-        original_value = self._cross_validation_lock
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Ping @ssanderson

traitlets/traitlets.py
```diff
@@ -1052,12 +1052,15 @@ def cross_validation_lock(self):
         At the end of the block, the lock's value is restored to its value
         prior to entering the block.
         """
-        original_value = self._cross_validation_lock
-        try:
-            self._cross_validation_lock = True
+        if self._cross_validation_lock is True:
```
minrk: Presumably this should be `if self._cross_validation_lock:`

SylvainCorlay: Fixed. There was another instance in the code which I removed too.

PR #291, opened 2016-08-25
Warning For Provisional Features

### LINKED ISSUE #281: Provisional Warnings - Addressing the Pace of Developement
With the pace of `traitlets` development in recent months, there's been a lot of concern over ease of adoption, and the ability of those who already have, to keep up with the each new release. To reduce this unease, It's been suggested that we implement "provisional warnings" which will raise an exception explaining the instability of various features by default, but that can be manually silenced with a context manager. 

The purpose of these warnings would be to create a space where `traitlets` can develop at a natural rate (fast or slow), but do so behind bold warning tape. A provisional warning will explain to any developer so bold as to bypass it, that the feature they are accessing is not guaranteed to exist over any period of time in the future, whether that be in the same form, or at all.

I hope to determine:
1. Whether provisional warnings are reasonable
2. What features ought to have provisional warnings
3. How provisional warnings should be implemented.

rmorshea:
closes #281 

a provisional warning:
- will always raise by default
- protects unstable or potentially temporary features
- is silenced if specifically referenced in a `silenced_provisional_warnings` context

basic usage:

``` python
try:
    warnings.warn(ProvisionalWarning("my feature"))
except ProvisionalWarning as e:
    # 'my feature' is a provisional feature - its use is not recommended
    print(e)

# specify the features to enable by name
with silenced_provisional_warnings("my feature") as w:
    warnings.warn(ProvisionalWarning("my feature"))

# a log of provisional warnings
print(w)
```

### DIFF
--- a/traitlets/utils/provisional_features.py
+++ b/traitlets/utils/provisional_features.py
@@ -1,0 +1,290 @@
+import sys
+import inspect
+import warnings
+from contextlib import contextmanager
+
+#-----------------------------------------------------------------------------
+# Warning Types
+#-----------------------------------------------------------------------------
+
+class LiabilityWarning(UserWarning): pass
+
+class ProvisionalWarning(FutureWarning):
+    """A warning for unstable or temporary features"""
+
+    template = "%r is a provisional feature - its use is not recommended"
+
+    def __init__(self, feature, *args):
+        """Warn users about the presence an unstable or temporary feature.
+
+        Provisional features are not guaranteed to exist over any period of time in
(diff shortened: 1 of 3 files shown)


minrk:
I don't think provisional warnings / context managers solve any real problems. Either people will use provisional features, in which case they will be just as frustrated if things break (being declared 'provisional' doesn't really change that), or (probably more likely) they won't be used until they make it into a stable release, in which case the provisional stuff may as well not exist until the provisional state is removed.

rmorshea:
This isn't going to happen.

PR #299, opened 2016-08-31
casting traits inherit from mixin

rmorshea:
Adds a generic `CastingMixin` class that all other casting `TraitTypes` inherit from (i.e. `CInt`, `CFloat`, etc.). Upon setting a value, these reconfigured traits first coerce the input value to the proper type, and then pass it to `super().validate(...)`. This mixin also makes new casting traits like `CDict` and `CList` more sensible to implement, because the complex base class validation heuristics don't need to be hard coded in their casting versions anymore:

```python
class A(HasTraits):
    cd = CDict()

a = A()
a.cd = [("a", 1)]
# prints: {'a': 1}
print(cd)
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1875,17 +1875,6 @@ class Int(TraitType):
         return _validate_bounds(self, obj, value)
 
 
-class CInt(Int):
-    """A casting version of the int trait."""
-
-    def validate(self, obj, value):
-        try:
-            value = int(value)
-        except:
-            self.error(obj, value)
-        return _validate_bounds(self, obj, value)
-
-
 if six.PY2:
     class Long(TraitType):
         """A long integer trait."""
@@ -1909,17 +1898,6 @@ if six.PY2:
 
         def validate(self, obj, value):
             value = self._validate_long(obj, value)
-            return _validate_bounds(self, obj, value)
-
-
-    class CLong(Long):
-        """A casting version of the long integer trait."""
-
-        def validate(self, obj, value):
-            try:
-                value = long(value)
-            except:
-                self.error(obj, value)
             return _validate_bounds(self, obj, value)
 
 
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -2689,3 +2602,101 @@ def info(self):
...
+        An attribute that can be assigned in subclasses
+        to limit what types of values are allowed to be
+        cast (this attribute is not defined in the base
+        class for mixin purposes).
+    """
+
+    cast_to = None
+
+    def validate(self, obj, value):
+        cast_types = getattr(self, '_cast_types', All)
```
minrk: This can be defined as the default on the mixin, rather than using getattr with a falllback

traitlets/traitlets.py
```diff
@@ -2689,3 +2602,101 @@ def info(self):
...
+        if cast_types is not All:
+            for cls in cast_types:
+                if isinstance(value, cls):
+                    break
+            else:
+                self.error(obj, value)
+        try:
+            value = self.cast_to(value)
```
minrk: Since this is a callable, I would prefer the verb `cast` to `cast_to`.

traitlets/traitlets.py
```diff
@@ -2689,3 +2602,101 @@ def info(self):
...
+        if hasattr(s, "validate"):
+            value = s.validate(obj, value)
+        return value
+
+    def info(self):
+        cast_types = getattr(self, '_cast_types', All)
+        if cast_types is not All:
+            return " or ".join(class_of(c) for c in set(cast_types).union([self.cast_to]))
+        else:
+            return "cast to %s" % (self.info_text or class_of(self.cast_to))
```
minrk: class_of isn't going to provide the right info for the cases where `cast` isn't a type (e.g. re.compile, below).

rmorshea:
Working on a more generic solution...

PR #302, opened 2016-09-02
Refresh sphinx docs

willingc:


docs/requirements.txt
```diff
@@ -1 +1,4 @@
-ipython_genutils
+ipython-genutils
+Sphinx
+sphinx-rtd-theme
+traitlets>=4.3.0b1
```
minrk: traitlets itself shouldn't be listed as a dependency in the docs requirements. That should be handled separately, otherwise you might get a version of traitlets installed that's not what is currently in the repo.

willingc:
@minrk I've added the ability to use conda to build. I have also removed traitlets from the requirements.txt.

PR #305, opened 2016-09-06
Fix error with documentation build on RTD

willingc:
Change dashes to underscores in package names for `environment.yml` and `requirements.txt`.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -4,6 +4,6 @@ channels:
   - conda
 dependencies:
   - python=3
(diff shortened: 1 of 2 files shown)


minrk:
Thanks! FWIW, pip does normalizing of `-,_`, so this kind of thing should be necessary for conda but not pip.

willingc:
@minrk Thanks for the tip. Didn't know that pip normalizes but makes sense why it failed when moving to conda.

PR #306, opened 2016-09-06
added a key_trait argument to Dict

### LINKED ISSUE #304: `Dict` trait is lacking a way to constrain all keys with a single trait
Coming from the original traits, I am used to `tr.Dict(TraitAppliedToKeys,TraitAppliedToValues)`. This functionality is lacking in traitlets. When describing a mathematical map, the two trait specifications would specify the domain and image of the map.
Obviously, this trait instead offers the functionality of something that I'd call _lightweight object specification_, where the keys are essentially fixed values. To me, that describes less of a mapping, but really more of a set of (typically named?) properties. Maaybe we should have two aliases `Mapping` and `PropertySet`, both describing instances of dict?

burnpanck:
Fixes #304


burnpanck:
I do think that constructor keywords are confusing as they are. I would propose to change `trait`->`value_trait` and `traits` to something more expressive, e.g. `per_key_traits`, `keyed_[value_]traits` or even `schema`.

traitlets/traitlets.py
```diff
@@ -2477,8 +2477,9 @@ def instance_init(self, obj):
 class Dict(Instance):
     """An instance of a Python dict."""
     _trait = None
+    _key_trait = None
 
-    def __init__(self, trait=None, traits=None, default_value=Undefined,
+    def __init__(self, trait=None, key_trait=None, traits=None, default_value=Undefined,
```
minrk: To avoid backward-incompatibility with positional args, can you add key_trait to the end?

minrk:
>I do think that constructor keywords are confusing as they are

If you want to add a value_trait arg and deprecate `trait`, that would be fine, too.

traitlets/traitlets.py
burnpanck: To be honest, I think this schema feature accessible via the `traits` argument doesn't belong into `Dict` at all. I'd rather not have it block access to `key_trait`. What about a compromise: Keep `key_trait` before `traits` and when an instance of `dict` (the python class, not the trait) is encountered in `key_trait`, raise a warning and map it to `traits`?

minrk: My preference would be for all of these after the first to be keyword-*only* arguments, but Python 2 doesn't make that very easy other than `**kwargs`. I would add it to the end and treat it as if it were keyword-only.

burnpanck: To me the keys of a dict are at least as important as the values if not more. I have the feeling everyone here seems to implicitly assume unicode keys for their dicts, but that is not validated in the current implementation. I'd put the key trait even before the value trait if not for backwards compatibility.

minrk: I'm concerned more with clarity than priority. I don't view the order as being significant (to me, there is only one positional arg for any TraitType, but Python 2 doesn't make this easy). If you'd like to make `value_trait` and `key_trait`, that would be fine by me, as long as it's done in a backward-compatible way (i.e. positional args retain their order). I wouldn't expect anyone to pass key_trait positionally ever, and I'm AOK encouraging people to pass value_trait by keyword-arg, as well, since that would be clearest.

burnpanck: I don't see how there is a general one positional arg that would have a well defined meaning for all TraitType's, so the positional args could equally well be trait specific. But then, I guess backwards compatibility shouldn't be broken lightly, so it's anyway too late. I'll shut up now and will just fix backwards compatibility.

minrk:
Thanks!

PR #313, opened 2016-09-13
Test on Python 3.6

minrk:
The `__dict__.pop` causes a catastrophic runaway memory leak *sometimes* on Python 3.6 (I couldn't reproduce it by hand, but it always happened at the same place in the test suite).

I think what I have here is a better way to accomplish what the problematic line was doing, so it's a net improvement, anyway.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1018,6 +1018,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
 
     def __getstate__(self):
         d = self.__dict__.copy()
+        d.pop('notify_change', None)
         # event handlers stored on an instance are
         # expected to be reinstantiated during a
         # recall of instance_init during __setstate__
@@ -1124,11 +1125,6 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
                 self.notify_change = notify_change
                 self._cross_validation_lock = False
 
-                if isinstance(notify_change, types.MethodType):
-                    # Presence of the notify_change method
-                    # on __dict__ can cause memory leaks
-                    # and prevents pickleability
-                    self.__dict__.pop('notify_change')
                 # trigger delayed notifications
                 for changes in cache.values():
                     for change in changes:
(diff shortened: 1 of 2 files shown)


traitlets/traitlets.py
```diff
@@ -1124,11 +1125,6 @@ def hold(change):
                 self.notify_change = notify_change
                 self._cross_validation_lock = False
 
-                if isinstance(notify_change, types.MethodType):
-                    # Presence of the notify_change method
-                    # on __dict__ can cause memory leaks
```
minrk: @rmorshea can you remember the details of the memory leak mentioned here? It turns out in Python 3.6 that this line *causes* a massive memory leak (as in all your machine's memory in a few seconds, clearly an upstream bug). The pickle issue I think is better fixed by the pop above in getstate, but I don't recall enough about the memory leak to verify that it's not an issue anymore.

SylvainCorlay:
@minrk did you report this upstream?

minrk:
@SylvainCorlay I'm planning to, but I wanted to understand it a bit more first, rather than "run this big test suite and run out of memory".

minrk:
Upstream Python bug reported, including proposed patch: http://bugs.python.org/issue28147

SylvainCorlay:
This may deserve a mailing list post. That would be unfortunate if your patch wasn't merged for 3.6.

minrk:
The Issue is pretty active already, so I'm optimistic.

traitlets/traitlets.py
rmorshea: @minrk sorry about the delay, I've been pretty swamped recently.

This was almost a year ago, but I'm pretty sure this originates with `hold_trait_notifications` where `HasTraits.notify_change` is set and restored:

```python
_ = self.notify_change

def hold_(*args):
	# ...
self.notify_change = hold

yield

self.notify_change = _
```

So you end up getting a circular reference.

However looking at it now, `__dict__.pop` isn't needed if we do this instead:

```python
self.notify_change = hold

yield

del self.notify_change
```

This will also resolve pickling.

rmorshea:
@minrk, I just want to verify that the memory was encountered while running the test suite. If that's the case, then #318 should solve both the pickling and circular reference issues.

Issue #315, opened 2016-09-17
Instance trait type default value confusion


rmorshea:
We raise here:

```python
class A(HasTraits):
	inst = Instance(Foo)

A().inst
```

But not here:

```python
class A(HasTraits):
	inst = Instance(Foo, allow_none=True)

A().inst
```

These seems magical, and is not at all implied by `allow_none=True`.

It seems like we should try to enforce this convention:

```python
class A(HasTraits):
	inst = Instance(Foo, default_value=None, allow_none=True)

A().inst
```

Or at the very least:

set `allow_none=True` if `default_value=None`

```python
class A(HasTraits):
	inst = Instance(Foo, default_value=None)

A().inst
```

rmorshea:
@minrk, this needs to be resolved before #317 is merged. However doing so will result in backward incompatible changes to subclasses of `Instance`, but similarly to #342, I feel that this use of "invalid" defaults via `make_dynamic_default` is an exploit. I will attempt to put together a comprehensive set of pr's across the notebook.

rmorshea:
@minrk, thoughts on this?

minrk:
> This seems magical, and is not at all implied by allow_none=True.

I don't agree with this. It's not magical to have default values for arguments.

I don't see an issue with the current behavior on this. A trait isn't validated until it gets a value (this is deliberate and important). Accessing that value the first time will cause its initialization, which if a value has not been assigned and the default value is not allowed, will raise.

I strongly disagree with this:

> We should probably try to enforce this convention:

> ```python
>  class A(HasTraits):
>      inst = Instance(Foo, default_value=None, allow_none=True)
> ```

That's an extremely verbose signature for a very common, supposedly simple case, especially one that was the default behavior for years. When simple, common things are complicated, the design is probably not right. This is an improvement:

```python
class A(HasTraits):
    inst = Instance(Foo, default_value=None)
```

but ultimately, not an improvement over the current behavior of:

```python
class A(HasTraits):
    inst = Instance(Foo, allow_none=True)
```

We also have to take into account that changing this would break *lots* of existing code throughout IPython, which we simply cannot do here except over multiple major releases and long deprecation cycles.

What I would *really* prefer is to switch Instance back to its original default `allow_none=True`, matching the longstanding behavior of `traits`. Switching to `allow_none=False` by default caused a lot of problems, and doesn't appear to have improved the experience of using traitlets.

rmorshea:
[...]

PR #317, opened 2016-09-18
[WIP] Enforce Logical Inheritance From MustBeInstance

rmorshea:
Many trait implementations perform the same typical validation check:

``` python
if isinstance(value, some_type):
    # value is of the right type
    return value
elif isinstance(value, casting_types):
    # value can be cast to the right type
    return some_type(value)
else:
    raise TraitError()
```

I've introduced a base class `MustBeInstance` which encapsulates this formula in a way such that most `TraitTypes` can subclass it, and hook into this kind of simple validation check. Everything from `Bool` to `List` now inherit from `MustBeInstance`.

Some example implementations:

``` python
# NumberBase simply enforces min
# and max values (e.g. `Int(min=0)`)
# and inherits from MustBeInstance
class Int(NumberBase):
    """An int trait."""

    klass = int
    default_value = 0

class Unicode(MustBeInstance):
    """A trait for unicode strings."""

    default_value = u''
    info_text = 'a unicode string'
    klass = six.text_type

    _cast_types = (bytes,)
    def _cast(self, value):
        try:
            return value.decode('ascii', 'strict')
        except UnicodeDecodeError:
            msg = "Could not decode {!r} for unicode trait '{}' of {} instance."
            raise TraitError(msg.format(value, self.name, nameit(obj)))
```

---

I've also added `nameit`, a new function for naming values - this is a substitute for `class_of`.

Indefinite naming:

``` python
>>> nameit(object())
'an object'
>>> nameit(object)
'an object'
>>> nameit(type(object))
'a type'
```

Definite naming:

``` python
 >>> nameit(object(), definite=True)
"the object at '0x10741f1b0'"
>>> nameit(object, definite=True)
"the type 'object'"
>>> nameit(type(object), definite=True)
"the type 'type'"
```

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1074,6 +1074,8 @@ class TestThis(TestCase):
 class TraitTestBase(TestCase):
     """A best testing class for basic trait types."""
 
+    bad_value_raises = TraitError
+
     def assign(self, value):
         self.obj.value = value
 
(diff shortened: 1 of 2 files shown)


minrk:
Seems largely simpler. What's the difference now between Instance and MustBeInstance?

rmorshea:
@minrk the difference comes down to `Instance.make_dynamic_default` and #315.

I should have linked to that issue before, because we could inherit from `Instance` instead of `MustBeInstance` if `Instance` traits with `allow_none=True` did not have a default value.

So whether we need `MustBeInstance` comes down to how we handle #315.

rmorshea:
Reworking this...

PR #289, opened 2016-08-24
config: generate sample-config files containing all and only classes relevant to configurable traits

ankostis:
The motivation is the user to know the relationships between the classes
in the generated sample configuration-file, so that he can choose at
which inheritance-level to override some trait value.

Issue #180, opened 2016-03-02
Documentation For Specifying Container Arguments


Lull3rSkat3r:
I have searched the docs, examples, and tests for an example of specifying a container type as a command line argument. Since I found nothing, I am assuming it is not supported at the moment. In particular I am trying to get a set of values as an argument. I tried the following:

```
myapp --containerarg={'A', 'B', 'C'}
myapp --containerarg=Set({'A', 'B', 'C'})
myapp --containerarg=['A', 'B', 'C']
myapp --containerarg='A' --containerarg='B' --containerarg='C'
```

Carreau:
IIRC, you'll have to quote each:

`myapp --Class.containerarg="{'A', 'B', 'C'}"` But it might depend on the shell and other things.

Lull3rSkat3r:
Works as expected. Might be worth documenting somewhere to help future users.

minrk:
I've hesitated to document container traitlets on the command-line, since it's so unpleasant it's practically not supported. The gist is: A string that will be eval'd.

It's this way because we originally parsed the command-line before knowing anything about the classes to be configured. This is still true *in general*, but to generate the help, we populate a list of known configurable classes for each Application. With this information, we could conceivably populate proper argparse constructs for container traitlets (and everything else, for that matter) of classes in the known-configurables list.

ankostis:
I agree the current situation with containers is unpleasant.

If you plan to correct it, I want to give an example of some notorious programs (java interpreter) that accept an arbitrary number of flags with constructs like that:

    java -Dfoo=1  -D bar=12 ...

I really miss this from *traitlets* so far. 
So I suppose by utilizing the list of known configurables, *argparse* `action='append'` behavior like above might be possible for containers, correct?
An alternative would be to use `argparse.add_args(nargs='*')` for list-traits.

Would a PR like this be in the right direction?
Or are you thinking to enhance flags/aliases somehow for achieving this?

ankostis:
@minrk any opinions on scanning all conf-classes to collect container-traits and append any matching aliases into `ArgParseConfigLoader` either with `action='append'` or `nargs='*'`?

minrk:
I'm a bit torn on doing it for flags/aliases, or for everything. I might lean ever so slightly toward just the flags & aliases, but the implementation may make it easier to apply to everything. I think we should do it, though.

ankostis:
Let's start with the aliases-only, where *argparse* lib is already used (for flags this proposal does not make sense).

[...]

PR #319, opened 2016-09-21
app: support flags/aliases given as (--long, -short) tuple

### LINKED ISSUE #296: Enhance `--help-all` to print all relevant  configurables, along with their relations
Now that an elaborate traversal of **"Configurable-inheritance"**  got merged (#289) for when generating the sample config-file (hopefully everything works all right),
is there a reason why the same logic is not used when printing the `--help-all`  message?
(i.e. to reduce the clutter in the terminal?)

Specifically in [traitlets/config/application.py#L384](https://github.com/ipython/traitlets/blob/master/traitlets/config/application.py#L384) where the classes to examine gets decided, there is this line:

    help_classes = self.classes

I would expect to find:

    help_classes =  = self._classes_inc_parents(self)  

but I've tested with even with the new `self._classes_in_config_sample()` method and it works fine.

A PR for this would be trivial, but now that `parent` is appropriately set on sub-commands (#292),
~~we can~~ *it may be possible to*  traverse the **"commands-chain"**, to provide also hints about them on the help & sample-config-file sections, like that:  

    [root-cmd  previous-sub-cmd]  *sub-cmd(parent1, parent2)

[*] It is the `sub-cmd` that is documented above.

ankostis:
+ Specify both short/long forms of flags & aliases at once.
+ Explain explicitely equivalence of flags/aliases.
+ Handle and log exceptions while printing flag/alias help-msgs - stacktrace was too convoluted otherwise.
+ FIX #296: print all relevant (sub)classes in help-msg.

An example of the new flags/aliases help message:

```
-f, --force
    Force various sub-commands perform their duties without complaints.
    Equivalent to: [Spec.force=True]
-v, --verbose
    Make various sub-commands increase their verbosity (not to be confused with --debug):
    Equivalent to: [Spec.verbose=True]
-d, --debug
    Equivalent to: [Cmd.raise_config_file_errors=True Cmd.print_config=True Spec.log_level=0 Application.log_level=0]
--log-level=<Enum>
    Default: 30
    Equivalent to: [Application.log_level]
```

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -326,11 +326,11 @@ class Application(SingletonConfigurable):
                                   for m in alias)
 
                 # reformat first line
-                fhelp[0] = fhelp[0].replace(longname, alias)
+                fhelp[0] = fhelp[0].replace('--'+longname, alias)
                 lines.extend(fhelp)
-                lines.append(indent("Equivalent to: [%s]" % longname))
+                lines.append(indent("Equivalent to: [--%s]" % longname))
             except Exception as ex:
-                self.log.error('Failed collecting help-message for aias %r, due to: %s',
+                self.log.error('Failed collecting help-message for alias %r, due to: %s',
                                alias, ex)
                 raise
 
@@ -352,7 +352,7 @@ class Application(SingletonConfigurable):
                               for m in flags)
                 lines.append(flags)
                 lines.append(indent(dedent(fhelp.strip())))
-                cfg_list = ' '.join('%s.%s=%s' %(clname, prop, val)
+                cfg_list = ' '.join('--%s.%s=%s' %(clname, prop, val)
                             for clname, props_dict
                             in cfg.items() for prop, val in props_dict.items())
                 cfg_txt = "Equivalent to: [%s]" % cfg_list
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -312,16 +312,26 @@ class Application(SingletonConfigurable):
                 classdict[c.__name__] = c
 
         for alias, longname in self.aliases.items():
-            classname, traitname = longname.split('.',1)
-            cls = classdict[classname]
+            try:
+                classname, traitname = longname.split('.',1)
+                cls = classdict[classname]
 
-            trait = cls.class_traits(config=True)[traitname]
-            help = cls.class_get_trait_help(trait).splitlines()
-            # reformat first line
-            help[0] = help[0].replace(longname, alias) + ' (%s)'%longname
-            if len(alias) == 1:
-                help[0] = help[0].replace('--%s='%alias, '-%s '%alias)
-            lines.extend(help)
+                trait = cls.class_traits(config=True)[traitname]
+                fhelp = cls.class_get_trait_help(trait).splitlines()
+
+                if not isinstance(alias, tuple):
+                    alias = (alias, )
+                alias = sorted(alias, key=len)
+                alias = ', '.join(('--%s' if len(m) > 1 else '-%s') % m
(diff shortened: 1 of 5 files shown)


ankostis:
Now it is ready to merge.
An example of the new flags/aliases help message:

```
-f, --force
    Force various sub-commands perform their duties without complaints.
    Equivalent to: [Spec.force=True]
-v, --verbose
    Make various sub-commands increase their verbosity (not to be confused with --debug):
    Equivalent to: [Spec.verbose=True]
-d, --debug
    Equivalent to: [Cmd.raise_config_file_errors=True Cmd.print_config=True Spec.log_level=0 Application.log_level=0]
--log-level=<Enum>
    Default: 30
    Equivalent to: [Application.log_level]
```

ankostis:
This PR tries to maintain compatibility with existing behavior, and adds single letter flags/alias prefixed both with single(`'-'`) and double(`'--'`) dashed.
The end result is that `(-v, --verbose)` gets added as `(-v, --v, --verbose)`.

I would prefer a simplification using the standard convention where a single-letter flag/alias are added only once, with single-dash.
What do you think?

traitlets/config/application.py
```diff
@@ -381,7 +404,7 @@ def print_help(self, classes=False):
         self.print_options()
 
         if classes:
-            help_classes = self.classes
+            help_classes = self._classes_in_config_sample()
```
minrk: Why is this not `self.classes`?

ankostis: See #289 discussion, and my example just entered in #180, specifically the comment in `classes` trait.

traitlets/config/application.py
```diff
@@ -312,16 +312,26 @@ def print_alias_help(self):
...
+                alias = sorted(alias, key=len)
+                alias = ', '.join(('--%s' if len(m) > 1 else '-%s') % m
+                                  for m in alias)
+
+                # reformat first line
+                fhelp[0] = fhelp[0].replace(longname, alias)
+                lines.extend(fhelp)
+                lines.append(indent("Equivalent to: [%s]" % longname))
+            except Exception as ex:
+                self.log.error('Failed collecting help-message for aias %r, due to: %s',
```
ankostis: While developing it, the stack-traces for such errors happened very early in the boot-strap process and it was extremely confusing for me to debug such issues (practically speaking); eventually I had to manually wrap the offending code in a try-catch every time; eventually, left it there for others, as explained in the commit message.

You can try it yourself if you have some class without `description` trait, or badly formed flag.
Maybe we can retrofit the error behavior to help developers.

minrk: Raising a clearer error sounds great. But I still think it should be an error, not a message that allows the application to continue. If you add `raise` below here, it should be fine.

traitlets/config/application.py
minrk: Makes sense, thanks.

traitlets/config/application.py
ankostis: Sorry, I will leave it because by catching exceptions, the log-message knows which property failed and it DOES help; it will just raise as you said.

minrk:
This is adding four `--` to aliases, so the output looks like:

    ----alias=<Unicode>

ankostis:
>This is adding four -- to aliases, so the output looks like:  `----alias=<Unicode>`

Thanks, the TC were not that accurate; fixed both.

ankostis:
Note, this has been merged (probably on 4.3.1) bust is not mentioned in the Changelog.

minrk:
This is not in 4.3.1, only unreleased master (will be 5.0).

PR #320, opened 2016-09-21
docs: Move changelog as /CHANGES.rst

ankostis:
The purpose is to allow a user visiting the project to see what has changed prima facie.


minrk:
I don't think so. The changes should be in the docs.

ankostis:
Ok then, it maybebe the other way around (/CHANGES --> `doc/changelog`).

Just a quick search for some github projects:

- pandas: `/RELEASE.md` referring to wiki.
- pip: `/CHANGES.txt`
- git: `/RelNotes` referring to `Documentation/RelNotes/` folder
- numpy: `/doc/release/`
- jupyter-client: `/docs/changelog.rst`
- sckit: `/doc/whats_new.rst`

It's better to have a changelog visible in the top-level, other projects include this section in README.

minrk:
If you want to add a top-level CHANGES that points people to the docs, that's fine.

ankostis:
That can be done, but it doen't make sense to be an RsT import directive - no code would evaluate it, that is why I did the other way around, where sphinx evaluate the import statement when generating docs.

minrk:
I meant have a single file whose only contents are "The changes are at docs/changelog.rst"

Carreau:
Thanks ! You haven't restored the content of `docs/source/changelog.rst` though, which is now empty except `.. include:: ../../CHANGES.rst` so we have an infinite recursion now :-)

ankostis:
Done.

But note that there is no point in having an RsT directive in the ./CHANGES.txt;  ths is a human readable file with just a line. It's only in the other way round where a directive would make sense.

Carreau:
> Done.
> 
> But note that there is no point in having an RsT directive in the ./CHANGES.txt; this is a human readable file with just a line. It's only in the other way round where a directive would make sense.

Yes, we understand that, the issue is that the `changelog.rst` file is autogenerated/autoupdated, and the builds of the docs is done by Readthedocs and rely on all the files being contained in the `docs` directory. While I do see the purpose of  `changes.rst` it would be misrepresenting what the current changes are and complexify the release documentation process if we had a top level documentation file.

PR #322, opened 2016-09-22
Use argparse to configure container-traits from command-line

ankostis:
- The idea was to  scan all conf-classes to collect `List/Set/Tuple/Dict` traits and use them to setup *argparse* for reading cmd-line args either with action='append' or nargs='*'`.
- Additionally, do the same for any matching aliases (for flags this does not make sense).
- But eventually had to feed all config-traits through *argparse* when classes are provided.
The reason is that any "regular"(not handled by *argparse*) trait with a name that is a prefix of some container-trait will get miss-classified as that container-trait (disabling this [`allow_abbrev`](https://docs.python.org/3/library/argparse.html#allow-abbrev) behavior is not supported in py2).

## Design
1. The `Application` on `parse_command_line()` uses its `classes` field to collects all relevant `HasTraits` classes and subclasses to push them into the `KVArgParseConfigLoader` constructor;
2. The `KVArgParseConfigLoader.load_config()` invokes `_add_arguments()` for all *list/Set/tuple* traits found on the collected classes , and appends them as *argparse* arguments according to the `multiplicity` metadata found on these traits:

    1. `multiplicity == 'append'`: use `add_argument(..., action='append')` (default if metadata missing);
    2.  `multiplicity == [ '*' | '+' | N ]`: use `add_argument(..., nargs=multiplicity)`.

3. During the above step, all container traits are collected in the new loder's field `argparse_traits` to be used in the next step to create the appropriate config-tree instance.
4. After *argparse* has parsed cmd-line arguments, the `KVArgParseConfigLoader._convert_to_config()` py-evaluates any lists-of-values (or list-of-tuples for `Dict` traits) one by one before creating and assigning the appropriate instance (i.e. `dict` for `Dict` traits) into the returned config-instance .

## Example
Assuming this 2 configurables + 1 app:

```python
class Base(Configurable): 
    a = List([]).tag(config=True)

class Sub(Base): 
    atuple = Tuple([]).tag(config=True, multiplicity='*')
    aset = Set(set()).tag(config=True, multiplicity='append')

class Cmd(Application):
    classes = [Sub]  ## Note on #289: frequently list given like that.
    aliases = {
        'a':  'Base.a', 
        'aa': 'Sub.a', 
        't':  'Sub.atuple', 
        's':  'Sub.aset'
    }
```

Then, in command-line it should be possible to enter all these commands:

```
cmd  --Base.a=1 -a=foo      ## Base.a = [1, 'foo']
cmd  --aa=1  --Sub.a=foo    ## Sub.a  = [1, 'foo']
cmd  -t 1 ab cd 4  --Sub.aset a  -s b  -s a  ## Sub.b = (1, 'ab', 'cd', 4), Sub.set = {'a', 'b'}
cmd  --Base.a=1 --Base.a=2  --Sub.a=foo      ## Base.a = [1, 2], Sub.a = ['foo']
```
* Notice that in the last case, `Sub` class trumps over any `Base` values.

For a `Dict` trait, the cmd-line syntax is this:

      cmd   --Base.adict kk=vv -D=key1=v1  -D key2=True -Dkey3="'True'"

resulting in this:

    Base.adict = {'kk': 'v', 'key1': 'v1', 'key2': True, 'key3: 'True'}

## Future
- Merge parts of `KVArgParseConfigLoader` into `ArgParseConfigLoader` class - actually I would try to compress the *loader* hierarchy.

### DIFF
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -50,6 +50,7 @@ class Bar(Configurable):
     enabled = Bool(True, help="Enable bar.").tag(config=True)
     tb = Tuple(()).tag(config=True, multiplicity='*')
     aset = Set().tag(config=True, multiplicity='+')
+    adict = Dict().tag(config=True)
 
 
 class MyApp(Application):
@@ -68,6 +69,7 @@ class MyApp(Application):
                     'name' : 'Foo.name',
                     'la': 'Foo.la',
                     'tb': 'Bar.tb',
+                    'D': 'Bar.adict',
                     'enabled' : 'Bar.enabled',
                     'log-level' : 'Application.log_level',
                 })
@@ -131,6 +133,16 @@ class TestApplication(TestCase):
         app.init_bar()
         self.assertEqual(app.bar.aset, {'S1', 'S2'})
         self.assertEqual(app.bar.tb, ('AB', 2))
+
+    def test_config_dict_args(self):
+        app = MyApp()
+        app.parse_command_line("--Bar.adict k=1 -D=a=b -D 22=33".split())
+        config = app.config
+        self.assertDictEqual(config.Bar.adict,
+                {'k': 1, 'a': 'b', '22': 33})
+        app.init_bar()
+        self.assertEqual(app.bar.adict,
+                {'k': 1, 'a': 'b', '22': 33})
 
     def test_config_propagation(self):
         app = MyApp()
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -34,7 +34,7 @@ from traitlets.config.application import (
 
 from ipython_genutils.tempdir import TemporaryDirectory
 from traitlets.traitlets import (
-    Bool, Unicode, Integer, List, Dict
+    Bool, Unicode, Integer, List, Tuple, Set, Dict
 )
 
 class Foo(Configurable):
@@ -42,12 +42,14 @@ class Foo(Configurable):
     i = Integer(0, help="The integer i.").tag(config=True)
     j = Integer(1, help="The integer j.").tag(config=True)
     name = Unicode(u'Brian', help="First name.").tag(config=True)
-
+    la = List([]).tag(config=True)
 
 class Bar(Configurable):
 
     b = Integer(0, help="The integer b.").tag(config=True)
     enabled = Bool(True, help="Enable bar.").tag(config=True)
+    tb = Tuple(()).tag(config=True, multiplicity='*')
+    aset = Set().tag(config=True, multiplicity='+')
 
 
 class MyApp(Application):
@@ -64,6 +66,8 @@ class MyApp(Application):
                     ('fooi', 'i') : 'Foo.i',
                     ('j', 'fooj') : 'Foo.j',
                     'name' : 'Foo.name',
+                    'la': 'Foo.la',
+                    'tb': 'Bar.tb',
                     'enabled' : 'Bar.enabled',
                     'log-level' : 'Application.log_level',
                 })
@@ -114,6 +118,19 @@ class TestApplication(TestCase):
         self.assertEqual(config.Foo.j, 10)
         self.assertEqual(config.Bar.enabled, False)
         self.assertEqual(config.MyApp.log_level,50)
+
+    def test_config_seq_args(self):
+        app = MyApp()
+        app.parse_command_line("--la 1 --tb AB 2 --Foo.la=ab --Bar.aset S1 S2 S1".split())
+        config = app.config
+        self.assertEqual(config.Foo.la, [1, 'ab'])
+        self.assertEqual(config.Bar.tb, ['AB', 2])
+        self.assertEqual(config.Bar.aset, 'S1 S2 S1'.split())
+        app.init_foo()
+        self.assertEqual(app.foo.la, [1, 'ab'])
+        app.init_bar()
+        self.assertEqual(app.bar.aset, {'S1', 'S2'})
+        self.assertEqual(app.bar.tb, ('AB', 2))
 
     def test_config_propagation(self):
         app = MyApp()
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -13,7 +13,7 @@ import logging
 import os
 import sys
 from io import StringIO
-from unittest import TestCase
+from unittest import TestCase, skip
 
 try:
     from unittest import mock
@@ -43,6 +43,7 @@ class Foo(Configurable):
     j = Integer(1, help="The integer j.").tag(config=True)
     name = Unicode(u'Brian', help="First name.").tag(config=True)
     la = List([]).tag(config=True)
+    fdict = Dict().tag(config=True, multiplicity='+')
 
 class Bar(Configurable):
 
@@ -50,7 +51,7 @@ class Bar(Configurable):
     enabled = Bool(True, help="Enable bar.").tag(config=True)
     tb = Tuple(()).tag(config=True, multiplicity='*')
     aset = Set().tag(config=True, multiplicity='+')
-    adict = Dict().tag(config=True)
+    bdict = Dict().tag(config=True)
 
 
 class MyApp(Application):
@@ -69,7 +70,7 @@ class MyApp(Application):
                     'name' : 'Foo.name',
                     'la': 'Foo.la',
                     'tb': 'Bar.tb',
-                    'D': 'Bar.adict',
+                    'D': 'Bar.bdict',
                     'enabled' : 'Bar.enabled',
                     'log-level' : 'Application.log_level',
                 })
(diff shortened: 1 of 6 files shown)


ankostis:
@minrk At [`traitlets/traitlests.py#1378`](https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L1378) it tests whether a variable can be called; why is it using `type(var) is types.FunctionType` instead of `iscallable(var)`?

The same question applies for [all such tests](http://stackoverflow.com/questions/624926/how-to-detect-whether-a-python-variable-is-a-function).

minrk:
@ankostis I suspect checking for `callable` is the right thing to do in those cases. I've updated the checks in #323. That's a real old line, probably from the time when Python 3 had removed `callable`, which it's since put back.

ankostis:
@minrk  PLEASE REVIEW the code + specs in the OP, to eventual become material for the manual.

ankostis:
PR works fine - maybe some TCs could also test more fringe cases, e.g. what happens when a List trait without defaults and without `allow_null` does not receive cmd-line values, etc.
But I had some mild problems extending it to Dict-traits.

In general, for a next release it might make more sense to drop the "mixed" parsing of args when conf-classes are available and delegate everything to *arparse*.  
Specifically to rely on `type` functions of `ArgumentParser.add_argument()` method instead of `_exec_config_str()`, and introduce a new method that assembles all the collected values (list, dictionary) for some config-trait.
The later was the problem I mentoned for extending this PR also for Dict-traits.

By the way, why in [loader.py#L517](https://github.com/ipython/traitlets/blob/master/traitlets/config/loader.py#L517) it uses eval to set the Config tree?

ankostis:
ankostis asked:
> By the way, why in [loader.py#L517](https://github.com/ipython/traitlets/blob/master/traitlets/config/loader.py#L517) it uses eval to set the Config tree?

Answering myself, I found 329db369 where @ellisonbg in 2009(!) did this change:

```diff
-            if v is not NoDefault:
-                setattr(self.config, k, v)
+            if v is not NoConfigDefault:
+                exec_str = 'self.config.' + k + '= v'
+                exec exec_str in locals(), globals()
```

unfortunately without explaining the reason - the only relative part of the commit message must be this:

> ...  The config system now supports config 
  subsections, like config.Global.  These sections, which must start
  with an uppercase char, are auto-created upon a getattr.

When using `setattr()` many TCs fail with this message:

```
  File "D:\Work\traitlets.git\traitlets\config\tests\test_loader.py", line 336, in test_seq_traits
    config = cl.load_config(argv, aliases=aliases)
  File "D:\Work\traitlets.git\traitlets\config\loader.py", line 761, in load_config
    self._convert_to_config()
  File "D:\Work\traitlets.git\traitlets\config\loader.py", line 903, in _convert_to_config
    self._exec_config_str(k, v, trait=trait)
  File "D:\Work\traitlets.git\traitlets\config\loader.py", line 529, in _exec_config_str
    setattr(self.config, lhs, value)
  File "D:\Work\traitlets.git\traitlets\config\loader.py", line 287, in __setattr__
    self.__setitem__(key, value)
  File "D:\Work\traitlets.git\traitlets\config\loader.py", line 272, in __setitem__
    'char must be Config instances: %r, %r' % (key, value))
ValueError: values whose keys begin with an uppercase char must be Config instances: 'CSub.e', [1, 'a', 'bcd']
```

Any explaination why?

ankostis:
With the latest commit, also `Dict` traits can be set by cmd-line *argparse*  options.
Design OP document updated.

minrk:
A simple setattr doesn't work because it's nested. It can be A.B.c. You could parse out the name and make a nested series of setattrs. I'm not sure that would help anything. I don't think it's relevant to this PR.

ankostis:
Thanks. So `eval` is not used for the `globals()`, and I can actually remove it; will re-write last commit.

You see, after my changes, in py2.7 I was getting [the following error](https://travis-ci.org/ankostis/traitlets/jobs/162436926) in `exec` line, and I had to add `locals()` and probably `globals()`.  My question was whether the `eval` was needed because internally the `config` object did something with the globals

```
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/_pytest/python.py:463: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=True)
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/py/_path/local.py:641: in pyimport
    __import__(modname)
traitlets/config/__init__.py:6: in <module>
    from .application import *
traitlets/config/application.py:19: in <module>
    from traitlets.config.configurable import Configurable, SingletonConfigurable
traitlets/config/configurable.py:12: in <module>
    from .loader import Config, LazyConfigValue, _is_section_key
E   SyntaxError: unqualified exec is not allowed in function '_exec_config_str' because it contains a nested function with free variables (loader.py, line 528)
```

ankostis:
@minrk there is a subtlety with the new code where half args are fed into *arparse* and the rest parsed as "unknown-args".  Since `ArgumentParser`  is configured with `allow_abbrev=True`, any regular trait with name that is a prefix of some container-trait will be miss-classified by argparse as that container-trait.

Can be dealt:
- by disabling `allow_abbrev`, BUT the flag was introduced in >py3.5 and behavior cannot change in older pythons, or
- by feeding all traits into *argparse*, and keep abbreviations for all traits!

The later may be beneficial to the users of programs (i.e. Jupyter) that have long options, so abbreviations would be a real bonus to them.

In any case, I retrofitted app-TCs to detect it the problem.

ankostis:
Well, tried the "future" (7678ffc78, all traits parsed with *arparse*, the 2nd option above), and it worked like charm, except from 1 TC.  I had to disable the `test-app.test_ipython_cli_priority()` TC because of the following error: 

```
 File "D:\Work\traitlets.git\traitlets\config\tests\test_application.py", line 208, in test_ipython_cli_priority
    app.parse_command_line(['--v=cli'])
  File "<decorator-gen-4>", line 2, in parse_command_line
  File "D:\Work\traitlets.git\traitlets\config\application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  ... 
  File "D:\Apps\WinPython-64bit-3.5.2.1\python-3.5.2.amd64\lib\argparse.py", line 1497, in _check_conflict
    conflict_handler(action, confl_optionals)
  File "D:\Apps\WinPython-64bit-3.5.2.1\python-3.5.2.amd64\lib\argparse.py", line 1506, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument --TestApp.config_file_loaded: conflicting option string: --TestApp.config_file_loaded
```

The situation is a kind of magic: the code failing, till that point is exactly the same with the previous TC.  Actually there is this comment on the TC:

>  this test is almost entirely redundant with above,
> but we can keep it around in case of subtle issues creeping into
> the exact sequence IPython follows.

@minrk  any ideas why it is failing?

traitlets/config/tests/test_application.py
```diff
@@ -179,6 +186,7 @@ class TestApp(Application):
             assert app.config.TestApp.value == 'cli'
             assert app.value == 'cli'
 
+    @skip('TO DEL? see https://github.com/ipython/traitlets/pull/322')
```
minrk: We need to make sure this test still passes. If it doesn't, then something is broken that needs to be fixed.

minrk:
I'm not sure where argparse is getting the conflicting string, that's a bit baffling to me. But it is pretty important that this test passes without modification to the test case.

I'll try to think about what's going on.

minrk:
#325 fixes the issue with the failing test. It was an unrelated bug that this PR triggered a test for. Yay!

ankostis:
@minrk wrote:
> I'm not sure where argparse is getting the conflicting string, that's a bit baffling to me. 

Look at the conflicting definitions of [`Bar.b`](https://github.com/ankostis/traitlets/blob/c1453515eb360b7aea575261136ef32d407ade70/traitlets/config/tests/test_application.py#L50) and [`Bar.bdict`](https://github.com/ankostis/traitlets/blob/c1453515eb360b7aea575261136ef32d407ade70/traitlets/config/tests/test_application.py#L54) on `traitlets/config/tests/test_application.py`.  If user specifies `Bar.b=1`, *arparse* assumes abbreviation of `Bar.bdict=1` which of course is not what we want.


> But it is pretty important that this test passes without modification to the test case.

PR rebased on latest master to get #325, and still `test-app.test_ipython_cli_priority()` TC fails.
See https://travis-ci.org/ankostis/traitlets/jobs/162732578

minrk:
I hadn't merged #325 yet, so rebasing wouldn't have picked it up. It's merged now, though.

ankostis:
Indeed, the problem is fixed, TC passes.

minrk:
I've tested and this is working great, thanks!

Pinging @jhamrick, who I believe has used the current `--Class.trait="['complicated', 'list', 'config']"` syntax before. This should make it much easier to specify list traits on the command-line, but I think it's going to be quite difficult to keep backward-compatibility with that syntax.

ankostis:
@minrk @jhamrick please check the OP, material that can be included in the docs.

jhamrick:
Cool! I had basically given up on trying to use lists and dictionaries on the command line, so this won't cause any problems for me. Thanks! 🎉

ankostis:
Since you're both reviewing, a note from my side for the multiple places where cmd-line values are parsed: 

- once by *argparse* when extracting them out from `argv` (for `Dict`-traits, this entails matching them against the [`_kv_opt_pattern` regex ](https://github.com/ankostis/traitlets/blob/list_args/traitlets/config/loader.py#L794)),
- then by [`ast.literal_eval()](https://github.com/ankostis/traitlets/blob/list_args/traitlets/config/loader.py#L506), for each `Container`-item or `Dict`-value, and finally,
- by the `Trait` that is the receiver of the value, when assign the `Config` object holding those values.

Maybe the places where these parsings take place can be centralized a bit, regarding the Loader hierarchy of classes (see my *Future* section on the opening pos)t.

ankostis:
Is there any plans for releasing this?

minrk:
@ankostis not immediately, we still have a few things to get through.

PR #323, opened 2016-09-23
use simpler callable check for meta_eval

minrk:
instead of `type() is FunctionType`

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1381,7 +1381,7 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
         result = {}
         for name, trait in traits.items():
             for meta_name, meta_eval in metadata.items():
-                if type(meta_eval) is not types.FunctionType:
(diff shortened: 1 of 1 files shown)


ankostis:
@minrk now that you are at it, you can change the following docstring to explain that plain-values as filters are also supported?

```python
    def traits(self, **metadata):
        """Get a ``dict`` of all the traits of this class.  The dictionary
        is keyed on the name and the values are the TraitType objects.

        The TraitTypes returned don't know anything about the values
        that the various HasTrait's instances are holding.

        The metadata kwargs allow to filter traits based on metadata values.
        Each keyword might be a callable or a plain value.

        - If it is a callable, the signature must be::

              filter(metadata_value) -> bool

          Return False to skip the trait from the output. If a metadata key
          doesn't exist, None will be passed to the function.

        - Otherwise, the metadata named as the key is checked for equality
          against its value, and included in the output if they match.

          Example:

              ## Fetch only traits with `config=True` metadata.
              has_traits.traits(config=True)
        """
```

PR #326, opened 2016-09-27
changes for 4.3.1

minrk:
Mainly for Python 3.6 workaround.

I'll release once this is merged.


Carreau:
Thanks !

minrk:
Released 4.3.1, which should allow downstream projects to test on 3.6 more reliably.

PR #327, opened 2016-09-29
[WIP] Add Eventful Traits

rmorshea:
# Summary

An alternative to #278 

While #278 enables custom events to be built into `TraitType` subclasses, via custom classes with event hooks. This pull request exposes a generic `Eventful` base class for automatically creating event hooks for the desired type. Eventful subclasses need only defined these three class attributes:

+ `event_map` used to distribute notifications to `@observer` handlers:
	+ `{'setitem': '__setitem__'}` maps `obj.a[k] = v` to `@observer('a', type='setitem')`
+ `klass` (optional) the base class from which eventful classes are created:
	+ `klass` can also by inferred from the type of a default value.
	+ To create an eventful dictionary `klass` should be `dict`
+ `type_name` the name of the eventful type:
	+ `type(obj.a).__name__ == type_name`

# Callbacks & Redirects
	
To make callbacks that trigger before, and after an event (hereafter referred to as beforebacks and afterbacks respectively) define methods `_before_<event-type>` or `_after_<event-type>` on `Eventful` subclasses (e.g. `_before_setitem` and `_after_setitem`).

### Beforebacks
+ Signature: `(inst, call)`
	1. `inst`: the value to be set on the trait.
	2. `call`: a bunch with the keys:
		+ `name`: name of the method called
		+ `args`: the arguments that method was called with
		+ `kwargs`: the keyword arguments the method was called with
+ Return: `None`, value, or an afterback
	1. If `None` no notifications are distributed to `@observer` handlers.
	1. The value is distributed to `@observer` handlers under `change['event']`
	2. The afterback is a callable with a signature `(returned)`
		+ `returned is the output of the called method.
		+ The afterback returns `None` to prevent notifications, or a value as in [ii].

### Afterbacks
+ Signature: `(inst, answer)`
	1. `inst`: the value to be set on the trait.
	2. `call`: a bunch with the keys:
		+ `name`: name of the method called
		+ `value`: the value returned by that call
		+ `before`: if a beforeback was defined, this is the value it returned.
+ Return: `None` or a value
	1. `None` prevents notifications
	2. Any other value gets distributed to `@observer` handlers under `change['event']`

Beforebacks and Afterbacks give you the tools to create your notifications, however it's often the case that an object's method acts like it wraps several other method calls (e.g. `dict.update` acts like it wraps calls to `dict.__setitem__`). To accommodate this there are `Redirect` objects which are callable and can be returned by a beforeback in order to trigger one or more notifiers.

# Examples

We can easily create an `Eventful` subclass which notifies when new items are assigned:

```python
class EItems(Eventful):
    
    # we will infer the class from a default value
    event_map = {'setitem': ('__setitem__')}
    type_name = 'eitems'
    
    @staticmethod
    def _before_setitem(inst, call):
        """Expect call.args[0] = key"""
        key, = call.args[0]
        try:
            old = inst[key]
         except:
            old = Undefined
        return key, old
    
    @staticmethod
    def _after_setitem(inst, answer):
        key, old = answer.before
        if inst[key] != old:
            return Bunch(key=key, old=old, new=new)
```

We also include basic implementations for eventful lists and eventful dictionaries:

+ [eventful dictionary](https://github.com/rmorshea/traitlets/blob/61918970a51c2abdb4d0164148975a3da5016365/traitlets/eventful.py#L190-L223)
+ [eventful list](https://github.com/rmorshea/traitlets/blob/61918970a51c2abdb4d0164148975a3da5016365/traitlets/eventful.py#L226-L271)

### DIFF
--- a/.cache/v/cache/lastfailed
+++ b/.cache/v/cache/lastfailed
@@ -1,0 +1,4 @@
+{
+  "traitlets/config/tests/test_loader.py": true,
+  "traitlets/tests/test_traitlets.py": true
+}
--- a/.cache/v/cache/lastfailed
+++ b/.cache/v/cache/lastfailed
@@ -1,4 +1,0 @@
-{
-  "traitlets/config/tests/test_loader.py": true,
-  "traitlets/tests/test_traitlets.py": true
-}
--- a/traitlets/__init__.py
+++ b/traitlets/__init__.py
@@ -1,3 +1,4 @@
 from .traitlets import *
+from .eventful import *
 from .utils.importstring import import_item
 from ._version import version_info, __version__
--- a/traitlets/eventful.py
+++ b/traitlets/eventful.py
@@ -1,40 +1,34 @@
 from .utils.bunch import Bunch
 from contextlib import contextmanager
 from .utils.spectate import watched_type
-from .traitlets import (TraitType, Dict,
-    List, Undefined, equivalent, Set)
+from .traitlets import (TraitType, Dict, List,
+    Undefined, equivalent, Set, TraitError)
 
-class Beforeback(object):
 
-    def __init__(self, trait, etype, before):
+class Callback(object):
+
+    def __init__(self, owner, trait, etype, callback):
+        self.owner = owner
         self.trait = trait
         self.etype = etype
-        self.before = before
+        self.func = callback
+
+
+class Beforeback(Callback):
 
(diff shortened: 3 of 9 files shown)


rmorshea:
Close in favor of https://github.com/ipython/ipywidgets/pull/1071

PR #332, opened 2016-10-21
Create All Default Values With Generators

### LINKED ISSUE #256: how to extend dynamic defaults with super
With traitlets 4.0, it's easy to extend dynamic defaults of parent classes, because the name of the default generator is part of the public API:

``` python
class SubClass(Parent):
    def _some_trait_default(self):
        previous_default = super()._trait_default()
        return modified(previous_default)
```

Now that the name of the method that happens to implement the default generator is no longer a public API, what's the best way to determine what the dynamic default _would have been_ in the parent class, and then extend it in a subclass?

rmorshea:
Closes #256 

``` python
class MyInt(Int):
    def make_dynamic_default(self):
        if self.default_value != 0:
            return self.default_value
        else:
            return 10

class A(HasTraits):
    i = MyInt()

class B(A):
    i = MyInt(20)

class C(B):
    @default('i')
    def _i_default(self):
        return 30

grandparent = super(B, C)
parent = super(C, C)
child = C

c = C()

print(
    grandparent._trait_default_generators['i'](c),
    parent._trait_default_generators['i'](c),
    child._trait_default_generators['i'](c)
)
```

```
10, 20, 30
```

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -112,7 +112,7 @@ class Configurable(HasTraits):
         If I am Bar and my parent is Foo, and their parent is Tim,
         this will return merge following config sections, in this order::
 
-            [Bar, Foo.bar, Tim.Foo.Bar]
+            [Bar, Foo.Bar, Tim.Foo.Bar]
 
         With the last item being the highest priority.
         """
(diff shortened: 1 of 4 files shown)


rmorshea:
Default cross-validation exposes exploits in notebook via #342 as detailed in #315

SylvainCorlay:
@rmorshea this is the PR that breaks widgets (cf #363 ). Can you try instantiating e.g. an `IntSlider` thanks.

SylvainCorlay:
cc @minrk

PR #333, opened 2016-10-21
Add a Callable Trait

### LINKED ISSUE #321: Add a callable traitlet type
This would be useful when you need to pass in a function. I've run into this several times when writing JupyterHub spawners. Environment variables can be specified as a dict whose value can be callable in JupyterHub, for example. I'm implementing this in one of my spawners just now, and think it'll be useful in the core library too.

Would people be open to a patch that added a callable traitlet type? if so, should it support further validation? The only one I could think of that'll make sense in Python would be an arity check...

rmorshea:
Closes #321

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -2686,3 +2686,19 @@ class UseEnum(TraitType):
         if self.allow_none:
             return result + " or None"
         return result
+
+class Callable(TraitType):
+    """A trait which is callable.
+
+    Notes
+    -----
+    Classes are callable, as are instances
+    with a __call__() method."""
+
+    info_text = 'a callable'
+
+    def validate(self, obj, value):
+        if six.callable(value):
+            return value
+        else:
+            self.error(obj, value)


Carreau:
Thanks !

krinsman:
Is there anywhere to look to track the progress towards releasing 4.3.3? 

Despite this being merged nearly three years ago, `Callable` hasn't been included in any releases yet, so when installing traitlets from conda, it's not possible to import `Callable`. 

(Which is really frustrating since it would be _perfect_ for my use case. Yes the definition is simple and it would be possible to essentially copy-paste it into my code, and that's what I'll do, but of course it's not very elegant compared to `from traitlets import Callable`, nor "official".)

Carreau:
> Is there anywhere to look to track the progress towards releasing 4.3.3?

I'm starting to re-review all the work that has been done in the past 3 years and we're going to see how we want to move forward. Either backport some to 4.x and say that we give up on 5.x or do a 5.0.

PR #334, opened 2016-10-22
[WIP] update traitlets to avoid comparison to none

AndrewMicallef:


AndrewMicallef:
This was a quick, and in my minds nasty looking, change,  to the traitlets.py code because I got this Warning using ipywidgets in a python 3  notebook:

```
c:\program files\python35\lib\site-packages\traitlets\traitlets.py:565: FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.
  silent = bool(old_value == new_value)
```

rmorshea:
this would be cleaner:

```python
silent = old_value is new_value or bool(old_value == new_value)
```

SylvainCorlay:
If you are using traitlets for things like numpy arrays or pandas dataframe for which `==` has a nonstandard behavior, I recommend using [traittypes](https://github.com/jupyter-incubator/traittypes) where `set` is re-implemented in the proper ad-hoc way for these types.

PR #335, opened 2016-10-24
[WIP] Add Chained Validators

### LINKED ISSUE #298: Custom validator registration
I think that it would be interesting to push upstream the abiility that was added in jupyter-incubator's traittypes: registration of custom validators (non-cross) with a chaining `valid` method.

```python
from traitlets import HasTraits, TraitError
from traittypes import Array

def shape(*dimensions):
    def validator(trait, value):
        if value.shape != dimensions:
            raise TraitError('Expected an of shape %s and got and array with shape %s' % (dimensions, value.shape))
        else:
            return value
    return validator

class Foo(HasTraits):
    bar = Array(np.identity(2)).valid(shape(2, 2))
foo = Foo()

foo.bar = [1, 2]  # Should raise a TraitError
```

rmorshea:
Closes #298

Allows for validator functions via arguments and simple attribute validation via keywords.

``` python
class Z(object):
    def __init__(self, x, y):
        self.x, self.y = x, y

def valid_x(trait, value):
    if value.x != 1:
        raise TraitError("'x' must be 1")
    return value

class A(HasTraits):
    b = TraitType().allows(valid_x, y=1)

a = A()
a.b = Z(1, 1) # pass
a.b = Z(0, 1) # raise
a.b = Z(1, 0) # raise
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -426,6 +426,7 @@ class TraitType(BaseDescriptor):
         Extra metadata can be associated with the traitlet using the .tag() convenience method
         or by using the traitlet instance's .metadata dictionary.
         """
+        self._allows = []
         if default_value is not Undefined:
             self.default_value = default_value
         if allow_none:
@@ -587,6 +588,8 @@ class TraitType(BaseDescriptor):
             return value
         if hasattr(self, 'validate'):
             value = self.validate(obj, value)
+        for validator in self._allows:
+            value = validator(self, value)
         if obj._cross_validation_lock is False:
             value = self._cross_validate(obj, value)
         return value
@@ -660,6 +663,25 @@ class TraitType(BaseDescriptor):
 
         self.metadata.update(metadata)
         return self
+
+    def allows(self, *validators, **attributes):
+        for f in validators:
+            if f not in self._allows:
+                self._allows.append(f)
+        for k, v in attributes.items():
+            def has_valid_attr(trait, value):
+                try:
+                    attr = getattr(value, k)
+                except Exception as e:
+                    raise TraitError(e)
+                if attr != v:
+                    raise TraitError("The '%s' attribute of"
+                        " %s must be %s" % (k, value, v))
+                else:
+                    return value
+            self._allows.append(has_valid_attr)
+        return self
+
 
     def default_value_repr(self):
         return repr(self.default_value)
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -97,6 +97,25 @@ class TestTraitType(TestCase):
             tt = MyIntTT('bad default')
         self.assertRaises(TraitError, B)
 
+    def test_allows(self):
+        class Z(object):
+            def __init__(self, x, y):
+                self.x, self.y = x, y
+
+        def test_x(trait, value):
+            if getattr(value, 'x', None) != 1:
+                raise TraitError("'x' must be 1")
+            return value
+
+        class A(HasTraits):
+            b = TraitType().allows(test_x, y=1)
+
+        a = A()
+        self.assertRaises(TraitError, setattr, a, 'b', Z(0, 1))
+        self.assertRaises(TraitError, setattr, a, 'b', Z(1, 0))
+        # should pass chained validators
+        a.b = Z(1, 1)
+
     def test_info(self):
         class A(HasTraits):
             tt = TraitType


minrk:
cc @SylvainCorlay for review

SylvainCorlay:
I looked a bit at it earlier. I like the `allows` name. I will comment further inline re: the kw args.

rmorshea:
@SylvainCorlay, let me know if you think the attribute validation via `**kwargs` seems common enough to dedicate the signature space to. If not I can roll that one back.

SylvainCorlay:
OK, I don't really see the point of adding kwargs just to pin values.

If only certain values are valid, why not using those as default values and setting the trait to be readonly?

SylvainCorlay:
I like the function name `allows` more than `valid`. (validate was already used in the API).

rmorshea:
@SylvainCorlay, just to be clear, `**kwargs` are used to validate the *attributes* of a given value, not the value itself.

So in the case of a numpy array, one can easily validate shape:

```python
class A(object):
    x = Array().allows(shape=(2, 2))

a = A()
a.x = np.zeros((2, 2)) # pass
a.x = np.zeros((3, 3)) # raise
```

SylvainCorlay:
OK, I would prefer to go for simplicity on this and remove the keyword args validation.

rmorshea:
Got it. Plus, it would be easy to reintroduce if it proved to be a common enough use case.

burnpanck:
Have you considered making the `.allows` API validation-only as opposed to validation/coercion? I am convinced that traitlets would deeply profit from a more clear line between these two concepts. Obvjously, the exising API does not make that distinction, but now would be the moment at least for this new API. I am particularly concerned by the importance of the ordering of individual coercion functions, which is not only not required for validators, but prohibits any further extensions that would rely on the orthogonality of individual validators. This is especially important with respect to #336, where the ordering of child class validators vs. parent class validators currently cannot be specified. The current use-case for `.allows` is clearly validation, why not prevent any future headaches related to the ordering of validator functions by making sure they do not interact. And it even matches the name :-).

traitlets/traitlets.py
```diff
@@ -587,6 +588,8 @@ def _validate(self, obj, value):
             return value
         if hasattr(self, 'validate'):
             value = self.validate(obj, value)
+        for validator in self._allows:
+            value = validator(self, value)
```
burnpanck: Leave out the `value = ` part to restrict `allows` to validation and prevent issues related to the ordering of the validators.

rmorshea:
@burnpanck, errors due to chained coercion could be quite harry to resolve. Plus, if a user truly needs chained coercion they could always do so manually in `@validate`. @SylvainCorlay, thoughts? 

Concerning order agnosticism though, even without coercion, I think that there will be cases where it's valuable for a validator to assume certain things about the values it receives.

For #336 though, the order is well defined - parent validators are run first, then those of the child. Thus extending `allows` in a child would increase the granularity of validation rather than expanding scope.

burnpanck:
I agree, it is probably impossible to enforce order agnosticism in `allows`. But by ignoring the result value of the validators and sending each validator the same object at least, you would at least not encourage people to rely on such behaviour. Precisely because chained coercion problems are hard to resolve, leaving that option out would make users stop first to think about what they want. If someone really needs chained coercion, then he'd probably want a well designed API for that purpose, and he'd also give a usage example to base the API design choices on. However, by offering a coercion API here, it will be used as such, and eventually an issue of coercion composition (or even composition between coercion and validation) will show up. Trying to clean-up the API at that point is nearly impossible without breaking backward compatibility. 

I'd say that most of the uses of `.allows` would remain validation. If there is indeed a need for coercion too, why not add a separate `.coerce` API for these cases? @SylcainCorlay once mentioned in a similar discussion over at traittypes that traitlets probably would have separated coercion and validation if it started all over again, but is bound by backwards compatibility now. This is a new API, so now would be the time to make such design decisions!

minrk:
Maybe it would be better to only allow one callable here, to avoid the chained coercion issue? I can't really imagine a realistic use case for chaining validators that aren't solved in a clearer way by passing a single callable.

burnpanck:
I guess the point of this PR is to offer the option to compose multiple (often independent) validators, particularly in conjunction with #336, which I think is a good thing. My point is that the natural composition method is different for validation and coercion: Validators enforce constraints that are effectively composed via a logical `and`, while coercion are chained functions. Therefore I would like to suggest to separate these two concepts well enough in any new API. The nice thing about order free composition of validation allows orthogonal validation concepts to co-exist without knowing of each other. This has many merits for introspection/metaprogramming. Take for example an automatic GUI generation framework that translates traits to a UI: Knowing about the constraints allows it to reflect these constraints to the user. If there are many orthogonal constraints, it could make use of all known ones and ignore the rest. Take an automatic remote procedure call library that is able to expose traits on the proxy object: Knowing the constraints allows it to optimise the protocol. I have written and used such a library for the original traits, as well as a few other meta-programming tricks. Several times I had to work around poor separation of concepts. Traitlets has proven to value good API design (deprecating magic method names in favour of explicit decorators), why not continue that here?

minrk:
@burnpanck that makes sense, thanks. Combining coercion and validation was perhaps not the best choice.

I still haven't seen an actual use case for the chaining, though. Does anyone have a real example of code that would be improved by allowing chaining of validators like this? A single validator seems simpler, clearer, and just as useful, without any of the confusion introduced by chaining.

rmorshea:
@minrk, chained validators are useful when trying to modularize validation utilities.

However the only real improvement brought by `allows` is to readability.

Consider the following class definitions (deliberately shown without context):

```python
class A(HasTraits):
    
    x = Array()
    
    @validate('x')
    def _validate_x(self, trait, x):
        if not x.shape() == (1, 3):
            raise TraitError("Must have shape %s, not %s" % (s, x.shape()))
        if magnitude(x) < 5:
       	    raise TraitError("Must have magnitude less than 5")
        elif not np.cross(x, xaxis)/(mag(x)*mag(xaxis)) < np.sin(np.pi/4):
            raise TraitError("Not within %s radians of %s" % (np.pi/4, list(x)))
        else:
            return x
```

or

```python
class A(HasTraits):
    
    x = Array().allows(
    	shape((1, 3)),
    	shorter_than(5),
    	inside_cone(45, xaxis),
    )
```

The purpose of the latter class is clearly more evident than the first.

One could argue that this is also a valid approach:

```python
class A(HasTraits):
    
    x = Array()
    
    @validate('x')
    def _validate_x(self, trait, x):
        same_shape(x, (1, 3))
        shorter_than(x, 5)
        inside_cone(x, 45, xaxis)
        return x
```

But at this point, it seems like a worse version of `allows`. However this is, of course, subjective.

minrk:
While I recognize the theoretical advantage if you did want to specify complex requirements like that, what concerns me is that I have never seen a desire or need to do it in practice. Is anybody building traits that want complex, chained geometric validation?

SylvainCorlay:
In bqplot, we use chained validation (with the equivalent `valid` API from https://github.com/jupyter-incubator/traittypes)

https://github.com/bloomberg/bqplot/blob/master/bqplot/marks.py#L333

minrk:
Thanks for the example, @SylvainCorlay. I'm fine with this in principle, I just wanted to see a real use case because we have to be extremely conservative in this package, given that it underpins all of IPython/Jupyter, and has proven to be extremely sensitive to subtle changes.

What do you want to do about @burnpanck's proposal that validation via this new API should not do any coercion like other validators? Would that impede your ability to use it where you are already `.valid()`?

rmorshea:
@minrk it looks like there are some cases where coercion is used: [bqplot/traits.py](https://github.com/bloomberg/bqplot/blob/master/bqplot/traits.py#L166-L170)

SylvainCorlay:
We have used chained coercion in bqplot indeed.

However I don't think that there is an example where we have multiple of those validators that do coercion. Although we could imagine an example where we first coerce via squeeze, and then coerce values...

One point though: in all traitlets, coercion and validation come together. Would it make it less consistent to not have the same behavior here?

minrk:
> Would it make it less consistent to not have the same behavior here?

Yes, absolutely. I think the question is about whether we want to separate validation from coercion moving forward, and if so it makes sense to start with new APIs (i.e. deprecate validators that coerce). I'm fine with keeping the current validation concept.

Carreau:
Tagging as `closed-PR`, and closing as this is seem to have stalled a bit. 
This is just to keep the PR queue a bit shorter and focusing on what is actually worked on.
Feel free to reopen, (or ask for reopening if you don't have enough right to do so). 
If you were not the original author, and want to revive this, please feel free to do so. 

Also, we are doing our best to enable people to learn how to contribute to the project, and get commits in. If you are missing time or have any other issues that prevent you to work on that, please feel free to tell us, we can try to build on top of what you did. 

Thanks a lot and looking forward to see this moving forward again.

PR #336, opened 2016-10-24
Extendable TraitType Tags and Validators

rmorshea:
Depends on #335

Extend chained validators:

``` python
def less_than(n):
    def validator(trait, value):
        if value < n:
            return value
        else:
            raise TraitError("%s is not less than %s" % (value, n))
    return validator

def greater_than(n):
    def validator(trait, value):
        if value > n:
            return value
        else:
            raise TraitError("%s is not greater than %s" % (value, n))
    return validator

class A(HasTraits):
    i = Int(5).allows(less_than(10))

class B(A):
    i = Int(5, extends='allows').allows(greater_than(0))

a = A()
a.i = -5 # pass
a.i = 15 # raise

b = B()
b.i = -5 # raise
b.i = 15 # raise
```

Extend trait metadata:

``` python
class A(HasTraits):
    i = Int().tag(x=1)
class B(A):
    i = Int(extends='tags').tag(y=2)

B.i.metadata # {'x':1, 'y':2}
```

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -116,6 +116,43 @@ class TestTraitType(TestCase):
         # should pass chained validators
         a.b = Z(1, 1)
 
+    def test_extend_allows(self):
+        def less_than(n):
+            def validator(trait, value):
+                if value < n:
+                    return value
+                else:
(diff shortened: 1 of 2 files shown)


rmorshea:
superceded by https://github.com/ipython/traitlets/pull/429

PR #338, opened 2016-10-25
[proposal] Make new argparse container CLI opt-in

minrk:
to avoid a dependency bump causing backward-incompatible changes to downstream project APIs.

This would dramatically slow adoption of the new options, but it would allow downstream packages to take the step on their own terms.

Anywhere with common subclasses, such as `jupyter-core` would immediately run into the same problem, though.

cc @takluyver @Carreau @ankostis @ellisonbg @fperez @jhamrick

Another stab at backward-compatibility would be to try to detect users passing old-style args. We can't however, easily distinguish between users having passed old-style list args and actually asking for that string that looks a bit like a list via the new style.

My current inclination is to move to the new way, and detect old-style args, WARNING that the API changed, but not trying to keep the old arg working (which would break people's code trying to adopt the new API).

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -18,11 +18,13 @@ from decorator import decorator
 
 from traitlets.config.configurable import Configurable, SingletonConfigurable
 from traitlets.config.loader import (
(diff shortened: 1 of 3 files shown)


minrk:
Closing in favor of #340

PR #340, opened 2016-10-25
reintroduce deprecated support for old-style container args

minrk:
alternative to #338, we should only pick one.

The old, undocumented (I think?) way of specifying containers on the command-line continues to work with a deprecation message.

The way I've implemented deprecations here prevents new-style args that pass the 'looks like a literal' test. That means there would be no way to properly specify the list `["[1]"]` on the command-line. I'm not too concerned about it, but it's a theoretical issue. The alternative to _that_ is to keep the logic here for warnings, but _don't_ try to support the old-style args. That way you can specify anything in the new-style, and users will get a pointer if they use the old-style, but attempts to use the old-style will generally result in errors.

I have a slight inclination toward deprecate-and-keep-working over break-with-informative-message, but if someone can point to a more serious problem that the deprecation will break, I'm happy to change.

closes #338

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -16,8 +16,10 @@ from ast import literal_eval
 from ipython_genutils.path import filefind
 from ipython_genutils import py3compat
 from ipython_genutils.encoding import DEFAULT_ENCODING
-from six import text_type
-from traitlets.traitlets import (HasTraits, Container, List, Dict, Any)
+from six import text_type, string_types
+from traitlets.traitlets import (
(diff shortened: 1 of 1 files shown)


ankostis:
But if the old behavior was undocumented, and problematic, why should we keep it around?
Probably would cause more problems in the future if people start really (ab)using it.

minrk:
@ankostis while rare, people used it. Deprecations, where feasible, are always nicer than breaking things.

Carreau:
+1

minrk:
Thanks, @Carreau!

PR #341, opened 2016-11-01
add Application.show_config[_json]

minrk:
setting this value will skip Application start and show loaded config on stdout, instead.

if `show_config_json` is given, it will be json-formatted.

- [x] tests

Applications can inherit the default flags from `traitlets.config.application.default_flags`,
but this can be triggered immediately with:

    any-jupyter-command --Application.show_config=True

cc @rgbkrk

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -8,15 +8,12 @@ from __future__ import print_function
 
 from collections import defaultdict, OrderedDict
 from copy import deepcopy
-import io
 import json
 import logging
 import os
 import pprint
 import re
 import sys
-
-from decorator import decorator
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Yay, this is great! I'll use this for smoke testing a Docker image before promoting to production.

ankostis:
But why stop the app?
The option may just dump the configs in the logger (i.e. at INFO level), and continue normally.

This option is useful for every combination sub-commands and it was one of the first "extensions" I had built my self.

rgbkrk:
The reason I was hoping for something like this is to be able to run validation tests against Jupyter installations.

PR #345, opened 2016-11-10
fix union default - no need for builtin dynamic default

### LINKED ISSUE #344: Default value of Union is not respected
Ran into this today: it seems the specified default value for ``traitlets.Union`` is not respected.

```python
import traitlets as T

class Foo(T.HasTraits):
    unicode = T.Unicode(allow_none=True, default_value=None)
    unicode_or_list = T.Union([T.Unicode(), T.List(T.Unicode())],
                              allow_none=True, default_value=None)
    
f = Foo()
print(f.unicode)  # This is None as expected
print(f.unicode_or_list)  # This is an empty string, should be None?
```

The workaround is to specify the default value within the first entry of the Union's list, but I found this behavior surprising. Is it a bug?

rmorshea:
Closes #344 

No more need for dynamic default in `Union` as a result of #332.

Also fixes a test case that was inheriting from `HasTraits` instead of `TestCase`, which masked #344.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ __pycache__
 \#*#
 .#*
 .coverage
+.cache/v/cache/lastfailed
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1488,7 +1488,7 @@ class UnionListTrait(HasTraits):
 
     value = List(Int() | Bool())
 
-class TestUnionListTrait(HasTraits):
+class TestUnionListTrait(TraitTestBase):
 
     obj = UnionListTrait()
 
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1792,13 +1792,13 @@ class Union(TraitType):
         super(Union, self).__init__(**kwargs)
 
     def class_init(self, cls, name):
-        for trait_type in reversed(self.trait_types):
+        for trait_type in self.trait_types:
             # enfore priority of the first given traitytype
             trait_type.class_init(cls, None)
         super(Union, self).class_init(cls, name)
 
     def instance_init(self, obj):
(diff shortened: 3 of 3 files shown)


jakevdp:
Awesome – thanks!

SylvainCorlay:
@rmorshea the reversion of the trait types validation is a backward incompatible change (we rely on the current order in bqplot).

What is the reason for the inversion?

rmorshea:
@SylvainCorlay, seeing as validation prefers the first given `TraitType`, it seemed reasonable to remain consistent and prefer the default value of the first given `TraitType` as well.

I can make that a separate PR though since I didn't think about the compatibility when including that here.

rmorshea:
@SylvainCorlay, for some reason I thought that validation prefered the first given `TraitType`, and just trait to make it consistent. Clearly that's not the case, so I'll revert that.

rmorshea:
@SylvainCorlay, since validation prefers the first given `TraitType`, it seemed reasonable to prefer the default value of the first given `TraitType` as well.

I'll put that in a separate PR though since I didn't think about compatibility when including that here.

jakevdp:
If validation order is an important part of the API, it might be worth putting in a unit test that will catch if that ever breaks.

SylvainCorlay:
Ah now I understand. Sorry I was mistaken. I agree that it would be more logical for it to be reversed.

SylvainCorlay:
> If validation order is an important part of the API, it might be worth putting in a unit test that will catch if that ever breaks.

It is just that the first trait type that validate is the type that will be used.

rmorshea:
> it might be worth putting in a unit test that will catch if [union validation order] ever breaks.

#347

rmorshea:
Found a more fundamental bug. Will rework this shortly.

PR #346, opened 2016-11-10
invert class/instance initialize priority in union "subtraits"

rmorshea:
`Union` validation prefers the first given `TraitType` so it seems reasonable to remain consistent and prefer the default value of that first `TraitType` as well.

Right now, we prefer the default value of the last given `TraitType` which seems odd since that default will first have to pass validation from all the `TraitTypes` before it.


minrk:
Merged as c523781

PR #347, opened 2016-11-11
test union valdiation priority

rmorshea:
```python

class Foo(HasTraits):
    bar = Union([CInt(), Unicode()])
foo = Foo()
foo.bar = '1'
# validation in order of the TraitTypes given
assert foo.bar == 1
```


rmorshea:
@minrk rebased.

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
minrk: looks like this got added by accident

PR #354, opened 2016-12-08
Improve generated help and config messages

ankostis:
- Assume a "user" viewpoint when explaining how configurables work - avoiding traitlets terminology where not necessary.
- Use `====` and `---` headers to denote hierarchical sections in the help message.
- Don't add blanklines below section-headers, `===` should be enough;
 this makes also the separation with the next section more visible.
- Distinguish "class-parameters" from "class-options" in the help-messages.
- Print `Application.__doc__` if no `app.description`.

For facilitating comparison, I'm providing sample help-msgs BEFORE and AFTER:

BEFORE:
```
Subcommands
-----------

Subcommands are launched as `co2dice cmd [args]`. For information on using
subcommand 'cmd', do: `co2dice cmd -h`.

project
    Commands to administer the storage repo of TA *projects*.
report
    Extract the report parameters from the co2mpas input/output files, or from *current-project*.

Options
-------

Arguments that take values are actually convenience aliases to full
Configurables, whose aliases are listed on the help line. For more information
on full configurables, see '--help-all'.

-d, --debug
    Equivalent to: [--Spec.log_level=0 --Application.log_level=0 --Cmd.raise_config_file_errors=True --Cmd.print_config=True]

Class parameters
----------------

Parameters are set from command-line arguments of the form:
`--Class.trait=value`. This line is evaluated in Python, so simple expressions
are allowed, e.g.:: `--C.a='range(3)'` For setting C.a=[0,1,2].

Application(SingletonConfigurable) options
------------------------------------------
--Application.log_level=<Enum>
    Default: 30
```

AFTER:
```

Subcommands
===========
Subcommands are launched as `co2dice cmd [args]`. For information on using
subcommand 'cmd', do: `co2dice cmd -h`.

project
    Commands to administer the storage repo of TA *projects*.
report
    Extract the report parameters from the co2mpas input/output files, or from *current-project*.

Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

-d, --debug
    Equivalent to: [--Application.log_level=0 --Spec.log_level=0 --Cmd.print_config=True --Cmd.raise_config_file_errors=True]

Class options
=============
The command-line option below sets the respective configurable class-parameter:
    --Class.parameter=value
This line is evaluated in Python, so simple expressions are allowed.
For instance, to set `C.a=[0,1,2]`, you may type this:
    --C.a='range(3)'

Application(SingletonConfigurable) options
------------------------------------------
--Application.log_level=<Enum>
    Default: 30
```

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -33,15 +33,15 @@ import six
 #-----------------------------------------------------------------------------
 # merge flags&aliases into options
 option_description = """
-The options below are convenience aliases to configurable class parameters, 
-as listed in the "Equivalent to" line for each option, below.
-Use '--help-all' to see all configurable class-parameters for some command.
-""".strip() # trim newlines of front and back
+The options below are convenience aliases to configurable class-options,
+as listed in the "Equivalent to" description-line of the aliases.
+To see all configurable class-options for some <cmd>, use:
+    <cmd> --help-all
+""".strip()  # trim newlines of front and back
 
 keyvalue_description = """
-To set the configurable class parameters listed in the sections below, 
-use command-line arguments like this: 
-    --Class.trait=value
+The command-line option below sets the respective configurable class-parameter:
+    --Class.parameter=value
 This line is evaluated in Python, so simple expressions are allowed.
 For instance, to set `C.a=[0,1,2]`, you may type this:
     --C.a='range(3)' 
@@ -374,7 +374,6 @@ class Application(SingletonConfigurable):
             return
         lines = ['Options']
         lines.append('=' * len(lines[0]))
-        lines.append('')
         for p in wrap_paragraphs(self.option_description):
             lines.append(p)
             lines.append('')
@@ -390,7 +389,6 @@ class Application(SingletonConfigurable):
 
         lines = ["Subcommands"]
         lines.append('=' * len(lines[0]))
-        lines.append('')
         for p in wrap_paragraphs(self.subcommand_description.format(
                     app=self.name)):
             lines.append(p)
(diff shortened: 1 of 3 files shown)


traitlets/config/application.py
```diff
@@ -374,7 +374,6 @@ def print_options(self):
             return
         lines = ['Options']
         lines.append('=' * len(lines[0]))
-        lines.append('')
```
minrk: Can we have the blank lines back? I think it reads clearer when they are there, and I generally think it's always nice to have a blank line after a heading.

minrk:
Thanks! I think this is an improvement. I've asked for a couple small changes inline.

traitlets/config/application.py
```diff
@@ -6,46 +6,45 @@
...
 # Descriptions for the various sections
 #-----------------------------------------------------------------------------
-
 # merge flags&aliases into options
 option_description = """
-Arguments that take values are actually convenience aliases to full
-Configurables, whose aliases are listed on the help line. For more information
-on full configurables, see '--help-all'.
-""".strip() # trim newlines of front and back
+The options below are convenience aliases to configurable class-options,
```
ankostis: Well, my idea was to separate cmd-line *class-options* (e.g. `--Foo.bar`) from source-code *class-parameters* (e.g. the `bar` parameter of `Foo` class). 
So it is correct in this case because the next phrase refers to the "Equivalent to" desc-line, which indeed it contains *options*, not *parameters* (see PR text for comparison).

Another alternative would be to use *class-attributes* instead of *parameters` which is accurate, but I wouldn't dare such an invasive terminology change.

traitlets/config/application.py
ankostis: (I'm not sure if this comment is still valid)

The pyrpose behind this changes was to reduce the total number of lines printed in the console, to avoid scrolling.
Additionally, I wanted to apply the "Law of Proximity" design principle, and place the section-body closer to its header, in comparisons to the preceding section-body.

Is thatt OK?

ankostis:
Rebased on latest master f60e06f8b.

PR #360, opened 2017-01-11
Add factory method on Application for customizing the Loader parsing cmd-line args

ankostis:
Add capability to modify the the Loader used to parse cmd-lines used by `Application` class (currently the `KVArgParseConfigLoader` class) *without* fully reimplementing `Application.parse_command_line()` method.

### Rationale
With the above customization it is easier to modify the *argparse* parser, which has been enabled by #322.
A case where the above is needed arrives in projects with multiple sub-cmds, where some cmd-names clash, and argparse "conflicts" occur (when adding the same option, twice).  

While the [*argparse* constructor](https://docs.python.org/3.6/library/argparse.html#argumentparser-objects) offers the [`conflict_handler='resolve'`](https://docs.python.org/3.6/library/argparse.html#conflict-handler) strategy,and there exists an [`argparse` factory](/ipython/traitlets/blob/master/traitlets/config/loader.py#L802-L804) method on the Loder, there is no equivalent factory method on the Application for the Loader itself - that means that you cannot modify the former factory by switching Loader implementations.

An alternative would be to add all `*argparse_args` and `**argparse_kwds` on`Application` class, and pass them when constructing the Loader, which should then pass them to `argparse`, but a) this violates blatantly the Demeter law adding dead weight for most cases, and b) it should be *easier* to override the Loader, which might be necessary for other cases.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -33,15 +33,15 @@ import six
 #-----------------------------------------------------------------------------
 # merge flags&aliases into options
 option_description = """
-The options below are convenience aliases to configurable class parameters, 
-as listed in the "Equivalent to" line for each option, below.
-Use '--help-all' to see all configurable class-parameters for some command.
-""".strip() # trim newlines of front and back
+The options below are convenience aliases to configurable class-options,
+as listed in the "Equivalent to" description-line of the aliases.
+To see all configurable class-options for some <cmd>, use:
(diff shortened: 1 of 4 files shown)


ankostis:
(Just based on top of  #354, I can rebase it on master if asked)

minrk:
Makes sense, thanks

PR #361, opened 2017-01-13
help: Alias overriding help-text (like flags), options list human-help before automated infos

ankostis:
To keep the API symmetric to *flags*, add the capability to *aliases* to override the printed help-msg.

With this PR, the `Application.aliases` dict may contain *two-tuples* as values like this `(alias: (longname, help-text)}`:

```python
    aliases = Dict({
                    ('fooi', 'i') : 'Foo.i',                          # simple longname
                    ('j', 'fooj') : ('Foo.j', "`j` terse help msg"),  # longname, help-text
    )
```

- Renamed the use of builtin `help` in one occasion;
- changed 2 TCs to test with two-tuples; 
- improved slightly the documentation.

## Rational:
Currently code that prints the help-text for the *aliases* fetches the help-text from the aliased trait, and that is possible since there is a 1-1 correspondence (in contrast to *flags* where it is *1-n*, and thus had to introduce the two-tuple with help-text).
But since this might be long, the end result will be a long help message, scrolling past the end of the screen.
For flags, it is possible (actually forced) to provide *shorter* versions of the help-messages, and use `--help-all` to read all content, but this flexibility is impossible for aliases.

To facilitate comparison, the *before* and *after* help messages are provided for the `--config-paths` alias:

## "Before" Help Messages:
```
$ co2dice --help
prepare/sign/send/receive/validate & archive Type Approval sampling

Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

-d, --debug
    Equivalent to: [--Application.log_level=0 --Spec.log_level=0 --Cmd.raise_config_file_errors=True]
--config-paths=<list-item-1>...
    Default: []
    Absolute/relative path(s) to read "static" configurable parameters from.
    If false, and no `CO2DICE_CONFIG_PATH` envvar is defined, defaults to:
        ['D:\\Apps\\cygwin64\\home\\anastkn\\.co2dice\\co2dice_config', 'd:\\work\\compas.vinz\\co2mpas\\sampling\\co2dice_config']
    Multiple values may be given, and each value may be a single or multiple paths
    separated by ';'.  All paths collected are considered in descending order
    (1st one overrides the rest).
    For paths resolving to folders, the filename `co2dice_config.[json | py]` is appended;
    otherwise, any file-extensions are ignored, and '.py' and/or '.json' are loaded (in this order).
    Tip:
        Use `config init` sub-command to produce a skeleton of the config-file.

To see all available configurables, use `--help-all`
```

## "After" Help Message:
```
$ co2dice --help
prepare/sign/send/receive/validate & archive Type Approval sampling

Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

-d, --debug
    Equivalent to: [--Application.log_level=0 --Spec.log_level=0 Cmd.raise_config_file_errors=True]
--config-paths=<list-item-1>...
    Default: []
    Absolute/relative path(s) to read "static" configurable parameters from.

To see all available configurables, use `--help-all`
```

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -253,6 +253,13 @@ class Configurable(HasTraits):
             header = '%s=<%s>' % (header, trait.__class__.__name__)
         #header = "--%s.%s=<%s>" % (cls.__name__, trait.name, trait.__class__.__name__)
         lines.append(header)
+
+        if helptext is None:
+            helptext = trait.help
+        if helptext != '':
+            helptext = '\n'.join(wrap_paragraphs(helptext, 76))
+            lines.append(indent(helptext, 4))
+
         if inst is not None:
             lines.append(indent('Current: %r' % getattr(inst, trait.name), 4))
         else:
@@ -268,11 +275,6 @@ class Configurable(HasTraits):
             # include Enum choices
             lines.append(indent('Choices: %r' % (trait.values,)))
 
-        if helptext is None:
-            helptext = trait.help
-        if helptext != '':
-            helptext = '\n'.join(wrap_paragraphs(helptext, 76))
-            lines.append(indent(helptext, 4))
         return '\n'.join(lines)
 
     @classmethod
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -227,11 +227,14 @@ class Configurable(HasTraits):
         return '\n'.join(final_help)
 
     @classmethod
-    def class_get_trait_help(cls, trait, inst=None):
-        """Get the help string for a single trait.
+    def class_get_trait_help(cls, trait, inst=None, helptext=None):
+        """Get the helptext string for a single trait.
 
-        If `inst` is given, it's current trait values will be used in place of
-        the class default.
+        :param inst:
+            If given, it's current trait values will be used in place of
+            the class default.
+        :param helptext:
+            If not given, uses the `help` attribute of the current trait.
         """
         assert inst is None or isinstance(inst, cls)
         lines = []
@@ -265,10 +268,11 @@ class Configurable(HasTraits):
             # include Enum choices
             lines.append(indent('Choices: %r' % (trait.values,)))
 
-        help = trait.help
-        if help != '':
-            help = '\n'.join(wrap_paragraphs(help, 76))
-            lines.append(indent(help, 4))
+        if helptext is None:
+            helptext = trait.help
+        if helptext != '':
+            helptext = '\n'.join(wrap_paragraphs(helptext, 76))
+            lines.append(indent(helptext, 4))
         return '\n'.join(lines)
 
     @classmethod
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -66,7 +66,7 @@ class MyApp(Application):
 
     aliases = Dict({
                     ('fooi', 'i') : 'Foo.i',
-                    ('j', 'fooj') : 'Foo.j',
+                    ('j', 'fooj') : ('Foo.j', "`j` terse help msg"),
                     'name' : 'Foo.name',
                     'la': 'Foo.la',
                     'tb': 'Bar.tb',
@@ -194,7 +194,7 @@ class TestApplication(TestCase):
         class TestApp(Application):
             value = Unicode().tag(config=True)
             config_file_loaded = Bool().tag(config=True)
-            aliases = {'v': 'TestApp.value'}
+            aliases = {'v': ('TestApp.value', 'some help')}
         app = TestApp()
         with TemporaryDirectory() as td:
             config_file = pjoin(td, name)
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -324,6 +324,10 @@ that are single characters, in which case they can be specified with a single ``
 Flags and aliases are declared by specifying ``flags`` and ``aliases``
 attributes as dictionaries on subclasses of :class:`~traitlets.config.Application`.
 
+A key in both those dictionaries might be a string or tuple of strings.
+One-character strings are converted into "short" options (like ``-v``); longer strings
+are "long" options (like ``--verbose``).
+
 Aliases
 *******
 
@@ -337,6 +341,9 @@ to specify the whole class name:
     $ ipython --profile='myprofile'
     # are equivalent to
     $ ipython --BaseIPythonApplication.profile='myprofile'
+
+When specifying ``alias`` dictionary in code, the values might be the strings
+like ``'Class.trait'`` or two-tuples like ``('Class.trait', "Some help message")``.
 
 Flags
 *****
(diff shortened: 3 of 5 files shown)


ankostis:
Push another help-message modification:   put human explanation for opts above automated infos

## "Before" Help Message:
```
-d, --debug
    Equivalent to: [--Spec.log_level=0 --Cmd.print_config=True --Cmd.raise_config_file_errors=True --Application.log_level=0]
-v, --verbose
    Make various sub-commands increase their verbosity (not to be confused with --debug):
    Equivalent to: [--Cmd.verbose=True --Spec.verbose=True]
-f, --force
    Force various sub-commands to perform their duties without complaints.
    Equivalent to: [--Cmd.force=True --Spec.force=True]
--config-paths=<list-item-1>...
    Default: []
    Absolute/relative path(s) to read "static" configurable parameters from.
    Equivalent to: [--Cmd.config_paths]
--log-level=<Enum>
    Default: 30
    Equivalent to: [--Application.log_level]
--persist-path=<Unicode>
    Default: None
    Absolute/relative path to read/write persistent parameters on runtime, if
    `CO2DICE_PERSIST_PATH` envvar is not defined.
    Equivalent to: [--Cmd.persist_path]
```

## "After" Help Message:
```
-v, --verbose
    Make various sub-commands increase their verbosity (not to be confused with --debug):
    Equivalent to: [--Cmd.verbose=True --Spec.verbose=True]
-d, --debug
    Equivalent to: [--Cmd.raise_config_file_errors=True --Cmd.print_config=True --Spec.log_level=0 --Application.log_level=0]
-f, --force
    Force various sub-commands to perform their duties without complaints.
    Equivalent to: [--Cmd.force=True --Spec.force=True]
--persist-path=<Unicode>
    Absolute/relative path to read/write persistent parameters on runtime, if
    `CO2DICE_PERSIST_PATH` envvar is not defined.
    Default: None
    Equivalent to: [--Cmd.persist_path]
--log-level=<Enum>
    Default: 30
    Equivalent to: [--Application.log_level]
--config-paths=<list-item-1>...
    Absolute/relative path(s) to read "static" configurable parameters from.
    Default: []
    Equivalent to: [--Cmd.config_paths]
```

ankostis:
Regarding my last 6a93d8c for putting human option-infos on top, as it was before, the human explanation was "sanditched" between the `Default`  and the `Equivalent to` lines, and I had introduced that with #319, because it was the easiest thing to do.

I'm now thinking that it may be appropriate to keep the `Equivalent to` line to the top (or immediately below the human-text, because it is part of the explanation, and to have the `Default` and the other auto-infos at the end, which is customary for cmd-line tools.

(regardless, this last commit still has to apply)

ankostis:
Pushed fix commit 41480333b: 
since #289 has ensured that all parent hierarchy gets to be printed, 
we may avoid clutter by reprinting all traits on child classes.

ankostis:
Is there any problem with this PR?

traitlets/config/configurable.py
```diff
@@ -221,17 +221,20 @@ class defaults.
         base_classes = ','.join(p.__name__ for p in cls.__bases__)
         final_help.append(u'%s(%s) options' % (cls.__name__, base_classes))
         final_help.append(len(final_help[0])*u'-')
-        for k, v in sorted(cls.class_traits(config=True).items()):
+        for k, v in sorted(cls.class_own_traits(config=True).items()):
```
minrk: It seems like this should still be `.class_traits`, since it should include all configurables on the class, not just those it didn't inherit.

traitlets/config/tests/test_application.py
```diff
@@ -66,7 +66,7 @@ class MyApp(Application):
 
     aliases = Dict({
                     ('fooi', 'i') : 'Foo.i',
-                    ('j', 'fooj') : 'Foo.j',
+                    ('j', 'fooj') : ('Foo.j', "`j` terse help msg"),
```
minrk: The point of the help string on traitlets is to specify this help string. Under what circumstances would one set a help string that one didn't want to use?

ankostis: The point is `foo --help` and `foo --help-all` to print gradually more text for options.

Optimally, with `--help`, each option "should" occupy a single line, as most unix-utils do, to waste as little screen estate as possible (ok, we're cannot achieve 1-line bc there are default values, equivalent flags, etc). That is important for apps with a lot of aliases.

The `--help-all` is like printing the full man-page.

Of course it is still possible to use the original help-string.
Personally I use a little function that puts in the tuple the 1st line from the full help-msg.

More detailed reasoning on the PR OP.

traitlets/config/configurable.py
ankostis: No strong opinion on that. 
As reasoned in [this comment](/ipython/traitlets/pull/361#issuecomment-272519698), this is also to avoid clutter on `--help-all`: since #289 class-parents are printed, so the user may connect the dots and deduce which traits belong to subclasses.

But I will retract it, to keep the `--help-all` extra helpfull.

traitlets/config/tests/test_application.py
minrk: I have difficulty with that, since I can't think of any cases where I would want the two messages differ. I see the rationale in the OP, but it's not accurate, for instance:

> For flags, it is possible (actually forced) to provide shorter versions

There's a field for a help string only because it's not retrievable from the traits. There is no restriction on its length.

The only difference between `--help` output and `--help-all` output should be whether the entirety of classes are walked, vs only those items with aliases and flags. The `.help` key exists solely for the purpose of populating this output, so I don't understand why it wouldn't be used.

ankostis: The same reasoning you explained could have been applied also in the flags, like that:
Since `.help` keys exist for all options a flag masp to, then just concatenate all of them in one big help-message, and use this for the flag - no need to have a shorter version in a 2-tuple.
But hopefully the case is not like that.

I have a need for this because my app has ~15 subcommands with 3 levels of nesting. And since i did a trick to *merge* aliases and flags from parent cmds, so at time the output of ` root foo bar  --help`  is simply overwhelming, with tens of of them.  So I'm using shorter versions to conserve screen estate and help the users see what's important.

I repeat this PR provides an optional functionality, so by default the longer-help message is used.

ankostis:
@minrk  I retracted the 41480333b8401 commit which had the conversation about listing "own only" traits in help msg or not, so you cannot see it anymore.

traitlets/config/tests/test_application.py
ankostis: By the way, the number of config-params above is an order of magnitude bigger if you include `has_traits()` instead of `has_own_traits()`.  That is why my retracted 41480333b840 was trying to save space also for `--help-all`  case.

ankostis: For the future reader: describing here cases where it is totally necessary to override the help-msg on the alias: the trait is defined in some base-class, and each alias wants to explain its exact usage.

PR #362, opened 2017-01-20
Support callable subcommands

ankostis:
Subcommands might be classes, strings (name of classes) and now Callables with a single argument: the parent Application instance (i.e. to read configs from).

That way it is possible to create different subcommands based on the same Application-class just with different constructor arguments.

I could not fins any test-cases about subcommands, so this PR is not considered finished - submitted for discussion.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -468,18 +468,24 @@ class Application(SingletonConfigurable):
     @catch_config_error
     def initialize_subcommand(self, subc, argv=None):
         """Initialize a subcommand with argv."""
-        subapp,help = self.subcommands.get(subc)
+        subapp, _ = self.subcommands.get(subc)
 
         if isinstance(subapp, six.string_types):
             subapp = import_item(subapp)
-        elif isinstance(subapp, callable):
-            subapp = subapp()
 
-        # clear existing instances
-        self.__class__.clear_instance()
-        # instantiate
-        self.subapp = subapp.instance(parent=self)
-        # and initialize subapp
+        ## Cannot issubclass() on a non-type (SOhttp://stackoverflow.com/questions/8692430)
+        if isinstance(subapp, type) and issubclass(subapp, Application):
+            # Clear existing instances before...
+            self.__class__.clear_instance()
+            # instantiating subapp...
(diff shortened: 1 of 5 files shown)


traitlets/config/application.py
```diff
@@ -246,11 +246,13 @@ def _flags_changed(self, change):
             assert isinstance(value[1], six.string_types), "Bad flag: %r:%s" % (key, value)
 
 
-    # subcommands for launching other applications
-    # if this is not empty, this will be a parent Application
-    # this must be a dict of two-tuples,
-    # the first element being the application class/import string
-    # and the second being the help string for the subcommand
+    #: subcommands for launching other applications
```
minrk: looks like these colons were added by accident

traitlets/config/application.py
```diff
@@ -246,11 +246,13 @@ def _flags_changed(self, change):
...
+    #: if this is not empty, this will be a parent Application
+    #: this must be a dict of two-tuples,
+    #: the first element is the application class, or and import string
+    #: or a callable with a single argument for the parent returnng
```
minrk: returnng

traitlets/config/application.py
```diff
@@ -464,13 +466,16 @@ def initialize_subcommand(self, subc, argv=None):
         """Initialize a subcommand with argv."""
         subapp,help = self.subcommands.get(subc)
 
-        if isinstance(subapp, six.string_types):
-            subapp = import_item(subapp)
+        if callable(subapp):
+            self.subapp = subapp(self)
```
minrk: classes are callable, so this check will catch the common case of subapp being an Application and handle it incorrectly.

ankostis: Indeed!

Would this be ok?

    if callable(subapp) and not isinstance(subapp, Application):
        ...

minrk: I would think so, yes.

ankostis: Editing my self, more like `issubclass(subapp, Application)`...and also avoid invoking `ConfigSingleton.instance()` when factory method supplied.

minrk: Thanks!

PR #365, opened 2017-01-26
FIX: KVConfigLoader could not handle lone dashes('-') as extra arguments

ankostis:
`KeyValueConfigLoader.load_config()` was considering lone dash(`-`) args as invalids
(these are frequently used to denote STDIN).

+ Fixed and added a TCs on App & Loader.
+ Minor refactorings 7 pep8 on the method fixed.

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -699,13 +699,9 @@ class KeyValueConfigLoader(CommandLineConfigLoader):
             item = raw.lstrip('-')
 
             if not item:
-                # I.e. it was a lone '-' used for denoting STDIN, or '---'.
-                self.extra_args.append('-')
-                continue
-
-            if not item:
-                # I.e. it was a lone '-' used for denoting STDIN, or '---'.
-                self.extra_args.append('-')
+                ## It was either a lone '-' (i.e. denoting STDIN),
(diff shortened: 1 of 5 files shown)


ankostis:
Added a rather unrelated commit for site-docs: 
added more project coordinates (links to other site) that is annoying not to have on every landing page of this project.
Actually If given the permission I would port a9680b5 also on the top of `README.md` so that the links are visible in GitHub landing page.

traitlets/config/loader.py
```diff
@@ -687,19 +687,29 @@ def load_config(self, argv=None, aliases=None, flags=None):
...
-                break
+            if not item:
+                # I.e. it was a lone '-' used for denoting STDIN, or '---'.
+                self.extra_args.append('-')
+                continue
+
+            if not item:
+                # I.e. it was a lone '-' used for denoting STDIN, or '---'.
+                self.extra_args.append('-')
+                continue
```
minrk: duplicated block?

ankostis: Yes, duplicate code. 
Also previously was replacing user-arg always with '-' - now adds user-arg verbatim, even if it is 3 or more dashes concatenated.

PR #372, opened 2017-02-01
add rollback function to change dicts

rmorshea:
Adds a `'rollback'` key to change dicts that is called to reverse changes in case of a `TraitError`.

This enables rollbacks for different kinds of changes (e.g. element changes from eventful traits))

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -28,7 +28,7 @@ from traitlets import (
 import six
 
 def change_dict(*ordered_values):
-    change_names = ('name', 'old', 'new', 'owner', 'type')
+    change_names = ('name', 'old', 'new', 'owner', 'type', 'rollback')
     return dict(zip(change_names, ordered_values))
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


minrk:
What are some cases where this has come up? What happens for future trait notifiers if rollback is called by the first listener?

SylvainCorlay:
:-1: on this. The original proposed method was that the rollback function would be a member of the change dictionary.

rmorshea:
@SylvainCorlay it is a part of the change dictionary. I just happen to have defined it as a static method on `TraitType` - I can change that though.

rmorshea:
@minrk this idea originates [here](https://github.com/SylvainCorlay/traitlets/blob/5ef4db537e535a2c5c711c657dd65e96a219cd0d/traitlets/traitlets.py#L1118-L1120) in @SylvainCorlay's implementation of eventful trait's. However, I can see your point. Maybe we could pass it to `notify_change(change, rollback=None)`?

rmorshea:
@minrk I think rollback functions only make sense with respect to #251.

Each change type ought to be internally responsible for how it is reversed.

Additionally...

We could allow us to have generic event compression logic:

```python
   if hasattr(event, "compress") and type(event) is type(other):
      event.compress(other)
```

 Right now we have a hard coded `type="change"` compression function in `hold_trait_notifications`.

@SylvainCorlay, thoughts?

Issue #287, opened 2016-08-24
Default values are not cross-validated with `@validation`


ankostis:
While the defaults values are validated against the type of their trait, they are not cross-validated against any `@validate` rules.

This may lead to conflicting behavior in `HasTraits` subclasses when they contain invalid defaults, as exemplified in #263, and copied here for clarity:

``` python
class A1(HasTraits):
   virtue = Unicode('BAD')

   @validate('virtue')
   def _no_evil(self, proposal):
       if proposal['value'] == 'BAD':
           raise TraitError()
       return proposal['value']

class A2(HasTraits):
   virtue = Unicode('GOOD')

   @validate('virtue')
   def _no_evil(self, proposal):
       if proposal['value'] == 'BAD':
           raise TraitError()

>>> A1('virtue='BAD')  ## Pass
>>> A2('virtue='BAD')  ## Screams!
traitlets.traitlets.TraitError
```

SylvainCorlay:
Thanks for reposting here. There were some unrelated discussions on the other thread.

SylvainCorlay:
I just wanted to point out that it is not a trivial problem and not doing it was a deliberate choice. Basically, it breaks the possibility of a delayed initialization for a value if it is accessed at the cross-validation stage.

rmorshea:
I looked into this recently and the problem here is quite subtle:

1. Setting via `A1(virtue='BAD')` invokes a `hold_trait_notifications` context.
2. Cross validation never occurs inside this context `=>` validation of default values cannot occur here.
	+ In other words dynamic defaults created in the context **never** get validated. So even if A1 had a dynamic default of "BAD" via `@default` (which is normally validated) the problem noted above would still persist.
3. The fact that invalid default values are allowed to generate inside `hold_trait_notifications` (dynamic or otherwise) means that they will be used to determine whether change notifications should occur in `TraitType.set`.
	+ This implies that the failure of `A1(virtue=invalid)` to raise is actually an edge case where `invalid` happens to equal `A1.virtue.default_value` and a change notification is never sent to `HasTraits.notify_change`. That failure to notify means that `hold_trait_notifications` never retraces it in the cache of changes.
	+ In short, if we chose `A1(virtue="EVIL")` an error would be raised.

If you take all this together, and we say that validating default isn't something that we will guarantee, then the bug here is actually that `hold_trait_notifications` operates under the assumption that all prior trait values have been validated.

cc: @SylvainCorlay

rmorshea:
[...]

PR #373, opened 2017-02-01
Fix Change Dict Bug - Improper Old Values

rmorshea:
# The Problem

When it comes to default values, the `'old'` key in change dicts is often improperly assigned.

For example:

```python
class Foo(HasTraits):
    i = Int()
    @default('i')
    def _i_default(self):
        return 1
    @observe('i')
    def _i_changed(self, change):
        raise ValueError(str(change))

foo = Foo()
foo.i = 1
```

Surprisingly, the observer gets called here because `'old'` is `0`!

This is because `'old'` gets assigned via `i.default_value` rather than by a default generator.

# The Fix

Just call `getattr` to try and make a default value. If that fails `'old'` is `Undefined`.

side note: *the default must be generated **before** validating the new value. This is because any state changes induced by default generation should be overridden by setting/validating the new value.*

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -499,6 +499,25 @@ class TraitType(BaseDescriptor):
             # Undefined will raise in TraitType.get
             return self.default_value
 
+    def obj_default(self, obj):
+        """Return this trait's default value on the given object.
+
+        Pulls from the object's ``_trait_default_generators`` using
+        ``self.name`` to generate a default value and validate it.
+        If that value is ``Undefined`` then an error is raised.
+        """
+        try:
+            dgen = type(obj)._trait_default_generators[self.name]
+        except KeyError:
+            raise TraitError("Impropper default generator setup "
+                "for '%s.%s'" % (type(obj).__name__, self.name))
+        default = dgen(obj)
+        if default is Undefined:
+            raise TraitError("No default value found for "
+                    "the '%s' trait named '%s' of %r" % (
+                    type(self).__name__, self.name, obj))
+        return self._validate(obj, default)
+
     def get_default_value(self):
         """DEPRECATED: Retrieve the static default value for this trait.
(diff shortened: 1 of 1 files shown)


minrk:
I'm pretty sure it was a deliberate, important change to *prevent* the default value generator from being called on first assignment, which caused problems in IPython when there are more complex default value generators. I think using `Undefined` or similar for the old value is better than generating a never-used value for `old`.

rmorshea:
@minrk since defaults now only get generated at the time they are requested, `Undefined` makes sense. Either way though drawing `'old'` from `TraitType.default_value` doesn't seem right.

minrk:
Most of the time, it is the right value. It's only when a dynamic default is defined that it gets skipped.

minrk:
Using Undefined when there's a dynamic default makes sense, though.

rmorshea:
Because of #349, I'm not so sure that it's valuable to distinguish between a default declared here `TraitType(default_value=<value>)` and one defined via `@default` anymore - now, a trait's default is **always** decided at the time it is first requested, and never during `instance_init` as was previously the case for "static" defaults.

Using `TraitType.default_value` means that it's possible for `'old'` to be inconsistent with expectations whenever `@default` is used. It seems that it would be better to give no information about the old value, than to give information that's false.

rmorshea:
@minrk looking through old issues I found this [bug](https://github.com/ipython/traitlets/issues/287#issuecomment-255456913) wherin using `TraitType.default_value` can result in a failure to validate new values in `hold_trait_notifications`.

minrk:
That makes sense.

rmorshea:
@minrk, `old=Undefined` breaks multiple tests...

Is it possible that the new on-request default generation, might bypass the problems you alluded to?

*the old way being default state initialization during `HasTraits.__init__`*

PR #376, opened 2017-02-17
Bless config metadata in trait constructors

minrk:
After updating many applications to the 4.1 API, requiring `.tag(config=True)` does not seem to be an improvement, plus the definition of the config metadata is defined in this package.

Blessing the longstanding behavior makes upgrading a lot less tedious.

### DIFF
--- a/traitlets/config/tests/test_configurable.py
+++ b/traitlets/config/tests/test_configurable.py
@@ -439,7 +439,7 @@ class TestConfigContainers(TestCase):
 
         # reset deprecation limiter
         _deprecations_shown.clear()
-        with expected_warnings(['metadata should be set using the \.tag\(\) method', "use @default decorator instead\\."]):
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
I think that the separation of 

 - keyword arguments (like `allow_none`) 
 - from metadata  (`sync=True`, `from_json`, `to_json`)

made things a lot clearer.

minrk:
In practice in applications, I don't think it has made it clearer for `config`, specifically. Consumers of config ~never need to know that `config=True` is implemented via metadata vs. attributes. It's just an opaque signature change.

I think it *is* clearer for metadata in general and widgets in particular, where it's a different package defining its own extra things to store on the trait.

SylvainCorlay:
:+1:

minrk:
Thanks!

PR #377, opened 2017-02-17
undeprecate `_trait_default`

minrk:
After upgrading several applications to the 4.1 API, requiring the default decorator results in a bunch of redundant code.

Since the signature and behavior are unchanged, just remove the deprecation message on default-value generators.

I won't propose undeprecating change/validate methods, since their signatures change with the decorators.

### DIFF
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -118,11 +118,12 @@ by notification type.
         def handler_all(self, change):
             pass
 
-Deprecation of magic method for dynamic defaults generation
------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
This makes sense for defaults.

Although I like using the `@default` decorator in a class where I have calls to `@validate` and `observe` which make the whole thing more uniform...

minrk:
@SylvainCorlay agree 100%. I like that it's available as a more consistent "the new way". Just reducing the deprecation burden.

PR #378, opened 2017-02-17
Changelog for 4.3.2

minrk:
Tiny release, removing a couple of deprecations to make upgrading existing packages more pleasant.

cc @JamiesHQ


SylvainCorlay:
@minrk I will put some cycles to investigate the regressions since 4.3 in the beginning of next week.

minrk:
@SylvainCorlay excellent, thanks. I don't think 5.0 is very close to ready.

PR #381, opened 2017-03-17
feat(cfg): write also enum choices when generating config.py files

ankostis:
When generating the `config.py` file, explain the available enum-choices in the comment-lines, like help-message does.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -310,10 +310,14 @@ class Configurable(HasTraits):
 
         for name, trait in sorted(cls.class_own_traits(config=True).items()):
             lines.append(c(trait.help))
+
             if 'Enum' in type(trait).__name__:
                 # include Enum choices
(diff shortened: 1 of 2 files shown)


ankostis:
I would like to add one commit more that appends the "Default=<value>" line to the comments above each config-parameter, and totally replicate the help-message.

Currently the default-value is printed in the `Class.trait =def_value` line.
But if you un-comment and edit that line, you don't know what was the default.

Should I do that?

minrk:
sure, that seems ok.

ankostis:
With latest 2 commits now default-value are also printed on comments, and this PR is finished on my part.

An additional change(169e8492) is that the enum-choices are printed *before* the default-value.
Enhanced TCs to check this ordering.

ankostis:
My apologies, just seen that there is a minor glitch in the generated configuration file (described in the review).

PR #383, opened 2017-03-22
notify on default creation

rmorshea:
# Summary

Notify upon the creation of a default value. The current scheme for this event report is:

```python
{
    'name': <the trait's name>,
    'owner': <it's HasTraits object>,
    'value': <default value>
    'type': 'default',
}
```

# A Possible Use Case

An `@access` decorator that acts like `@property` but has the added benefits of `traitlets` validation.

```python
class AccessHandler(DefaultHandler):
    """An event handler that allows a trait to act like a getter"""

    def instance_init(self, obj):
        obj.observe(
            self._reset,
            self.trait_names,
            "default",
            self.metadata
        )

    @staticmethod
    def _reset(change):
        # this reset will cause the default mechanism
        # to reactivate whenever the trait is accessed
        del change.owner._trait_values[change.name]
```

## TODO

+ needs tests

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -523,6 +523,12 @@ class TraitType(BaseDescriptor):
                     type(self).__name__, self.name, obj))
             value = self._validate(obj, default)
             obj._trait_values[self.name] = value
+            obj.notify_change(Bunch(
+                name=self.name,
+                value=value,
+                owner=obj,
+                type='default',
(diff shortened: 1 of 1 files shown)


minrk:
This has broken widgets because it calls notify_change with events that are not change events. We either need to ensure that these events are supersets of change events, or trigger a different call that `notify_change`

PR #384, opened 2017-03-22
add `trait_values` method

rmorshea:
Returns a ``dict`` of trait values via `getattr`. This is useful for parsing a `HasTraits` object into json.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1412,6 +1412,28 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
         """Returns True if the object has a trait with the specified name."""
         return isinstance(getattr(self.__class__, name, None), TraitType)
 
+    def trait_values(self, **metadata):
+        """A ``dict`` of trait names and their values.
+
+        The metadata kwargs allow functions to be passed in which
+        filter traits based on metadata values.  The functions should
+        take a single value as an argument and return a boolean.  If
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -1412,6 +1412,34 @@ def has_trait(self, name):
...
+        Trait values are retrieved via ``getattr``, any exceptions raised
+        by traits or the operations they may trigger will result in the
+        absence of a trait value in the result ``dict``.
+        """
+        values = {}
+        for name in self.trait_names(**metadata):
+            try:
+                values[name] = getattr(self, name)
+            except:
+                pass
```
minrk: Should exceptions really be ignored? What's a case where an error should actually be suppressed?

rmorshea: I'm wasn't entirely sure of how to handle this, because traits which have no assigned value and which have no default value, will raise when requested via `getattr`. However, the intuitive expectation for such traits is that they simply be excluded from the returned `dict`. I would have just returned a copy of `HasTraits._trait_values` however, this would mean that traits which have valid default values would not be represented.

Perhaps `values[name] = error` if an exception is encountered?

minrk: I would let it raise. If getting any trait values would fail, I would expect getting all trait values to also fail in the same way.

rmorshea:
@minrk, `getattr` can now raise

PR #391, opened 2017-04-07
Suppress Redundant Configurable Validation

### LINKED ISSUE #390: Configurable validates too often
Consider this:

```python
from traitlets.config import Configurable
class A(Configurable):
    x = Int(1)
    
    @validate('x')
    def f(self, proposal):
        print('validate x')
        return proposal.value
A(x=5)
``` 
prints:
```
validate x
validate x
```

while this:

```python
class A(HasTraits):
    x = Int(1)

    @validate('x')
    def f(self, proposal):
        print('validate x')
        return proposal.value
A(x=5)
```
only validates x once:
```
validate x
```
For some reason, Configurable is validating x twice, but HasTraits is only validating once.

rmorshea:
Closes #390

## Summary

Configurables currently make a redundant series of `setattr` calls on themselves to override config values with those that have been explicitly hard-coded in the constructor. In the past this was required due to the timing of descriptor initialization via `instance_init`, which occurred in `HasTraits.__init__`. Now though, descriptor initialization occurs in `HasTraits.__new__` via `setup_instance`. Thus we can remove the redundant `setattr` in `Configurable.__init__` and simply use `super().__init__(**kwargs)` instead.

## Todo

1. I'm encountering a test failure complaining that a warning for a mistyped section did not occur. However, pytest shows that `stderr` received a call with the exact warning required by the test. This is specific to Python 3 since it requires `TestCase.assertLogs`.
2. Add a test case.

cc: @jasongrout

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -77,9 +77,6 @@ class Configurable(HasTraits):
 
         config = kwargs.pop('config', None)
 
-        # load kwarg traits, other than config
-        super(Configurable, self).__init__(**kwargs)
-
         # load config
         if config is not None:
             # We used to deepcopy, but for now we are trying to just save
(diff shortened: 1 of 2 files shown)


minrk:
While you're at it, can you make sure there's a test case for what this is supposed to resolve?

```python
class C(Configurable):
    x = Unicode(config=True)

cfg = Config()
cfg.C.x = 'from config'

c = C(x='from init')
assert c.x == 'from init'
```

rmorshea:
@minrk sure. Any thoughts on the test failure I'm getting here? I don't know what's causing it.

rmorshea:
Creative fix! +1

PR #392, opened 2017-04-13
style(cfg): apply review items in #385 for gen-config

ankostis:
Minor changes in the formatting of the generated sample configuration-file,
as reviewed by @ankostis on #385.

The differences in the generated config-file are shown below for this code:

```python
import traitlets.config as trtc
import traitlets as trt

class C(trtc.Configurable):
    a = trt.Enum(['aa', 'BB'], help='1st help line.\n\nOther lines.').tag(config=True)

class A(trtc.Application, C):
    pass

print(A().generate_config_file())
```

## Wihtout this PR:
```python
#------------------------------------------------------------------------------
# A(Application, C) configuration
#------------------------------------------------------------------------------
## This is an application.

## 1st help line.
#  See also C.a
# c.A.a = traitlets.Undefined
```

## WITH this PR:
```python
THIS PR:
##------------------------------------------------------------------------------
## A(Application, C) configuration
##------------------------------------------------------------------------------
## This is an application.

## 1st help line.
#  See also: C.a
#c.A.a = traitlets.Undefined
```

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -365,7 +365,7 @@ class Configurable(HasTraits):
                 # Truncate help to first line + "See also Original.trait"
                 if trait.help:
                     lines.append(c(trait.help.split('\n', 1)[0]))
-                lines.append('#  See also %s.%s' % (defining_class.__name__, name))
+                lines.append('#  See also: %s.%s' % (defining_class.__name__, name))
 
             lines.append('# c.%s.%s = %s' % (cls.__name__, name, default_repr))
             lines.append('')
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -438,13 +438,13 @@ class TestApplication(TestCase):
         self.assertIn('# c.NoTraits.i', conf_txt)
         self.assertIn('# c.NoTraits.j', conf_txt)
         self.assertIn('# c.NoTraits.n', conf_txt)
-        self.assertIn('#  See also Foo.j', conf_txt)
-        self.assertIn('#  See also Bar.b', conf_txt)
+        self.assertIn('#  See also: Foo.j', conf_txt)
+        self.assertIn('#  See also: Bar.b', conf_txt)
         self.assertEqual(conf_txt.count('Details about i.'), 1)
 
         # inherited traits, parent not in class list:
         self.assertIn("# c.NoTraits.from_hidden", conf_txt)
-        self.assertNotIn('#  See also NotInConfig.', conf_txt)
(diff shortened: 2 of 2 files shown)


traitlets/config/configurable.py
```diff
@@ -326,13 +326,13 @@ def c(s):
...
         # section header
-        breaker = '#' + '-' * 78
+        breaker = '##' + '-' * 78
         parent_classes = ', '.join(
             p.__name__ for p in cls.__bases__
             if issubclass(p, Configurable)
         )
 
-        s = "# %s(%s) configuration" % (cls.__name__, parent_classes)
+        s = "## %s(%s) configuration" % (cls.__name__, parent_classes)
```
minrk: Let's keep single `#`, not double here and above.

traitlets/config/configurable.py
```diff
@@ -365,9 +365,9 @@ def c(s):
                 # Truncate help to first line + "See also Original.trait"
                 if trait.help:
                     lines.append(c(trait.help.split('\n', 1)[0]))
-                lines.append('#  See also %s.%s' % (defining_class.__name__, name))
+                lines.append('#  See also: %s.%s' % (defining_class.__name__, name))
 
-            lines.append('# c.%s.%s = %s' % (cls.__name__, name, default_repr))
+            lines.append('#c.%s.%s = %s' % (cls.__name__, name, default_repr))
```
minrk: The space between `#` and `c` are helpful, let's keep them here.

traitlets/config/configurable.py
ankostis: Ok

traitlets/config/configurable.py
ankostis: Is it possible to keep it?

Otherwise, then the last code line does not differ from comments immediately above, and 
it needs 2 key strokes to enable the code.

Note, it was like these before #385 (actually i believe it was me that had passed this change in the first place, and had been accepted).

minrk: Let's keep the space. It's generally [recommended](https://www.python.org/dev/peps/pep-0008/#block-comments) to have a space between the comment symbol and the comment itself. Plus, editors that have a 'comment/uncomment' action generally deal with this in a single keystroke as well.

ankostis:
OK, kept only `See also:` change and rebased on latest master.

Regarding the code-lines that now have 1 space between the `#` char, 
it is important that we preserve *some* difference between the comment-lines.
Currently, the comment-lines start with 2 spaces after `#`.

So the following expression can collect all code-lines (and possibly some user-edited comments, but that's ok):

    sed -n '/^# \w/p'  config.py

PR #393, opened 2017-04-13
Link tranform

francisco-dlp:
Add ``transform`` argument to ``link`` that takes a tuple with the forward and backward transform.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1906,9 +1906,35 @@ class TestLink(TestCase):
         a.value = 5
         self.assertEqual(b.value, 10)
         # Change one the value of the target and check that it modifies the
-        # source. 
+        # source.
         b.value = 6
         self.assertEqual(a.value, 3)
+
+    def test_link_broken_at_source(self):
+        class MyClass(HasTraits):
+            i = Int()
+            j = Int()
+
+            @observe("j")
+            def another_update(self, change):
+                self.i = change.new * 2
+
+        mc = MyClass()
+        l = link((mc, "i"), (mc, "j"))
+        self.assertRaises(TraitError, setattr, mc, 'i', 2)
+
+    def test_link_broken_at_target(self):
+        class MyClass(HasTraits):
+            i =Int()
+            j = Int()
+
+            @observe("i")
+            def another_update(self, change):
+                self.j = change.new * 2
+
+        mc = MyClass()
+        l = link((mc, "i"), (mc, "j"))
+        self.assertRaises(TraitError, setattr, mc, 'j', 2)
 
 class TestDirectionalLink(TestCase):
     def test_connect_same(self):
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1882,6 +1882,30 @@ class TestLink(TestCase):
         self.assertEqual(''.join(callback_count), 'ab')
         del callback_count[:]
 
+    def test_tranform(self):
+        """Test transform link."""
+
+        # Create two simple classes with Int traitlets.
+        class A(HasTraits):
+            value = Int()
+        a = A(value=9)
+        b = A(value=8)
+
+        # Conenct the two classes.
+        c = link((a, 'value'), (b, 'value'),
+                 transform=(lambda x: 2 * x, lambda x: int(x / 2.)))
+
+        # Make sure the values are correct at the point of linking.
+        self.assertEqual(b.value, 2 * a.value)
+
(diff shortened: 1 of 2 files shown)


minrk:
cc @SylvainCorlay for review

SylvainCorlay:
Thanks.

To provide some background, one reason why we only added it to the directional case originally is that one needs to ensure that the two tranforms are inverses of each other to avoid infinite bouncing behavior.

francisco-dlp:
Thanks for the feedback. Why would two transforms that are not inverses of each other cause bouncing behaviour?

I think that this is an useful feature. For example, say that I define a span selector and I want to fix its with without changing the class definition. With ``transform`` it is as simple as:

```python

import traitlets

class SpanSelector(traitlets.HasTraits):
    x1 = traitlets.Float()
    x2 = traitlets.Float()

ss = SpanSelector()
fix_width = traitlets.link((ss, "x1"), (ss, "x2"), (lambda x: x + 10, lambda x: x - 10))

```

Of course things may not work as expected if the transforms are not inverses, but I think that it is fair to assume that the users know what they're doing.

I have merged the current master into the branch to solve the merge conflicts.

rmorshea:
@francisco-dlp the bouncing behavior would result from the fact that the two traits would always send each other different values from what they already have. Consider the following case of a 2D linked trait relationship:

```
a(b) = b**2
b(a) = a**2
```

Here, `a` and `b` are not related by inverse functions. Thus, if we were to set `a` with the value 2...

+ the change in `a` would cause `b` to update to 4
+ the change in `b` would cause `a` to update to 16
+ the change in `a` would cause `b` to update to 256
+ on and on forever...

The only reason that we don't get infinite recursion in 2D linked traits is when changes in a proverbial trait `b` do not cause a change in trait `a` - the lack of change in `a` is what kills the update notifications.

maartenbreddels:
I think this is a useful feature, and it would be nice to have this in ipywidgets as well. Can we check in `_update_source` and `_update_target` if the transformations are actually the inverse? And raise an Exception is they differ.

francisco-dlp:
I think that checking if the transformations are the inverse of each other is not trivial e.g. they may be the inverse of each other only in a certain domain.

rmorshea:
@francisco-dlp because we check whether `link.updating` I don't think we have to worry about bouncing. However, it would be possible to break the consistency of the relationship:

```python
import math

class MyClass(HasTraits):
    i = Int()
    j = Int()
    
    @observe("j")
    def another_update(self, change):
        self.i = change.new * 2

mc = MyClass()

trans = (lambda x: x**2, lambda x: math.sqrt(x))
l = link((mc, "i"), (mc, "j"), trans)

mc.i = 2
```

```python
>>> mc.i
8
>>> mc.j
4
```

Inconsistent states will result when a consequence of a link's update changes either the source or the target trait. To resolve this issue we could raise if, after the update context is closed, the transform's relationship has been broken or not.

SylvainCorlay:
> I think that checking if the transformations are the inverse of each other is not trivial e.g. they may be the inverse of each other only in a certain domain.

That is why we did not implement bidirectional transform originally!

francisco-dlp:
@rmorshea, I think that the issue that you mention is unrelated to this pull request e.g.:

```python

class MyClass(HasTraits):
    i = Int()
    j = Int()
    
    @observe("j")
    def another_update(self, change):
        self.i = change.new * 2

mc = MyClass()
l = link((mc, "i"), (mc, "j"))

mc.i = 2
```

```python

>>> mc.i
4
>>> mc.j
2
```

@SylvainCorlay, there is of course a danger of things not working as expected if the arguments passed to a function are wrong or inconsistent, but this is true in general. Hence, I don't see why this should be an issue for adding an useful feature in this case.

maartenbreddels:
If it doesn't cause bouncing around, and we just say it's the responsibly of the user/programmer to make sure the transformations are fine I don't see a reason not to implement it.

rmorshea:
@francisco-dlp, true - inconsistency was possible before - however if we decided that we should raise when it's encountered, that would have implications for how we implement transforms.

francisco-dlp:
@rmorshea, I have modified the code to raise a ``TraitError`` when the link doesn't work as intended. Thanks for pointing this out, as I think that it'll help prevent some hard to debug issues.

francisco-dlp:
If anybody would like to use this feature while waiting for this PR to get merged (or not) you could try [link_traits](https://pypi.python.org/pypi/link-traits). The purpose of the package is to implement the dlink and link functions of traitlets for enthought traits. It can be used to link traits to traits, traits to traitlets and (redundantly) traitlets to traitlets.

PR #402, opened 2017-05-08
Better Errors for Nested Traits

rmorshea:
# Summary

Right now nested traits have pretty horrid error messages.

The following code produces three different tracebacks with three bad error messages:

```python
class MyClass(HasTraits):
    
    l = List(List(Int()))

mc = MyClass()
mc.l = [["value"]]
```

Error messages (tracebacks excluded):

```
The 'None' trait of a MyClass instance must be an int, but a value of 'str' <class 'str'> was specified.``
```

```
Element of the 'None' trait of a MyClass instance must be an int, but a value of 'str' <class 'str'> was specified.
```

```
Element of the 'l' trait of a MyClass instance must be a list, but a value of ['str'] <class 'list'> was specified.
```

The First two messages, while accurate, don't have a trait name associated with them which makes them confusing to read. The last one has a trait name but is inaccurate since it complains of a list being provided to a `List` trait!

# Solution

A generic [`TraitType.error(obj, value, error=None)`](https://github.com/rmorshea/traitlets/blob/nested_errors/traitlets/traitlets.py#L618-L641) method that can be used to easily handle nested errors. To use it, one only needs to catch the nested error in a parent trait and pass it to `error`.

With the proposed changes, the code above would create one concise traceback and message:

```
TraitError: The 'l' trait of a MyClass instance contains an Int of a List of a List which expected an int, not the str 'value'
```

# Notes

There is an added utility file `descriptions.py` which has a new [`describe`](https://github.com/rmorshea/traitlets/blob/nested_errors/traitlets/utils/descriptions.py#L7-L109) function in addition to others that were previously present in `traitlets.py` like `add_article` or `repr_type`. The `describe` function provides a common way to represent a value as a readable string (see docstring for use details).

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -583,29 +583,60 @@ class TraitType(BaseDescriptor):
     def info(self):
         return self.info_text
 
-    def error(self, obj, value, error=None):
+    def error(self, obj, value, error=None, info=None):
+        """Raise a TraitError
+
+        Parameters
+        ----------
+        obj: HasTraits or None
+            The instance which owns the trait. If not
+            object is given, then an object agnostic
+            error will be raised.
+        value: any
+            The value that caused the error.
+        error: Exception (default: None)
+            An error that was raised by a child trait.
+            The arguments of this exception should be
+            of the form ``(value, info, *traits)``.
+            Where the ``value`` and ``info`` are the
+            problem value, and string describing the
+            expected value. The ``traits`` are a series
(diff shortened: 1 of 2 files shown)


rmorshea:
@minrk, should probably get the 5.0 milestone.

rmorshea:
@minrk, this now passes after https://github.com/ipython/traitlets/pull/406.

minrk:
Thanks, @rmorshea!

PR #403, opened 2017-05-09
New TypeCast

rmorshea:
Refactoring #317

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1585,8 +1585,8 @@ class TestTupleTrait(TraitTestBase):
         return value
 
     def test_invalid_args(self):
-        self.assertRaises(TypeError, Tuple, 5)
-        self.assertRaises(TypeError, Tuple, default_value='hello')
+        self.assertRaises(TraitError, Tuple, 5)
+        self.assertRaises(TraitError, Tuple, default_value='hello')
         t = Tuple(Int(), CBytes(), default_value=(1,5))
 
 class LooseTupleTrait(HasTraits):
@@ -1607,8 +1607,8 @@ class TestLooseTupleTrait(TraitTestBase):
         return value
 
     def test_invalid_args(self):
-        self.assertRaises(TypeError, Tuple, 5)
-        self.assertRaises(TypeError, Tuple, default_value='hello')
+        self.assertRaises(TraitError, Tuple, 5)
+        self.assertRaises(TraitError, Tuple, default_value='hello')
         t = Tuple(Int(), CBytes(), default_value=(1,5))
 
 
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1728,15 +1728,16 @@ class TypeCast(ClassBasedTraitType):
             self._cast_types = (self._cast_types,)
         return self
 
-    def __init__(self, default_value=Undefined, *args, **kwargs):
-        if default_value not in (Undefined, None):
+    def __init__(self, *args, **kwargs):
+        self.klass = kwargs.pop("klass", self.klass)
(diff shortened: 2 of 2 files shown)


rmorshea:
@minrk if you have any thoughts on why the "hard part" (a0296cf) is failing let me know.

Something subtle is wrong, so I'm tempted to just submit the "easy part" (6b66f1c) on it's own.

rmorshea:
Ugh... I just forgot to pass the default value through `super().__init__` :expressionless: - not nearly as subtle an error as I thought. Still have to make `Dict` inherit from `TypeCast` though.

rmorshea:
@minrk passing now.

rmorshea:
@minrk questions, comments, or concerns about this? I mostly need to add documentation.

minrk:
Sorry I haven't gotten to it, yet. This is a big rewrite of a lot of traitlets, which always makes me nervous. I just haven't had the time to go through it thoroughly and test it with every downstream IPython and Jupyter project, etc.. I'll try to get to this soon.

rmorshea:
@minrk, the traits of most concern would be those inheriting from 'Container' since their implementation is relatively complicated.

rmorshea:
@minrk, would it be better for me to break this into multiple PRs?

1. Introduce `TypeCast` without refactoring
2. Refactor trivial cases (e.g. `Int`, `Unicode`, etc.)
3. Refactor hard cases (e.g. `Container`, `Mapping`, etc.)

rmorshea:
@minrk 6.0? or 5.0?

traitlets/traitlets.py
```diff
@@ -2269,95 +2291,110 @@ def validate(self, obj, value):
...
+    def __init__(self, *args, **kwargs):
+        super(Mutable, self).__init__(*args, **kwargs)
+        self._default_progenitor = self.default_value
+        self.default_value = Undefined
 
-        trait : TraitType [ optional ]
-            the type for restricting the contents of the Container.  If unspecified,
-            types are not checked.
+    def make_dynamic_default(self):
+        return copy.copy(self._default_progenitor)
```
minrk: Shouldn't this be `copy.copy(self.default_value)`? Why is the default_value unused and a separate private attribute used for the same purpose?

traitlets/traitlets.py
```diff
@@ -2269,95 +2291,110 @@ def validate(self, obj, value):
...
-class Container(Instance):
-    """An instance of a container (list, set, etc.)
-
-    To be subclassed by overriding klass.
-    """
-    klass = None
-    _cast_types = ()
-    _valid_defaults = SequenceTypes
-    _trait = None
+class Mutable(TypeCast):
```
minrk: What is Mutable? If it's only here to be a base class, make it private: `_Mutable`. Also make sure that every one of these has a docstring.

traitlets/traitlets.py
```diff
@@ -2269,95 +2291,110 @@ def validate(self, obj, value):
...
             except TraitError as error:
                 self.error(obj, v, error)
             else:
-                validated.append(v)
-        return self.klass(validated)
+                new.append(v)
+        return new
+
+
+class BaseSequence(Container):
```
minrk: What is BaseSequence? It needs a docstring, and should be private if it's only here to be a helper for Sequence below.

traitlets/tests/test_traitlets.py
```diff
@@ -1585,8 +1585,8 @@ def coerce(self, value):
         return value
 
     def test_invalid_args(self):
-        self.assertRaises(TypeError, Tuple, 5)
```
minrk: Calling a constructor with the wrong arguments does normally raise TypeError. Why has this changed?

rmorshea:
Close in favor of https://github.com/ipython/traitlets/pull/477

PR #406, opened 2017-05-16
update pip before install

rmorshea:
As per https://github.com/kennethreitz/requests/issues/4006 the default version of `pip` used by travis is 6.x.x which has a problem installing requests. Even if that issue is resolved, we should probably be using the latest version of `pip` anyway.


minrk:
Thanks!

PR #409, opened 2017-05-28
Validates a correctly formatted FQDN if using string type verificatio…

crispybacon:
from RFC 1035
https://tools.ietf.org/html/rfc1035
The labels must follow the rules for ARPANET host names.  They must
start with a letter, end with a letter or digit, and have as interior
characters only letters, digits, and hyphen.  There are also some
restrictions on the length.  Labels must be 63 characters or less.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -2750,12 +2750,19 @@ class TCPAddress(TraitType):
 
     def validate(self, obj, value):
         if isinstance(value, tuple):
-            if len(value) == 2:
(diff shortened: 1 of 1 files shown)


minrk:
The trait you have modified defines itself as an `('ip', port)` tuple. I think the modification here no longer accepts ip addresses because the hostname regex is incorrect. It should allow at least: IPv4 or v6 addresses and hostnames. It's probably best to leave it allowing any string, rather than do strict validation of the string at this level.

PR #412, opened 2017-06-02
Trait Comparison Method

rmorshea:
Closes #276 

Adds a `compare` method to traits. This replaces a patch that handled equivalence checks which do not result in a boolean value. One layer above this, is a new `change` method that interprets the result of `compare`, and either forms an event that describes the comparison, or returns `None`. The owner of the trait is notified, only if `change` returns a description of the event.

# Todo

+ Add tests.
+ Better documentation.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -525,22 +525,37 @@ class TraitType(BaseDescriptor):
             return self.get(obj, cls)
 
     def set(self, obj, value):
-        new_value = self._validate(obj, value)
+        new = self._validate(obj, value)
         try:
-            old_value = obj._trait_values[self.name]
+            old = obj._trait_values[self.name]
         except KeyError:
-            old_value = self.default_value
+            old = self.default()
 
-        obj._trait_values[self.name] = new_value
-        try:
-            silent = bool(old_value == new_value)
-        except:
-            # if there is an error in comparing, default to notify
-            silent = False
-        if silent is not True:
+        obj._trait_values[self.name] = new
+
+        change = self.change(obj, old, new)
+        
+        if change is not None:
             # we explicitly compare silent to True just in case the equality
             # comparison above returns something other than True/False
-            obj._notify_trait(self.name, old_value, new_value)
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -525,22 +525,37 @@ def __get__(self, obj, cls=None):
...
+        if self.compare(old, new):
+            return Bunch(name=self.name, old=old,
+                new=new, owner=owner, type='change')
+        else:
+            return None
+
+    def compare(self, old, new):
+        """Compare old and new values.
+
+        While this often returns a boolean, it does not have to since
```
minrk: Clarify `this`. `this method` will always return True, whereas `==` may not. How about:

```python
def has_changed(self, old, new):
    """Return whether the value has changed

    While simple comparison usually returns a boolean,
    it may not. This method should always return a boolean.

    TraitTypes can override this method if simple comparison
    for their types is insufficient.
    """
    return old != new
```

rmorshea:
@minrk, @astrofrog, @dmadeka, @SylvainCorlay opinions on this?

traitlets/traitlets.py
```diff
@@ -525,22 +525,37 @@ def __get__(self, obj, cls=None):
...
-        if silent is not True:
+        obj._trait_values[self.name] = new
+
+        change = self.change(obj, old, new)
+        
+        if change is not None:
             # we explicitly compare silent to True just in case the equality
             # comparison above returns something other than True/False
-            obj._notify_trait(self.name, old_value, new_value)
+            obj._notify_trait(change.name, change.old, change.new)
```
minrk: I think this should be `notify_change` since `_notify_trait` just rebuilds the Change object you already have here.

traitlets/traitlets.py
minrk: I think this should also keep the try/except in the default implementation for backward compatibility. Introducing the method solves the problem, but we don't need to change the behavior in a backward-incompatible way.

Carreau:

PR #415, opened 2017-06-06
follow mro for trait default generators

minrk:
adds a regression test.

This appears to have been broken by the new default value generators, which don't resolve correctly on the mro.

From the looks of it, `_trait_default_generators` on classes should not include inherited generators, and lookups should instead walk the dicts on the mro every time.

cc @rmorshea

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -490,7 +490,7 @@ class TraitType(BaseDescriptor):
             value = obj._trait_values[self.name]
         except KeyError:
             # Check for a dynamic initializer.
-            default = cls._trait_default_generators[self.name](obj)
+            default = cls._get_trait_default_generator(self.name)(obj)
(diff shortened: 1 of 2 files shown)


rmorshea:
@minrk, makes sense.

traitlets/traitlets.py
```diff
@@ -490,7 +490,7 @@ def get(self, obj, cls=None):
             value = obj._trait_values[self.name]
         except KeyError:
             # Check for a dynamic initializer.
-            default = cls._trait_default_generators[self.name](obj)
+            default = cls._get_trait_default_generator(self.name)(obj)
```
rmorshea: We should use `obj.trait_defaults(self.name)` here.

minrk: Got it, thanks!

PR #420, opened 2017-06-21
BUG: remove redundant call to List.validate_elements

jakevdp:
Because of the ``super()`` call and the fact that ``validate_elements()`` is called from ``Container.validate()``, this validation does not need to be done again manually.


rmorshea:
You should be able to just remove the method entirely since the `if hasattr(obj, "validate"` check in `_validate` will catch the one defined by a parent class.

jakevdp:
Ah, good catch!

PR #424, opened 2017-07-01
refact(log): consolidate some duplicate log statements

ankostis:
Use string interpolation for printing log arguments instead of another multiple log statements.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -88,8 +88,7 @@ def catch_config_error(method, app, *args, **kwargs):
         return method(app, *args, **kwargs)
     except (TraitError, ArgumentError) as e:
         app.print_help()
-        app.log.fatal("Bad config encountered during initialization:")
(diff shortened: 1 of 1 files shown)


ankostis:
Note that the log-statement about change in [`Application.config` observer](../blob/master/traitlets/config/application.py#L321) fires even if the configuration dictionary is empty.
And this always happen at least once, when the class gets constructed (configurations are *loaded* later, on `initialize()`) so debug-logs have some redundant info.

Question:  Would it make sense to avoid a debug log-line if config is empty?
(it might still be useful when debugging cases where configs come up empty, while they shouldn't)

ankostis:
Also this log-statement in [`traitlets.config.Loader#718`](../blob/master/traitlets/config/loader.py#L718) deserves to be error-level.  On a respective problem for flags, below, it raises.

Should i fix add this in this PR?

PR #429, opened 2017-07-21
[WIP] Conferral Logic For Descriptors

rmorshea:
# Summary

If a descriptor has specified that it `inherits` attributes, and there are descriptors of the given ``name`` in the lineage of the ``cls`` which have bequeathed one or more of those attributes, then they are conferred to that descriptor. If there are attributes that should be inherited, but that have not been bequeathed, an error will be raised.

```python
class Grandparent(HasTraits):
    i = Int(help="something descriptive")

class Parent(Grandparent):
    # Has the same help string as Grandparent.i
    i = Int().inherit("help")

class Child(Parent):
    # raises an error because no descriptor 'i'
    # in the lineage of the class Child specified
    # that it was bequeathing its "default_value"
    i = Int().inherit("default_value")
```

# To Do

- [ ] Testing
- [ ] Docs for TraitType methods

# Possible Changes

+ Instead of having the methods `TraitType.inherit` and `TraitType.bequeath` we include them as argument in `TraitType.__init__`. I'm not sure what looks syntactically cleaner.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -352,6 +352,8 @@ class BaseDescriptor(object):
     """
 
     name = None
+    bequeaths = ()
+    inherits = ()
     this_class = None
 
     def class_init(self, cls, name):
@@ -366,6 +368,8 @@ class BaseDescriptor(object):
         """
         self.this_class = cls
         self.name = name
+        if self.inherits:
+            self._confer(cls, name)
 
     def subclass_init(self, cls):
         pass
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -352,6 +352,8 @@ class that does the following:
     """
 
     name = None
+    bequeaths = ()
```
minrk: What does `bequeaths` accomplish that `inherits` alone does not? Do we really need to specify an attribute as inheritable? Can we try a version with just `inherits`?

Carreau:

PR #431, opened 2017-07-24
WIP: calculation decorator for linking multiple inputs throught a calculation to a single output

maartenbreddels:
Sometimes you want to link the values of multiple traits to another one, using observe can be cumbersome, so I was using this decorator for it. The screenshot I think explains a lot. Not polished code, would like some feedback first. It seems quite similar to the transform in dlink, but that is one input to one output, this is multiple inputs to a single output (although I can imaging having multiple outputs as well).

screen shot 2017-07-24 at 21 40 49

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -326,6 +326,35 @@ class directional_link(object):
 
 dlink = directional_link
 
+def calculation(inputs, output, initial_calulation=True):
+    """Decorator: sets the output trait based on the outcome of a computation with multiple inputs
+
+    Parameters
+    ----------
+    inputs : [(object, attribute name), ...] list corresponding to arguments
+    output : (object, attribute name) pair
+    initial_calulation: boolean (optional)
+        Set the output value immediately
+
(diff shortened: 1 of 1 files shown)


minrk:
This is cool! (cc @SylvainCorlay for review)

jasongrout:
This is cool. I'm not sure this is the appropriate level to have a default attribute ('value'). That is pretty specific to ipywidgets, but I don't see many other HasTraits classes out there attach special meaning to a 'value' trait.

maartenbreddels:
The default is definitely up for discussion, but from a user point of view it's pretty user friendly. Whatever we decide on how choosing the default trait, we can do the same for `(js)(d)link`.
For widgets we can think about doing a `jscalcalation` variant, which can take a code snippet or (optionally) rely on [flexx transpiling](https://flexx.readthedocs.io/en/latest/pyscript/index.html), which would restore balance between `link/jslink` and `calculation/jscalculation`.

Carreau:

Issue #428, opened 2017-07-18
Any() trait not validated


jasongrout:
I'm seeing the following in traitlets 4.3.2. It appears that the validation for `x` is not getting called. If I change `x = Any(())` to `x = Tuple( () )`, the validator is called.

```
In [2]: from traitlets import HasTraits, Unicode, Any, validate, Tuple
   ...: class A(HasTraits):
   ...:     x = Any(())
   ...:     
   ...:     @validate('x')
   ...:     def f(self, proposal):
   ...:         print(proposal)
   ...:         return (1,2)
   ...: A().x
   ...: 
Out[2]: ()

In [3]: import traitlets
   ...: traitlets.__version__
   ...: 
Out[3]: '4.3.2'
```

minrk:
This is the general "default values are not cross-validated" issue. Default values are cross-validated in master ( #332), but in a way that causes infinite recursion if there are reference cycles in cross-validators. I'll see if I can fix the cycle without removing cross-validation of defaults altogether.

minrk:
Due to the problems encountered by cross-validating defaults seen in master, I think it's probably safest to close this as 'wontfix'. Default values are not cross-validated. We can consider coming back to it, but as mentioned in #434, the same can always be accomplished on an opt-in basis with a `@default` method that triggers the same validation logic.

PR #434, opened 2017-08-08
disable cross-validation of defaults

### LINKED ISSUE #363: Traitlets master breaks ipywidgets
I have not investigated yet, but we cannot make a release before we figure that one out.

minrk:
it causes loads of problems with nontrivial validation/observer combinations, including prematurely validating unused values and infinite recursion.

Given that validation of default values should only have an effect when those default values are initially invalid, there seems to be little benefit to introducing the complexity of validating defaults that may not be used.

With this change, ipywidgets tests pass again.

Closes #363
closes #435

#435 is an attempt to fix the problems without removing cross-validation of defaults. I think it's probably better to not cross-validate defaults, but it does get reported as a bug (#428), so we might give it a try.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -519,7 +519,8 @@ class TraitType(BaseDescriptor):
                 raise TraitError("No default value found for "
                     "the '%s' trait named '%s' of %r" % (
                     type(self).__name__, self.name, obj))
-            value = self._validate(obj, default)
+            with obj.cross_validation_lock:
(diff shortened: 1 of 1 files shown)


rmorshea:
As mentioned in other threads, I think that non-validated defaults make sense given the problem cases that arise otherwise. If there are more requests in the future to the contrary, there are backward compatible ways to enable it using the `@validate` or `@default` decorators.

minrk:
@rmorshea thanks!

PR #438, opened 2017-08-12
fix(help): finish #381 for RsT to print enum-choices

ankostis:
#381 inserted  a new line `Choices: x, y, z` into the help and config messages for all *Enum* traits, immediately above `Default: ...` line.
As had been mentioned in that PR, that feature was never added to the *RsT* documentation produced by `Configurable.class_config_rst_doc()` classs method.
This PR fixes that neglect.

+ Had to add an optional `as_rst=false` kwd to all`XxxEnum.info()` methods.
+ Added TCs for `info()`.

## RsT samples for `info()` method:
- `Enum` & `UseEnum` traits:

      any of ``'red'``|``'Green'``|``'BLUE'``|``'YeLLoW'`` or `None`

- `CaselessStrEnum`:

      any of ``'red'``|``'Green'``|``'BLUE'``|``'YeLLoW'`` (case-insensitive) or `None`

- `FuzzyEnum`:

      any case-insensitive prefix of ``'red'``|``'Green'``|``'BLUE'``|``'YeLLoW'`` or `None`

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -2272,12 +2272,18 @@ class Enum(TraitType):
             choices = repr(list(choices))
         return choices
 
-    def info(self, as_rst=False):
+    def _info(self, as_rst=False):
         """ Returns a description of the trait."""
         none = (' or %s' % ('`None`' if as_rst else 'None')
                 if self.allow_none else
                 '')
         return 'any of %s%s' % (self._choices_str(as_rst), none)
+
+    def info(self):
+        return self._info(as_rst=False)
+
+    def info_rst(self):
+        return self._info(as_rst=True)
 
 
 class CaselessStrEnum(Enum):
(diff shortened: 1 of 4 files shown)


traitlets/traitlets.py
```diff
@@ -2263,12 +2263,21 @@ def validate(self, obj, value):
...
+    def _choices_str(self, as_rst=False):
+        """ Returns a description of the trait choices (not none)."""
+        choices = self.values
+        if as_rst:
+            choices = '|'.join('``%r``' % x for x in choices)
+        else:
+            choices = repr(list(choices))
+        return choices
+
+    def info(self, as_rst=False):
```
minrk: Adding an `as_rst` arg is a backward-incompatible change because user traits can have defined `.info()` accepting no arguments, which must not result in any errors.

Maybe better to define a new `.info_rst()` method that calls `.info()` by default.

ankostis: ok

PR #441, opened 2017-08-15
allow traits to parse string config

minrk:
This is related to the new argparse CLI parsing and envvar config in #439 / #217

Adds `trait.from_string` for loading values from a strings-only source (e.g. CLI, env).

Config coming from CLI (or env vars) always come as strings. We have been using `ast.literal_eval` on these, principally because in the early traitlets config design, parsing had to happen before the target traits were known. This has been the source of many problems.

This assumption has changed, as aliases and config parsing *are* given the classes to be configured, so that traits can be relied upon to parse values.

For the fallback `--Class.trait` case, we can also solve this problem by deferring parsing until config loading. This means that config passed that isn't in the official supported class list will always be str (a str subclass for backward-compatibility), and call `trait.from_string(s)` at config load time for the given class. Any config values that in fact never get loaded will be left as un-eval'd strings.

It also eliminates the `eval` in most cases, so bugs related to that should be fixed. Most importantly, it eliminates the eval on Unicode traits, where it was always the wrong thing to do.

Backward-compatibility issues (why I would like to get this in 5.0, along with the other edge-case-breaking CLI changes in the new argparse handling):

- Due to the use of eval, some cases  put quotes *inside* string arguments, so that
  `eval("1")` would get `"1"` instead of `1`. Eliminating the eval would result in these traits
  getting `'"1"'` (failing to strip quotes). Alternately, stripping quotes would make it impossible for traits to get values *with* quotes from the CLI. This is already the case now due to the use of `eval`.
- Unparsed config from the CLI will be strings in the config object.
  This means anything reaching into the config object that relied on `eval`
  would no longer get the eval'd object (typically an integer),
  but instead the DelayedEvalString.
  For any cases accepting a text object, this should be fine,
  but cases where it is something else (e.g. an int), this is a breaking change.

TODO:

- [x] Handle container traits. The new argparse syntax has made this a little more complicated 
      because a trait may come in as a string literal (old way) or a list of values.
      We need to get the right trait to deal with this.
- [x] Deal specially with quotes on strings, which were swallowed by `eval`

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -317,8 +317,9 @@ class DeferredConfigString(text_type):
     in the configurable classes.
 
     When config is loaded, `trait.from_string` will be used.
-    
-    If an error is raised
+
+    If an error is raised in `.from_string`,
+    the original string is returned.
 
     .. versionadded:: 5.0
     """
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -317,12 +317,21 @@ class DeferredConfigString(text_type):
     in the configurable classes.
 
     When config is loaded, `trait.from_string` will be used.
+    
+    If an error is raised
 
     .. versionadded:: 5.0
     """
     def get_value(self, trait):
         """Get the value stored in this string"""
-        return trait.from_string(text_type(self))
+        s = text_type(self)
+        try:
+            return trait.from_string(s)
+        except Exception:
+            # exception casting from string,
+            # let the original string lie.
+            # this will raise a more informative error when config is loaded.
+            return s
 
     def __repr__(self):
         super_repr = super(DeferredConfigString, self).__repr__()
@@ -546,7 +555,12 @@ class CommandLineConfigLoader(ConfigLoader):
                         "Use %r instead of %r" % (rhs, old_rhs),
                         DeprecationWarning)
         if trait:
-            return trait.from_string(rhs)
+            try:
+                return trait.from_string(rhs)
+            except Exception:
+                # failed to eval, let later config loading raise,
+                # which will be more informative
+                return rhs
         else:
             return DeferredConfigString(rhs)
 
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -5,13 +5,14 @@
 # Distributed under the terms of the Modified BSD License.
 
 import argparse
+from ast import literal_eval
 import copy
 import functools as fnt
 import os
 import re
 import sys
 import json
-from ast import literal_eval
+import warnings
 
 from ipython_genutils.path import filefind
 from ipython_genutils import py3compat
@@ -321,7 +322,7 @@ class DeferredConfigString(text_type):
     """
     def get_value(self, trait):
         """Get the value stored in this string"""
-        return trait.from_string(self)
+        return trait.from_string(text_type(self))
 
     def __repr__(self):
         super_repr = super(DeferredConfigString, self).__repr__()
@@ -533,6 +534,17 @@ class CommandLineConfigLoader(ConfigLoader):
 
     def _parse_config_value(self, rhs, trait=None):
         """Python-evaluates any cmd-line argument values."""
+        rhs = os.path.expanduser(rhs)
+        if len(rhs) > 2:
+            # handle deprecated "1"
+            for c in "'\"":
+                if rhs[0] == rhs[-1] == c:
+                    old_rhs = rhs
+                    rhs = rhs[1:-1]
+                    warnings.warn(
+                        "Supporting extra quotes around strings is deprecated in traitlets 5.0. "
+                        "Use %r instead of %r" % (rhs, old_rhs),
+                        DeprecationWarning)
         if trait:
             return trait.from_string(rhs)
         else:
@@ -542,11 +554,9 @@ class CommandLineConfigLoader(ConfigLoader):
         """execute self.config.<lhs> = <rhs>
 
         * expands ~ with expanduser
-        * tries to assign with literal_eval, otherwise assigns with just the string,
-          allowing `--C.a=foobar` and `--C.a="foobar"` to be equivalent.  *Not*
-          equivalent are `--C.a=4` and `--C.a='4'`.
+        * interprets value with trait if available
         """
-        if isinstance(trait, Dict):
+        if isinstance(trait, Dict) and not isinstance(rhs, string_types):
             if len(rhs) == 1 and isinstance(rhs[0], string_types):
                 # check for deprecated --Class.trait="{'a': 'b'}"
                 self.log.warning(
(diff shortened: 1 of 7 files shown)


ankostis:
A relevant problem to this PR is when CLI provides a string to some `Unicode` trait that happens to be all digits .
In that case, the old loader kicks-in and `eval()` converts it into an int, and then tait chocks on it.

Simple test code:
```python
class MyApp(Application):
    a = Unicode().tag(config=True)

MyApp().initialize(['--MyApp.a=1234'])

[MyApp] CRITICAL | Bad config encountered during initialization:
[MyApp] CRITICAL | The 'a' trait of a MyApp instance must be a unicode string, but a value of 1234 <class 'int'> was specified.
```

minrk:
Indeed, and that's precisely the longstanding problem this aims to fix, which I tried to describe above. I've added a test case to make sure that it's covered. The most public example in the real world is:

    ipython -c 1

which is fixed by this PR.

minrk:
In fact, this logic is making me reconsider the argparse handling of lists and dicts that's currently in master. It's quite verbose and the implementation is rather complicated. With `.from_string`, we could enable things like:

    command --list-trait a,b,c

and parse that out, which seems a lot nicer than:

    command --list-trait a --list-trait b --list-trait c

and would be vastly simpler to implement. A disadvantage of the current mechanism is that, like the aborted env handling, it behaves differently for traits registered in the application.classes list vs classes not in that list (e.g. extensions), which is undesirable. If we leaned more heavily on from_string, then the difference would be eliminated, which seems desirable.

ankostis:
I would consider argparse a given and not go back.

Yopu see, old code breaks with quoting in lists; otoh, arparse parses dict-items with ease(e.g. `-D key1="va;l=ue" -D key2=bar`).  That's an impossible feat to do for POSIX and Windows with old-code.  And I remind, that was my main motive for argparse, to submit arbitrary key-value pairs (that cannot be trait-letized).

Consequently, i keep theopposite view (however that might weight for this project):  it's *old-code* is burdening *argparse*.
In my vendorized copy actually I now use a flag to disable old eval-code completely.

So, if this PR manages to support also old-format, then really, why do we still need the keep old code around?

minrk:
>I would consider argparse a given and not go back.

Indeed, and we have always used argparse. I wouldn't propose dropping argparse.

> In my vendorized copy actually I now use a flag to disable old eval-code completely.

That makes sense, and I would encourage you to continue to do that. If in your application all configurable traits are known at the time of constructing the parser, you can do a much better job than the traitlets default, even what we have now in master. For instance, you should not be using `type=text` for anything but actual text types, instead using `type=int` for Integers, and proper actions for containers, etc. Right now, I think we have a bit of a mess in the middle that's complicated by trying to preserve old behavior, while not achieving the goal of truly making the most of argparse. A parser that completely ignores the `--Class.trait=value` would be greatly simpler than what we have, it just can't be the default for Application. Nothing in the traitlets package prevents you from having the argument parsing you would prefer.

It's totally reasonable to use a different CLI parser in your own application, and I think you probably should unless you have the same goals as IPython. There's no reason that every possible parsing pattern needs to ship in traitlets.config, whose primary goal is to serve as a base for IPython/Jupyter applications.

This is the problem that I see with what we have in master:

    ipython --Class.dict_trait a=5

behaves differently, depending on whether Class is in IPython's preconfigured class list or not. That's not a good thing, and I think we must fix that before releasing 5.0. One of the primary goals of the IPython config system is that all objects, IPython's and those of any extension, are exposed the *same way* on the CLI and in every config file. This is not the case in master.

That doesn't need to extend to the aliases, which can get some help because the traits to which they map are required to be known. I think it's fine to have `-D a=5`, but I *don't* think that should extend to `--Class.d a=5` unless it is done in a way that extends to `Class.d` where `Class` is unknown at parsing time.

> So, if this PR manages to support also old-format, then really, why do we still need the keep old code around?

I'm not sure what old code you are referring to here.

ankostis:
> ...If in your application all configurable traits are known at the time of constructing the parser

Yes it does. 
And I consider it a bug not to include some class.
Otherwise...your analysis explains better the complications.

> > So, if this PR manages to support also old-format, then really, why do we still need the keep old code around?
>
> I'm not sure what old code you are referring to here.

`loader.KeyValueConfigLoader` and this particular this line https://github.com/ipython/traitlets/blob/4da01bfdd675783ab64a21df015e49378d91cbd2/traitlets/config/loader.py#L518

ankostis:
> Nothing in the traitlets package prevents you from having the argument parsing you would prefer.

Actually before #360, you it wasn't taht easy to switch loader. But anyway, what you suggest is to change the default loader, or drop and remove code from #322 completely?

traitlets/config/loader.py
```diff
@@ -715,10 +792,14 @@ def load_config(self, argv=None, aliases=None, flags=None):
...
-                    self.log.warning("Unrecognized alias: '%s', it will probably have no effect.", raw)
+                    self.log.warning("Unrecognized alias: '%s', it will have no effect.", raw)
+                    trait = None
+                else:
+                    trait = self._find_trait(lhs)
                 try:
-                    self._exec_config_str(lhs, rhs)
+                    self._exec_config_str(lhs, rhs, trait)
                 except Exception:
+                    raise
```
minrk: Good reminder. This is a debug bit because I think this shouldn't raise anymore unless there's a bug. It should be gone when this is done.

minrk:
> Actually before #360, you it wasn't taht easy to switch loaders.

Good point. And what you've raised is part of why it's important that it be possible to select different loaders. Overriding `parse_command_line` has always been possible, but that would require more copy/paste than should be necessary.

> what you suggest is to change the default loader, or drop and remove code from #322 completely?

I'd rather not revert it. It's an improvement to a problem we've had for a long time: no good way to expose container traits on the command-line. It's just not quite the complete solution we need. There are three cases: `--short` aliases, `--Class.trait` in the classes list, and `--Class.trait` *not* in the classes list. I think it would be better to either:

1. confine it to `--flag` short aliases, or
2. Find a way to extend it to `--Class.trait` where Class is not in the config list.

Either way, I think it's important to remove the distinction case 2 and 3.

Now, `from_string` makes this easy as long as args can only be one string. But as you pointed out, letting the shell split items is nicer than that. Maybe there's something like an `item_from_string` approach that will let us keep the improvements you've made while simplifying things for `--Class.trait` to defer to the traits more.

ankostis:
Why not add a enable/disable flag that by-default is confines *argparse* to flags/aliases, and if the library-user trusts his `classes` list contains everything, he can switch it `True` to start using it for them also?
Obviously, *jupyter/ipython* will work flawlessly by default.

ankostis:
Correction: it may be not a new flag, but just an unused `loader`?
(but i guess it's a bit more work).

minrk:
I'd like to find a way to keep it across traits rather than confine it to aliases, but that can be the fallback if we can't find a good solution.

Carreau:

PR #447, opened 2017-08-19
feat(traits): Implement ENV-VARS in traits-level

ankostis:
As discussed in #439, i extracted here the code related to traits-level only.
Specifically:
- added `evvar='MY_VAR'` tag with higher priority over defaults;
- modified the respective printout routines in `Configurable` just to append a new `'# 
 Environment variable: MY_VAR'` line in help/rst/config-file messages.

~Two~ These open issues remain from #439:
- [ ] Documentation: Where to describe the new capability? 
  - [ ] collect all metadata usages in some place? YES
  - [ ] Explain that `config-files < env-var < CLI` order is not obeyed.
- [x] Should we print the envvar-value in the help description, if var exists? NO 
- [x] Include non-config tagged taits in print-outs for class-help, config-file headers & RsT docs.
- [x] Parse env-vars using #441 machinery.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -9,7 +9,7 @@ from __future__ import print_function, absolute_import
 from copy import deepcopy
 import warnings
 
-from .loader import Config, LazyConfigValue, _is_section_key
+from .loader import Config, LazyConfigValue, DeferredConfigString, _is_section_key
 from traitlets.traitlets import (
     HasTraits,
     Instance,
@@ -165,6 +165,9 @@ class Configurable(HasTraits):
                         # without having to copy the initial value
                         initial = getattr(self, name)
                         config_value = config_value.get_value(initial)
+                    elif isinstance(config_value, DeferredConfigString):
+                        # DeferredEval tends to come from CLI/environment variables
+                        config_value = config_value.get_value(self.traits()[name])
                     # We have to do a deepcopy here if we don't deepcopy the entire
                     # config object. If we don't, a mutable config_value will be
                     # shared by all instances, effectively making it a class attribute.
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -230,10 +230,12 @@ class Configurable(HasTraits):
         final_help = []
         base_classes = ', '.join(p.__name__ for p in cls.__bases__)
         final_help.append(u'%s(%s) options' % (cls.__name__, base_classes))
-        final_help.append(len(final_help[0])*u'-')
-        for k, v in sorted(cls.class_traits(config=True).items()):
-            help = cls.class_get_trait_help(v, inst)
-            final_help.append(help)
+        final_help.append(len(final_help[0]) * u'-')
+        traits = cls.class_traits(config=True)
+        traits.update(cls.class_traits(envvar=lambda ev: bool(ev)))
+        for _, trait in sorted(traits.items()):
+            thelp = cls.class_get_trait_help(trait, inst)
+            final_help.append(thelp)
         return '\n'.join(final_help)
 
     @classmethod
@@ -276,7 +278,8 @@ class Configurable(HasTraits):
 
         env_var = trait.metadata.get('envvar')
         if env_var:
-            env_info = 'Environment variable: %s' % env_var
+            only = '' if trait.metadata.get('config') else '(only)'
+            env_info = 'Environment variable%s: %s' % (only, env_var)
             lines.append(indent(env_info, 4))
 
         if inst is not None:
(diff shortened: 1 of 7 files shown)


minrk:
Thanks for doing this one! I think it's all we should try for in 5.0 (and probably ever, in this repo, at least).

> Where to describe the new capability?

I think an example should be added to the traits in the docs (`using_traitlets').

> Maybe collect all metadata usages in some place?

That's a good idea. I think it's just `config` and `envvar` right now.

minrk:
> Should we print the envvar-value in the help description, if var exists?

I wouldn't, just the env is good enough.

ankostis:
Discovered just one more open-issue: 
printout messages do NOT include traits tagged as `(config=False, envvar=True)`.
For instance, in `Configurable.class_get_help()` https://github.com/ankostis/traitlets/blob/2aaef410db838393df882ea73d0479b9c09ffbfb/traitlets/config/configurable.py#L234

We can  "join" the results of 2 `metadata` queries like that:

```diff
-         for k, v in sorted(cls.class_traits(config=True).items()):
+         traits = cls.class_traits(config=True)
+         traits.update(cls.class_traits(envvar=True)
+         for k, v in sorted(traits.items()):
```
But that would convey the *false impression* that all these traits are configurables (ie they can be defined in files/cli).
Any ideas?

ankostis:
For non-`config` traits, 201db878 added `"(only)" in their info line:

    #    Environment variable(only): MY_ENVVAR

minrk:
> Discovered just one more open-issue:
printout messages do NOT include traits tagged as (config=False, envvar=True).

Indeed, and it shouldn't. From the config perspective, env vars are not 'configurable', in that they are not influenced by the config object. The help output is part of `traitlets.config`, and with this PR, envvar is not part of config at all, so envvar info should either be:

1. entirely eliminated from help output, or
2. limited only to those with `config=True`.

My inclination is to exclude it entirely, as it is purely a traitlets default feature, and not a config feature. This further clarifies that traitlets.config does not support environment variables at all.

traitlets/config/configurable.py
```diff
@@ -230,10 +230,12 @@ class defaults.
         final_help = []
         base_classes = ', '.join(p.__name__ for p in cls.__bases__)
         final_help.append(u'%s(%s) options' % (cls.__name__, base_classes))
-        final_help.append(len(final_help[0])*u'-')
-        for k, v in sorted(cls.class_traits(config=True).items()):
-            help = cls.class_get_trait_help(v, inst)
-            final_help.append(help)
+        final_help.append(len(final_help[0]) * u'-')
+        traits = cls.class_traits(config=True)
+        traits.update(cls.class_traits(envvar=lambda ev: bool(ev)))
```
minrk: make sure that `config=False, envvar` traits are *not* included here.

minrk:
> Explain that config-files < env-var < CLI order is not obeyed.

Re: docs, we should make sure that the envvar support is documented with the traitlets default generators, and *not* mentioned anywhere in the traitlets.config docs.

ankostis:
I repeat, envvar-only traits are irrelevant for configuration-files, and hadn't ever added them here.

But for applications, they ARE relevant - just not affected by files, only ENVVARS will do it.
Actually those are the only envvar-traits that will work *deterministically* - there rest will work only if not defined in config-files.

So I vote once more to keep it as it is, bc for pragmatic reasons, they do help users configure their apps.  
Maybe there is a way to make it more obvious the above to the user.

ankostis:
> > Maybe collect all metadata usages in some place?
> 
> That's a good idea. I think it's just `config` and `envvar` right now.

There is also the tag `multiplicity` from `argparse` loader: https://github.com/ipython/traitlets/blob/4da01bfdd675783ab64a21df015e49378d91cbd2/traitlets/config/loader.py#L904

minrk:
> There is also the tag multiplicity from argparse loader:

Ah, right. I think we should move those to an `argparse` namespace in metadata so that it's a passthrough to argparse rather than dedicating `multiplicity` to both `action` and `nargs`.

minrk:
> Finally, for RsT docs, they should be there - it's library documentation.

That's not true. It's a different rendering of the config options:

> Generate rST documentation for this class' config options.

It should be the same information as the other two places, and exclude any `config=False` traits. If you want to document all of the traits of a class that would be useful, but is not a feature that currently exists and should not be done via the `class_config` methods.

ankostis:
> That's a good point. It's okay to keep the env for `config=True` traits. But please exclude all `config=False traits`.

I know it's a good point.
Still, it didn't change your mind, not even a bit :-)

Carreau:

PR #448, opened 2017-08-19
feat(config): support MERGE-PRIORITIES for `Config` items

ankostis:
(I hope you don't get too mad at me for trying...)

Retrofitted `loader.Config` class to support merging-priorities for specific values based on ranks (e.g. integers).  Now `update_config()` always respects priorities.  The purpose it to use it to fully support also #99 environment-variable priorities.

For this PR, the new features are used also to deprecate the need to re-update `Application.cli_config`  on top of file-configs.  

## Implementation
- A lot of effort was put to have code and the memory-footprint as small as possible.  
  - `Config` instances by default don't waste extra memory - default priority `resolved from _merge_priority` class attribute.
  - Possible to set a single rank(number) that propagates to all config-object values; these apply to 1st level only.  But it can be set recursively to inner config-objects (wasting an in per config-object).
  - To support different ranks in the same config-object, the `_merge_priority` is upgraded to dict.
  - The original idea in #439 of storing *tags* (a new dict) into config ibjects would be an overkill.
  - Have to study performance-tax on the merges.
- A new `Config.substract()` method is used to limit value-changes, taking into account priority-ranks.
  This is useful to preserve traits that have been directly modified between loading of configs from different sources.
- The *ranks* must be comparable. besides that, the code does not enforce any other limitation on their type APART from setting the default `Config._merge_priority = 0` which in effect "force" the use of integers.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -64,6 +64,9 @@ subcommand 'cmd', do: `{app} cmd -h`.
 # Application class
 #-----------------------------------------------------------------------------
 
+CFG_RANK = 0
+ENV_RANK = 10
+CLI_RANK = 20
 
 
 _envvar = os.environ.get('TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR','')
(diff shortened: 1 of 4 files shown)


minrk:
I appreciate the effort, but as I described in #439, I think we should not introduce the concept of priority to config objects.

ankostis:
For whoever lands here, latest changes for the same functionality in [ankostis@confvars](https://github.com/ankostis/traitlets/tree/confvars).

PR #449, opened 2017-08-19
feat(app): support ENV-VAR with priority against CLI-configs

ankostis:
(I really hope you don't get too mad at me for trying :-))

After #439 study, this the 2nd attempt to implement environment-variables with proper priorities for configuration sources (as described in #99) following 
- the insight of @minrk (to construct a proper config-instance that will propagate wherever the correct values) and 
- building on the idea for config's MERGE-PRIORITIES.

## Implementation
- It is based(has merged) on Config-Priorities(#) and Traits-Env-Vars(#447).
- The `Configurable.update_config(#448)` is now modified to ignore any config-values below priority-20 (FIXED atm) for which there are variables defined in the environment.
- There is no need to define a new keword `skip_env` since config- priorities lifted this burden.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -64,6 +64,9 @@ subcommand 'cmd', do: `{app} cmd -h`.
 # Application class
 #-----------------------------------------------------------------------------
 
+CFG_RANK = 0
+ENV_RANK = 10
+CLI_RANK = 20
 
 
 _envvar = os.environ.get('TRAITLETS_APPLICATION_RAISE_CONFIG_FILE_ERROR','')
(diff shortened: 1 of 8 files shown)


minrk:
Thanks for this as an illustrative example, but I don't think we should merge it. I think we should stop at #447.

ankostis:
I remind that #447 had the nasty side effect that config-files overwrite env-vars.  Practically you cannot *reliably* use the simple bash-ism below, because the `foo` trait might be already defined in the config file:


    FOO_VAR="abs" jupyter notebook

ankostis:
For whoever lands here, latest changes for the same functionality in [ankostis@envvars4](https://github.com/ankostis/traitlets/tree/envvars4).

Issue #455, opened 2017-10-08
changing default value of containers to None


djarecka:
Hi,

I wanted to create my own trait types, so all have the same `default_value`, e.g. `None`, and call `tag` automatically,  but I'm having problems with `default_value` for containers. I will use `List` as an example, I'm defining a new class:

```
class List(traitlets.List):
    allow_none = True
    default_value = None 
    info_text = "a list with default_value = None"

    def __init__(self, trait=None, default_value=None, minlen=0, maxlen=sys.maxsize, kw=None,
                 read_only=None, help=None, config=None, **kwargs):
        super(List, self).__init__(trait=trait, default_value=default_value, minlen=0, maxlen=maxlen,
                                   kw=kw, read_only=read_only, help=help, config=config)
        self.tag(**kwargs)
``` 

and I'd like my new class to pass those test:

```
def test_list_default_1():
    class HasTr(traitlets.HasTraits):
        foo = List()
    ht = HasTr()
    assert ht.foo is None

def test_list_default_2():
    class HasTr(traitlets.HasTraits):
        foo = List(default_value=[1])
    ht = HasTr()
    assert ht.foo == [1]
```

I'm passing the first test, but not the second (in both cases `ht.foo is None`). From quick reading of your code, it looks like the method `make_dynamic_default` is not called when I define my `List` class with class attribute `default_value = None`. Is this on purpose? Do you have any other suggestions how I can write my class (for all traits) that I'm able to pass both tests?

**system Info**
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/dorota/anaconda/envs/py35_trlets/lib/python3.5/site-packages/IPython',
 'ipython_version': '6.1.0',
 'os_name': 'posix',
 'platform': 'Darwin-16.1.0-x86_64-i386-64bit',
 'sys_executable': '/Users/dorota/anaconda/envs/py35_trlets/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.5.3 | packaged by conda-forge | (default, May 12 2017, '
                '15:35:12) \n'
                '[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]'}

I'm using the current master.

rmorshea:
[...]

PR #456, opened 2017-10-09
adding tests for default values for traitlets types and new traits (i…

djarecka:
…nt and list only for now)

examples  of  tests for default value of original traits and new defined traits. Use only `Int` and `List` as examples for now, can write more.

Two of those test fail, see #455. Should I mar as `xfail`?

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -2625,3 +2625,82 @@ def test_override_default_instance():
     c._a_default = lambda self: 'overridden'
     assert c.a == 'overridden'
 
+
+# adding additional tests to test default values for original types and new ones
+
+@pytest.mark.parametrize("arg_lst, arg_dict, foo_val", [
+        ([],  {},                      0),
+        ([],  {"default_value": 1},    1),
+        ([1], {},                      1)
+        ])
+def test_int_default(arg_lst, arg_dict, foo_val):
+    class HasTr(HasTraits):
+        foo = Int(*arg_lst, **arg_dict)
+
+    ht = HasTr()
+    assert ht.foo == foo_val
+    assert type(ht.foo) is int
+
+
+class NewInt(Int):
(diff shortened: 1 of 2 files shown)


rmorshea:
@djarecka, I'm pretty certain that the issue doesn't have to do with subclassing, but is actually part of how `Container.__init__` is handling `Container(x)` vs `Container(default_value=x)`. As a result, you may be testing the same regression, in all tests.

I need to verify this though.

djarecka:
@rmorshea not sure if I understand your point. I'm assuming that `List([1])` should give the same as `List(default_value=[1])`, but just want to have both cases in my tests. And both cases give the same error for the test `test_newlist_default`

rmorshea:
@djarecka, ok I think I found the issue, and it does, in the case of `master` have to do with what happens when you subclass `Instance` with a different `default_value` class attribute. Thus, I think you're testing the issue correctly.

How to resolve this? We'll basically need to set `self.default_value = Undefined` [here](https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L2439).

djarecka:
This indeed helps. I can test for other containers tomorrow.

djarecka:
@rmorshea - I added more tests, but unfortunately some tests for `Tuple` and `Dict` still fails. 

`Tuple` doesn't even want to take `default_value=None`, so I had to change to `default_value=Undefined` in `NewTuple`.  But it only solves one problem.

rmorshea:
I'll look into it. `Dict` and `Tuple` pretty seriously override `Container.__init__` so I sort of expected this.

djarecka:
Thanks! Not sure if that helps, but I know I can pass my tests for `NewDict` if I remove completely the class attribute `default_value`, i.e. remove [this line](https://github.com/ipython/traitlets/pull/456/files#diff-7f3523ff5c29fcbcde995ac93a155a51R2796).

Carreau:

PR #458, opened 2017-11-10
Add Sphinx extension and autogen function for documenting config options

takluyver:
This is the machinery we prototyped for [documenting IPython config options](http://ipython.readthedocs.io/en/stable/config/options/terminal.html). By defining a simple Sphinx extension, individual config options can be linkable, cross-referenceable, and nicely searchable.

It was always my intention to move this into traitlets so we can easily use it for other Jupyter applications. I'm not aware of any problems since we started using it for IPython.

### DIFF
--- a/traitlets/config/sphinxdoc.py
+++ b/traitlets/config/sphinxdoc.py
@@ -1,0 +1,159 @@
+"""Machinery for documenting traitlets config options with Sphinx.
+
+This includes:
+
+- A Sphinx extension defining directives and roles for config options.
+- A function to generate an rst file given an Application instance.
+
+To make this documentation, first set this module as an extension in Sphinx's
+conf.py::
+
+    extensions = [
+        # ...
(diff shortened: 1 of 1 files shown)


rmorshea:
@takluyver Out of curiosity, does this supersede some prior automatic documentation systems that have been proposed, or would this compliment one?

takluyver:
I'm not sure which systems you're referring to. In some ways it supersedes the method in traitlets to generate rst docs, but the new function produces rst that is only valid in Sphinx with the extension loaded, so it's not a drop-in replacement for the older one.

takluyver:
On the face of it, I think this is complementary to #261. That's focused on documenting traitlets within a Python API, for interactive inspection and for API docs. This is specifically about documenting traitlets used as application config - so the starting point is an application rather than a class, it only selects configurable traits, and it focuses on things like command line aliases.

The main motivation for this is that I want to be able to link people to individual config options in our docs.

PR #462, opened 2017-12-07
drop Python 3.3, update to use 3.6

maartenbreddels:


setup.py
```diff
@@ -90,7 +90,6 @@
     # -- SUPPORT UNIFORM-WHEELS: Extra packages for Python 2.7, 3.3
```
blink1073: this line mentions 3.3 as well

blink1073:
This is another option: https://github.com/ipython/ipykernel/blob/master/setup.py#L105.

@minrk, I'd say dropping support for 3.3 is fine, but I'll defer to you here.

blink1073:
The end of life passed two months ago: https://www.python.org/dev/peps/pep-0398/#x-end-of-life

maartenbreddels:
I think this is a fluke, it had green lights before

minrk:
yup, fine. Don't forget to set `requires_python` when dropping Python support.

blink1073:
@maartenbreddels, for reference: https://packaging.python.org/tutorials/distributing-packages/#python-requires

minrk:
Thanks!

maartenbreddels:
👍

PR #465, opened 2018-01-25
[WIP] add absolute name

rmorshea:
# Summary

Allows a nested trait to know that attribute it is assigned to (i.e. it's "absolute name").

This PR is in anticipation of a downstream implementation of Eventful traits.

# Details

We add `BaseDescriptor._parent` which is `None` if it is the root descriptor. Otherwise it holds a weak reference to its parent trait. There is also a new method `BaseDescriptor._lineage()` which will recurse through its parents until it finds a descriptor with `name` that is not `None`.

These additions are leveraged by:

+ The new `BaseDescriptor.absolute_name` property which returns the attribute name the trait exists on.
+ A new `__str__` method for `BaseDescriptor` which will describe its lineage if it exists on a class.
+ and `TraitType.error` to more simply report error messages.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -437,15 +437,17 @@ class BaseDescriptor(object):
             yield parent
 
     def __str__(self):
-        lineage = list(self._lineage())
-        absolute_name = lineage[-1].name
-        info = " of ".join(
-            ["the " + type(self).__name__] +
-            [describe("a", t) for t in lineage[1:]]
-        )
-        if self.this_class is not None and absolute_name is not None:
+        if self.this_class is not None:
+            lineage = list(self._lineage())
+            absolute_name = lineage[-1].name
+            info = " of ".join(
+                ["the " + type(self).__name__] +
+                [describe("a", t) for t in lineage[1:]]
+            )
             class_name = describe(None, self.this_class, verbose=True)
             info += " at %s.%s" % (class_name, absolute_name)
+        else:
+            info = super(BaseDescriptor, self).__str__()
         return info
 
 
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -47,6 +47,7 @@ import re
 import sys
 import types
 import enum
+from weakref import ref
 try:
     from types import ClassType, InstanceType
     ClassTypes = (ClassType, type)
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
@rmorshea Why did you close this?

rmorshea:
@SylvainCorlay, this is now integrated into #466. I think I was making subtle changes there and didn’t want to have to keep updating this PR.

traitlets/traitlets.py
```diff
@@ -392,8 +393,13 @@ class that does the following:
...
     name = None
     this_class = None
+    _parent = None
 
-    def class_init(self, cls, name):
+    @property
+    def absolute_name(self):
+        return list(self._lineage())[-1].name
+
+    def class_init(self, cls, name, parent=None):
```
minrk: How do we handle custom TraitTypes that have defined `class_init(cls, name)` without breaking them?

traitlets/traitlets.py
```diff
@@ -421,6 +429,27 @@ def instance_init(self, obj):
...
+        parent = self
+        yield parent
+        while parent._parent is not None:
+            parent = parent._parent()
+            yield parent
+
+    def __str__(self):
+        if self.this_class is not None:
+            lineage = list(self._lineage())
+            absolute_name = lineage[-1].name
```
minrk: If this is the only thing the parent tree is used for and the name is always `lineage[-1].name`, why isn't `name` just passed down the lineage instead of recording the lineage itself?

rmorshea: @minrk we use the lineage in the next line down to indicate how traits are nested.

If we don't do this `obj.attr = (1, 2, 3)` for a `Tuple(String())` will give an error like:

```
The 'attr' trait of MyClass instances expected a string, but got an int instead.
```

This is quit confusing since the bad value is not the tuple, but the int inside the tuple.

traitlets/traitlets.py
rmorshea: @minrk the best we can do is check the function signature I guess.

traitlets/traitlets.py
minrk: If parent is truly needed, the other option is to do attribute assignment, rather than an API change:

```python
trait.parent = self
trait.class_init(obj, name)
```

ankostis: Now that you are on it, to me the message:

    The 'attr' trait of MyClass instances expected ..,

is disorientating.  It would be faster to grasp if it said:

    The `MyClass.trait` trait expected ...

rmorshea:
https://github.com/ipython/traitlets/pull/466 has been closed in favor of https://github.com/rmorshea/spectate

PR #468, opened 2018-02-03
add a clear method

### LINKED ISSUE #106: Replacement for "set_default_value"?
In knitpy, I use `set_Default_value` to reset (parts of) a traitlet based object to a known state. Unfortunately, this doesn't work anymore with the new standalone version of traitlets (after https://github.com/ipython/traitlets/pull/26):

``` python
    def execution_finished(self):
        self.output.flush()
        reset_needed = ["engine", "mode"
                        "chunk_label", "comment",
                        "include", "echo",  "include", "results"]
        for name in self.trait_names():
            if name in reset_needed:
                self.traits()[name].set_default_value(self)
```

-> https://github.com/JanSchulz/knitpy/blob/master/knitpy/knitpy.py#L853

What is the suggested way to "reset" a trait?

rmorshea:
closes #106

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -857,6 +857,24 @@ class TestHasTraits(TestCase):
         # should raise TypeError if no positional arg given
         self.assertRaises(TypeError, A)
 
+    def test_clear(self):
+        class A(HasTraits):
+            i = Int().tag(key="this")
+            j = Int().tag(key="this")
+            k = Int().tag(key="that")
+
+        a = A(i=1, j=2, k=3)
+        a.clear()
+        assert a.i == 0
+        assert a.j == 0
+        assert a.k == 0
+
+        a = A(i=1, j=2, k=3)
+        a.clear(key="this")
+        assert a.i == 0
+        assert a.j == 0
+        assert a.k == 3
+
 #-----------------------------------------------------------------------------
 # Tests for specific trait types
 #-----------------------------------------------------------------------------
@@ -2598,7 +2616,7 @@ def test_override_default():
         a = Unicode('hard default')
         def _a_default(self):
             return 'default method'
-    
+
     C._a_default = lambda self: 'overridden'
     c = C()
     assert c.a == 'overridden'
@@ -2609,7 +2627,7 @@ def test_override_default_decorator():
         @default('a')
         def _a_default(self):
             return 'default method'
-    
+
     C._a_default = lambda self: 'overridden'
     c = C()
     assert c.a == 'overridden'
@@ -2620,8 +2638,7 @@ def test_override_default_instance():
         @default('a')
         def _a_default(self):
             return 'default method'
-    
+
     c = C()
     c._a_default = lambda self: 'overridden'
     assert c.a == 'overridden'
-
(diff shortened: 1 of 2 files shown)


traitlets/traitlets.py
```diff
@@ -1644,6 +1644,17 @@ def trait_events(cls, name=None):
...
         return events
 
+    def clear(self, **metadata):
+        """Remove the values of traits with certain metadata.
+
+        If not metadata is given, then all trait values are cleared.
+        """
+        for name in self.traits(**metadata):
+            try:
+                self._trait_values[name]
```
jankatins: Should this be `del self._trait_values[name]`?

rmorshea: Yes, thank you.

rmorshea: This also needs a test.

traitlets/traitlets.py
```diff
@@ -1644,6 +1644,21 @@ def trait_events(cls, name=None):
                         events[k] = v
         return events
 
+    def clear(self, *names, **metadata):
+        """Remove the values of listed traits or those with certain metadata.
+
+        If no metadata or names are given, then all trait values are cleared.
+        """
+        if not metadata and not names:
+            names = self.trait_names(**metadata) + list(names)
```
jankatins: Thus looks like it belongs to the case where both have values not to the case where both have no values

traitlets/tests/test_traitlets.py
```diff
@@ -857,6 +857,30 @@ def __init__(self, i):
...
+        a.clear(key="this")
+        assert a.i == 0
+        assert a.j == 0
+        assert a.k == 3
+
+        a = A(i=1, j=2, k=3)
+        a.clear("i", "j")
+        assert a.i == 0
+        assert a.j == 0
+        assert a.k == 3
```
jankatins: Misses a case for both

traitlets/traitlets.py
```diff
@@ -1644,6 +1644,21 @@ def trait_events(cls, name=None):
                         events[k] = v
         return events
 
+    def clear(self, *names, **metadata):
+        """Remove the values of listed traits or those with certain metadata.
+
+        If no metadata or names are given, then all trait values are cleared.
```
jankatins: Misses docs for what happens when both are given

rmorshea:
@jankatins, should be covered now. Was rushing to push this stuff for no good reason.

traitlets/traitlets.py
```diff
@@ -1644,6 +1644,21 @@ def trait_events(cls, name=None):
                         events[k] = v
         return events
 
+    def clear(self, *names, **metadata):
```
minrk: `clear` is much too generic a name to put on a base class like HasTraits. The namespace of methods is *extremely* precious. All trait-related methods should contain `trait`, and ideally start with `trait_`. Maybe `.trait_clear()`?

traitlets/traitlets.py
```diff
@@ -1644,6 +1644,21 @@ def trait_events(cls, name=None):
                         events[k] = v
         return events
 
+    def clear_traits(self, *names, **metadata):
```
minrk: I think it's better to start as many (probably all) trait-related methods with `trait_`. We've been much too lax adding to the namespace of user objects with new methods, and I think a pseudo-namespace of a `trait_` prefix lowers the cost to new inherited methods because they are grouped together.

Carreau:

Issue #463, opened 2018-01-05
Profiling and optimizations


vidartf:
I've recently been profiling initialization of widgets in ipywidgets, and discovered that most of the time was spent in traitlets. As such, I figured I would share the profiling results, and the results of some optimizations (caching).

### Profiling
Here's the content of a basic script file that I ran through cProfile (traitlets 4.3.2):
```python
from ipywidgets.widgets.widget import Widget

for i in range(20000):
    w = Widget()
```

(the number 20'000 was chosen as it gave a reasonable execution time of around ~25 seconds.)

Here's a summary of the profiling results:

Total time: 25.87

Number of `HasTraits` instances: 40'000 (20k Widget, 20k Comm instances)

Most of the time is spent in initialization (`HasDescriptors.setup_instance`) or querying `HasTraits.traits` (with and without metadata filter).

Initialization:

| ncalls  	| tottime	| percall 	| cumtime	| percall 	| file:lineno(function)
| --------- | --------- | --------- | --------- | --------- | --------- 
|40000   	| 1.487 	| 3.718e-05	| 10.27 	| 0.0002568	| traitlets.py:961(setup_instance)
|560000  	| 1.119 	| 1.997e-06	| 5.388 	| 9.622e-06	| traitlets.py:516(instance_init)
|260000  	| 0.2687	| 1.033e-06	| 2.706 	| 1.041e-05	| traitlets.py:1690(instance_init)
|680000  	| 1.797 	| 2.642e-06	| 2.786 	| 4.097e-06	| traitlets.py:486(_dynamic_default_callable)

Query:

| ncalls  	| tottime	| percall 	| cumtime	| percall 	| file:lineno(function)
| --------- | --------- | --------- | --------- | --------- | --------- 
|60000   	| 1.095 	| 1.826e-05	| 8.536 	| 0.0001423	| traitlets.py:1407(traits)
|60048   	| 1.783 	| 2.970e-05	| 5.566 	| 9.269e-05	| traitlets.py:224(getmembers)
|60000   	| 0.9321	| 1.554e-05	| 1.581 	| 2.635e-05	| traitlets.py:1421(< listcomp>)

Common:

| ncalls  	| tottime	| percall 	| cumtime	| percall 	| file:lineno(function)
| --------- | --------- | --------- | --------- | --------- | --------- 
|100059  	| 3.082 	| 3.081e-05	| 3.082 	| 3.081e-05	| <builtins.dir>
[...]

PR #469, opened 2018-02-03
Potential Optimizations

rmorshea:
Using the same simple test as #463 I was able to improve performance by ~30% in contrast to #463's ~40% boost. However this PR doesn't introduce caching, and instead registers descriptors via `subclass_init` to `HasDescriptors._descriptors`.  Ultimately this means that neither `HasTraits.traits`, nor `setup_instance` traverse all of `dir(cls)` to find traits and descriptors.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -407,7 +407,8 @@ class BaseDescriptor(object):
         self.name = name
 
     def subclass_init(self, cls):
-        pass
+        if self.name is not None:
+            cls._descriptors.append(self.name)
 
     def instance_init(self, obj):
         """Part of the initialization which may depend on the underlying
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -791,6 +792,7 @@ def setup_class(cls, classdict):
         BaseDescriptor in the class dict of the newly created ``cls`` before
         calling their :attr:`class_init` method.
         """
+        cls._descriptors = []
```
minrk: Does this new state need to go in the `__getstate__`?

jf---:
A 30% performance increase is quite the accomplishment... given the obvious upside, why not merge things?

traitlets/traitlets.py
rmorshea: @minrk if I'm not mistaken you don't have to recreate state for classes  when pickling so `_descriptors` and `_trait_default_generators` shouldn't need to be in `__getstate__`. When the module containing these classes is reloaded then these fields should be repopulated automatically.

rmorshea:
@jf--- responded to @minrk and awaiting his approval.

minrk:
@rmorshea sorry this repo is so neglected.

Can you add a test that covers pickle/unpickle and dynamically-defined instance traits with `add_traits`? That would make me more sanguine about moving a bunch of state to the class-only level.

Carreau:

PR #470, opened 2018-02-04
Facilitate adding new config-file loaders

ankostis:
## Rational:
If someone needs to introduce a new config-loader, 
(e.g. YAML), currently  has must re-implent the full `Application._load_config_files()`,
and change just a single line.

Similar to #360 for cmd-line argument loaders, 
this PR facilitates introducing new config-loaders without estensive changes in the code.

NOTE: this PR conceptually contans 2 "tips", 
one before the merge with @minrk 's #242 and one with it (in case this is merged as well)

## Before merge with #242 
To introduce a newconfig-loader, simply overridde `Application._make_loaders()` to return an orderd list of loaders.

## After merge with #242
- To extend with a new config-file, simply add its loader class in the
`supported_cfg_loaders: {.ext: loader-class}` ordered mapping.
- Deprecated `Application` class-attributes:
  - `python_config_loader_class`
  - `json_config_loader_class`
  replaced by the `supported_cfg_loaders` mapping:

## Bonus
A minor speedup is introduced when accesing only once the `HasTrairs.section_names()` outside of `_find_my_config()` loop.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -722,7 +722,7 @@ class Application(SingletonConfigurable):
             for ext, loader in cls.supported_cfg_loaders.items():
                 if name.endswith(ext):
                     return loader(name, path, log=log)
-            raise AssertionError("Unknown file-extension in config-file %r!" % name)
+            raise ValueError("Unknown file-extension in config-file %r!" % name)
 
         if not isinstance(path, list):
             path = [path]
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -600,7 +600,7 @@ class Application(SingletonConfigurable):
             # or ask factory to create it...
             self.subapp = subapp(self)
         else:
-            raise AssertionError("Invalid mappings for subcommand '%s'!" % subc)
+            raise ValueError("Invalid mappings for subcommand '%s'!" % subc)
 
         # ... and finally initialize subapp.
         self.subapp.initialize(argv)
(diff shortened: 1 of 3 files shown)


traitlets/config/application.py
```diff
@@ -711,18 +718,31 @@ def _load_config_files(cls, basefilename, path=None, log=None, raise_config_file
         yield each config object in turn.
         """
 
+        def new_loader(name, path):
+            for ext, loader in cls.supported_cfg_loaders.items():
+                if name.endswith(ext):
+                    return loader(name, path, log=log)
+            raise AssertionError("Unknown file-extension in config-file %r!" % name)
```
minrk: AssertionErrors shouldn't generally be raised. Probably best to raise KeyError or ValueError. It might also be nice to include the supported extensions in the error message.

minrk:
Since this includes #242, that PR will need to be resolved and this rebased after that before it can be merged.

ankostis:
What can i do now while waiting #242?

traitlets/config/application.py
ankostis: Ok, converted to `ValueError`.

I did the same fix on an another case using `AssertionError` class (missing sub-command).
That's where i copied the code from.  It turned out that it was my code in the first place :-)

minrk:
I think this PR is fine, but since it pulled in #242 which isn't done, it needs to wait until that PR finishes and then be rebased. If it hadn't pulled in that PR, I would be happy to merge it.

Carreau:

PR #477, opened 2018-03-01
Instance Inheritance

rmorshea:
# Inheritance From Instance

A better inheritance model for traits simplifies most validation logic.

# Examples

```python
class Bool(Instance):
    klass = bool
    default_value = False
```

```python
class Tuple(Collection):

    klass = tuple
    _cast_types = list
    default_value = ()
```

# Notes

This is #403 with improvements from #466.

# Todo

Improve this description.

Make sure this is completely separated from other work.


traitlets/traitlets.py
```diff
@@ -2386,296 +2362,311 @@ def info_rst(self):
...
-    def __init__(self, trait=None, default_value=None, **kwargs):
-        """Create a container trait type from a list, set, or tuple.
-
-        The default value is created by doing ``List(default_value)``,
-        which creates a copy of the ``default_value``.
+class Mutable(Instance):
 
-        ``trait`` can be specified, which restricts the type of elements
-        in the container to that TraitType.
+    def default(self, obj=None):
```
minrk: Mutable's not doing anything here. I think we can get rid of it to cutdown on the significant increase in inheritance complexity in this PR.

traitlets/traitlets.py
```diff
@@ -2386,296 +2362,311 @@ def info_rst(self):
...
+            return self.make_dynamic_default()
+        else:
+            return Undefined
 
-        If only one arg is given and it is not a Trait, it is taken as
-        ``default_value``:
 
-        ``c = List([1, 2, 3])``
+class Container(Instance):
+    """A base class for iterable types"""
```
minrk: More info in the docstrings would be helpful. How would I subclass Container, for instance.

traitlets/traitlets.py
```diff
@@ -2386,296 +2362,311 @@ def info_rst(self):
...
+    def validate_elements(self, obj, value):
+        return value
 
-        default_value : SequenceType [ optional ]
-            The default value for the Trait.  Must be list/tuple/set, and
-            will be cast to the container type.
 
-        allow_none : bool [ default False ]
-            Whether to allow the value to be None
+class Collection(Container):
```
minrk: We can drop the distinction between Collection and Container

traitlets/traitlets.py
```diff
@@ -2386,296 +2362,311 @@ def info_rst(self):
...
-        The default value is created by doing ``list(default_value)``,
-        which creates a copy of the ``default_value``.
-
-        ``trait`` can be specified, which restricts the type of elements
-        in the container to that TraitType.
-
-        If only one arg is given and it is not a Trait, it is taken as
-        ``default_value``:
-
-        ``c = List([1, 2, 3])``
```
minrk: All 'real' (user-facing) traits need to keep these detailed docstrings with examples. If anything, every PR should aim to *increase* the amount of docstring info we have, not decrease it.

rmorshea:
@minrk, it's definitely good that I split this off. I totally missed that docstrings had gone missing. I'll be sure to put them back, and hopefully add new documentation if I get time. Merging `Collection` into `Container` also seems reasonable.

Carreau:

PR #479, opened 2018-03-02
--show-config dumps configurations and then launches start() normally

ankostis:
## Problem
[edit: misstook a my local modifications as traitlets's codebase!]

I want to combine the  `--debug` flag when it's true, to also dump configurations 
(set `show_config` app.parameter to true).
Unfortunately, as currently implemented, the app halts!
And that is quite untypical for a `--debug` flag.

The reason is that `show_config[_json]` params (#341) is replacing the `Application.start()` method with a new one, that halts after dumping.
It is quite possible to dump the configuration AND continue normally with the application.

## Solution
This PR is simply calls `start()` after dumping configurations.

But in principle, config-dumping can be simplified even more by avoiding the replacement of `start()` in the first place.
We may just inspect the `show_config[_json]` params in `initialize()` function, like that:
[edit originally code had neglected sub-commands]
```python
    def initialize(self, argv=None):
        self.parse_command_line(argv)
        if self.subapp is None and (  # Only final sub-command dumps.
                self.show_config or self.show_config_json):
            self._dump_config()
```

But the `test_show_config[_json]()` test-cases in #341 have been written to bypass `initialize()` and call immediately `start()` after app construction.  

I would prefer such a simple configuration, unless there is a concrete reason behind that decision, agree?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ install_requires = setuptools_args['install_requires'] = [
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['pytest'],
+    'test': ['pytest', 'pytest-mock'],
     'test:python_version=="2.7"': ["mock"],
     # -- SUPPORT UNIFORM-WHEELS: Extra packages for Python 2.7
     # SEE: https://bitbucket.org/pypa/wheel/ , CHANGES.txt (v0.24.0)
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -293,12 +293,6 @@ class Application(SingletonConfigurable):
         help="Dump configuration to stdout while starting application (as JSON)"
     ).tag(config=True)
 
-    @observe('show_config', 'show_config_json')
-    def _show_config_changed(self, change):
-        if change.new and not hasattr(self, '_original_start'):
-            self._original_start = self.start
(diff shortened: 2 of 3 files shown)


minrk:
There is a concrete reason, because the goal of `--show-config` is to give an entrypoint to all applications to show configuration alone and exit immediately without taking any other action. Specifically, `start` should not run.

ankostis:
I have the feeling that similar functionality belongs to *mixins*, and not into base `Application` class,
so that downstream projects are allowed to choose what to use.

ankostis:
Travis is having some heisenberg problems with builds...in my account [all pass](https://travis-ci.org/ankostis/traitlets/builds/348335984).

minrk:
A mixin is okay, I suppose, as long as:

1. it's possible to have `--show-config` halt execution and prevent initialization and start
2. this is the default behavior

Flags on the base class are probably simpler and better than mixins, though.

ankostis:
Closing this PR because same functionality implemented by #480 
with 2 mixin classes, offering both cases (dump configs & exit app / launch normally).

> A mixin is okay, I suppose, as long as:
> ...
> 2. this is the default behavior

@minrk having the mixin "by default" activated (inherited) doesn't make sense; mixins cannot deactivate.

PR #480, opened 2018-03-17
Implement --show-config using mixin class for Application

ankostis:
Replaces #479 and simplifies dumping of configurations (a new feature not released yet in 5.0.x). 
Instead of monkeypatching `application.start()`, it uses plain inheritance and *mixins* classes.

- 2 mixin classes implemented in this PR:
  - `DumpConfig`: dump config and continue wth launchng application as normal.
  - `DumpConfigAndStop`: dump config and return from `start()`.
- Users can choose to opt-in any of the 2 functionalities by inheriting their app-classes from these mixins.
- Users may easily customize a new mixin classes to i.e. dump configs in a fie, instead of to `stdut`, simply by inheriting from them.
- Added a new section in docs to explain the 2 new mixin classes.

Open issue:
-------------
`Application.flags` may need a better way to enusre new flags are added if users want to modify them.

General note about Mixins on traitlets
=========================
It is not always possible to have all desired functionality built-in to classes.
Sometimes the requirements are contradictory;  other times, testing all functionality combinations is difficult.  

Mixins can be used for extra features that users can enable/disable in their sources, 
while testing becomes simpler because mixn-classes may specify in their documentation "incompatibilities" against other mixins, and avoid testing alltogether.

In my opinion, they should be used more often.

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -408,6 +408,15 @@ application, simply pass ``-h`` or ``--help``. And to see the full list of
 configurable options (*very* long), pass ``--help-all``.
 
 
+Mixins
+======
+- :class:`~traitlets.config.DumpConfig`: a :class:`~traitlets.config.Application`
+  mixin that dumps configurations and continue with launching the app as normal.
+- :class:`~traitlets.config.DumpConfig`: a :class:`~traitlets.config.Application`
+  mixin that dumps configurations, but stops app after dumping them
+  (just return from :meth:`~traitlets.config.Application.start()`).
+
+
 Design requirements
 ===================
 
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -818,7 +818,9 @@ def get_config():
 
 class _DumpConfigBase(six.with_metaclass(MetaHasTraits)):
     """
-    Baseclass for Application mixins that dumps configs on `--show-config(-json)` flags.
+    Abstract base for Application mixins that dump configs on `--show-config(-json)` flags.
+    
+    Subclasses need to override ``start_show_config()`` method.
     """
 
     show_config = Bool(
@@ -841,6 +843,16 @@ class _DumpConfigBase(six.with_metaclass(MetaHasTraits)):
             },
         }, "Show the application's configuration (json format)"),
     })
+
+    @observe('show_config_json')
+    def _show_config_json_changed(self, change):
+        self.show_config = change.new
+
+    @observe('show_config')
+    def _show_config_changed(self, change):
+        if change.new:
+            self._save_start = self.start
+            self.start = self.start_show_config
 
     def _dump_config(self):
         """start function used when show_config is True"""
@@ -886,22 +898,17 @@ class DumpConfigAndStop(_DumpConfigBase):
     """
     Application mixin that dumps configs & stops when `--show-config(-json)` given.
     """
-    def start(self):
-        if self.show_config or self.show_config_json:
-            self._dump_config()
-        else:
-            return super(DumpConfigAndStop, self).start()
+    def start_show_config(self):
+        self._dump_config()
 
 
 class DumpConfig(_DumpConfigBase):
     """
     Application mixin that dumpsconfigs when `--show-config(-json)` given.
     """
-    def start(self):
-        if self.show_config or self.show_config_json:
-            self._dump_config()
-        return super(DumpConfig, self).start()
-
+    def start_show_config(self):
+        self._dump_config()
+        return self._save_start()
 
 
 if __name__ == '__main__':
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -21,7 +21,9 @@ from traitlets.config.loader import (
     KVArgParseConfigLoader, PyFileConfigLoader, Config, ArgumentError, ConfigFileNotFound, JSONFileConfigLoader
 )
 from traitlets.traitlets import (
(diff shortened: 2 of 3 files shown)


ankostis:
Regarding the `flags`, to reuse any modified flags from baseclasses (and not just take always `Application.flags`), the mixn class can either do it in `__init__()` or in `setup_instance()`, but i don't know which one is the correct approach.

Below is the code for the 2nd approach:
```python
    def setup_instance(*args, **kwargs):  # @NoSelf  to allow "self" as kwarg
        self = args[0]
        self.flags = self.flags.copy()
        self.flags.update({
                'show-config': ({
                    'Application': {
                        'show_config': True,
                    },
                }, "Show the application's configuration (human-readable format)"),
                'show-config-json': ({
                    'Application': {
                        'show_config_json': True,
                    },
                }, "Show the application's configuration (json format)"),
            })
        super(_DumpConfigBase, self).setup_instance(*args{1:], **kwargs)
```

Any preference?

minrk:
> In my opinion, they should be used more often.

Having used mixins a fair amount and found lots of hard-to-follow complexity as a result, I personally think they should be used *less* often, and are not a good fit for enabling/disabling behavior when a single boolean flag will suffice. What functionality in this PR would not be solved by a single flag?

minrk:
This doesn't work:

```python
from traitlets.config import DumpConfigAndStop, Application

class BaseApp(DumpConfigAndStop, Application):
    def start(self):
        print("base")

class SubApp(BaseApp):
    def start(self):
        print("sub")

SubApp.launch_instance(['--show-config'])
```

launches the application instead of showing config.

The reason being that the mixin's `start` method is never used because it's overridden in the App classes that subclass it. The mixin approach requires all classes with a `start` method (i.e. ~all Applications) to be defined twice:

```python
class _MyApp(Application):
    def start(self):
        print("base")

class MyApp(DumpConfigAndStop, _MyApp):
    # cannot define start or DumpConfig will be ignored!
    pass
```

because the DumpConfig.start must occur in the mro than the 'true' start. Further, because the dump-config mixin must be exclusively on the terminal class, this requires specifying the DumpConfig mixin on *every* Application, it cannot be inherited, which is a deal-breaker for the primary goal of traitlets.config, which is to serve IPython/Jupyter applications. So the above Base/Sub has to become four classes, with the mixins reapplied at every level.

```python
class _MyApp(Application):
    def start(self):
        print("base")

class MyApp(DumpConfigAndStop, _MyApp):
    pass

class _SubApp(_MyApp):
    def start(self):
        print("sub")

class SubApp(DumpConfigAndStop, _SubApp):
    pass
```

And due to MRO resolution, subclassing the 'real' MyApp is impossible because mixing in `DumpConfig` twice results in an error.

```
class _SubApp(MyApp):
...
TypeError: Cannot create a consistent method resolution
order (MRO) for bases DumpConfigAndStop, _SubApp
```

ankostis:
You were right about the bug.
But the mixins were not the culprit.

You see, i was hopping i could get away with the clever monkey-patching of `start()` - but was not that clever.
My apologies.

The re-written PR restored monkeypatching `start()`.
And added your test-code as a TC, along with more good stuff.

ankostis:
I would appriceiate also any hint for how the mixins should populate the `flags` dictionary (https://github.com/ipython/traitlets/pull/480#issuecomment-373928004).

minrk:
So the main problem solved by this PR is that the new features is difficult to opt-out of. The solution is to make it opt-in (off by default). Making the opt-in mechanism a Mixin doesn't seem relevant to that choice, only that it changes how opt-in is done. Mixins tend to need to occur only one place in an inheritance tree. Over-specifying an option via a flag is fine, but doing so via a Mixin tends to result in errors.

If in my subclass chain DumpConfig occurred twice, I would have an error:

```python
from traitlets.config import DumpConfigAndStop, DumpConfig, Application

class BaseApp(DumpConfigAndStop, Application):
    def start(self):
        print("base")

class SubApp(DumpConfigAndStop, BaseApp):
    def start(self):
        print("sub")

SubApp.launch_instance(['--show-config', '--A.trait=5'])
```

```python-traceback
TypeError: Cannot create a consistent method resolution
order (MRO) for bases DumpConfigAndStop, BaseApp
```

Here is a use case:

I want to opt-in to `ShowConfigAndStop` in my base application class (I do want to do this, and it's why I think it belongs in `traitlets.config.Application`, but if it's not there, I will put it in the base Jupyter and IPython Applications). How do I opt-out for some subset of my SubApp subclasses? If it's a simple flag, I can do this easily on any class. How do opt-out of DumpConfig if my parent class opted-in via Mixin?

ankostis:
> If in my subclass chain DumpConfig occurred twice, I would have an error: ...

Inheritance bugs are easy to detect because they fail early, on development time.

> Here is a use case: ...

Regarding your case:
- the simple flag remains as is (it's the `show_config` trait), 
  and you may use it selectively, assuming that makes sense - you see, monkey-patched `start()` is invoked **only on the parent command**, not the sub-commands.
- The most flexible way would be to mix-in the `ShowConfigAndStop` at the `Jupyter` and `Ipython` subclass definitions.  You `SubApp` would not then be affected.  
  Is that too much of an effort to do, on development time?

> So the main problem solved by this PR is ...

No. Actually monkey-patching `start()` of the parent command is just one way to dump-configs.
Other times you may need to call dump-configs on the sub-command instance, not the parent-command.
Mixins allow for pluggable implementations, per command.

Carreau:

Issue #481, opened 2018-03-21
copy(HasTraits) creates a clone with shared state from the original


ankostis:
Actually the results of using `copy()` on a `Hastrait` with [borged](https://www.google.it/search?q=python+borg) trait values:

```python
from copy import copy

class C(HasTraits):
    a = Int()

c = C(a=1)
copy(c).a = 2
assert c.a == 1  # This FAILS because it is also 2!
```

## Suggested solution
I guess we need this line: 
```python
        d['_trait_values'] = self._trait_values.copy()
```
in `HasTraits.__getstate__()`:
https://github.com/ipython/traitlets/blob/3164ffb43167e9d1688046fbe0aefa424af5b58d/traitlets/traitlets.py#L1084-L1091

And i'm wondering whether the lock must no be shared, either:
```python
        d['_cross_validation_lock'] = False
```

ankostis:
Also related to #94.

rmorshea:
@ankostis I think you're right about setting `_cross_validation_lock` to `False`.

The only way `_cross_validation_lock` would be `True` is if the object were copied inside a `hold_trait_notifications` context, in which case the copied object would never have the lock reverted back to `False` (since that would only happen to the original). So yes, we definitely should not be sharing that.

PR #482, opened 2018-03-21
Fix copy(HasTraits) by copying `_trait_values` dict

ankostis:
- Added also TC.
- Waiting for response on question whether must reset `_cross_validation_lock' lock back to `False` 
  (asked also in #481).

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -2598,7 +2598,7 @@ def test_override_default():
         a = Unicode('hard default')
         def _a_default(self):
             return 'default method'
-    
+
     C._a_default = lambda self: 'overridden'
     c = C()
     assert c.a == 'overridden'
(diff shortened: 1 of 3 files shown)


ankostis:
All test-cases pass, except from Appveyor on Windows (expecting #483 to merge for that).

minrk:
Makes sense. I'm not sure what to do when cross_validation_lock is held. It does make sense to clear it, but what's the state going to be if a HasTraits is copied in the middle of pending changes? Should the cross-validation be forced to finish as part of the copy?

ankostis:
Ok, pinned the `_cross_validation_lock=False` when cloning.

> Should the cross-validation be forced to finish as part of the copy?

Maybe raise an exception, to fail early?

ankostis:
Btw, I have pushed a relevant commit about `utils.sentinel.Sentinel` ignoring copy/deepcopy.
Sentinels must not change, or cloned traits fail defaults 
(a a really nasty bug discovered a year ago, please don't ask me more).

minrk:
Thanks!

PR #487, opened 2018-04-30
make subapp instance inheritance conditional

### LINKED ISSUE #474: Subcommands interact badly with Singleton.instance(), blocking any use
## Problem
As currently implemented (till 5.x), subapps are actually constructed  in `ìnitialize_subcommands()` by invoking `SingletonConfigurable.instance()`.  And that makes sense, to make it easily to locate a command instance, regardless of its position in the "command chain".

But since this `SingletonConfigurable.instance()` class-method store the new instance to **all baseclasses**, this leads easily to clashes in case 2 subcommands share the same baseclass (which is always the case, since `Application` is a common parent).

Actually from within IPython you cannot even use any sub-command at all:
```python
>>> class App(Application):
...     pass
...
>>> class Sub(Application):
...     pass
...
>>> app = App(subcommands={'sub': (Sub, 'cmd title')})
>>> sub.initialize(['sub'])
...
MultipleInstanceError: An incompatible sibling of 'Sub' is already instanciated as singleton: TerminalIPythonApp
```

To signify the problem, here is a trivial example that fails even in pure python REPL:
```python
>>> class Sub1(Base):
...   pass
...
>>> class Sub2(Base):
...   pass
...
>>> app = Application(subcommands={'sub1':(Sub1, 'help'), 'sub2': (Sub2, 'help')})

>>> app.initialize(['sub1'])
>>> app.subapp
<__main__.Sub1 object at 0x000001C2D5D77C50>

>>> app.initialize(['sub2'])
...
traitlets.config.configurable.MultipleInstanceError: An incompatible sibling of 'Sub2' is already instanciated as singleton: Sub1
```
Anothe side-effect is that test-cases on sub-commands require spurious `clear_instances()` invocations to work (see `test_application.test_subcommands_instanciation()`).
## Possible Mitigations
Either:
1. Modify `initialize_subcommands()` not to use `Singleton.instance()`, or
2. modify `Singleton.instance()` and `clear_instances()` not to set/clear newly instanciated app to all `mro()` classe-attributes.

I would prefer (2), assuming there is not some blocker reason for why `instance()` behaves like that.

Of course it  might be that my understanding is totally broken here.

minrk:
only make subapp the global instance if the parent is the instance at the time

avoids triggering instance behavior when the parent isn't the instance (e.g. another Application is running and should stay in control)

closes #474

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ __pycache__
 .#*
 .coverage
 .cache
+.pytest_cache
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -587,10 +587,18 @@ class Application(SingletonConfigurable):
 
         ## Cannot issubclass() on a non-type (SOhttp://stackoverflow.com/questions/8692430)
         if isinstance(subapp, type) and issubclass(subapp, Application):
-            # Clear existing instances before...
-            self.__class__.clear_instance()
+            if Application.initialized() and Application.instance() is self:
+                # if we are the global Application instance,
+                # make the subapp the global instance
+                self.__class__.clear_instance()
+                make_subapp = subapp.instance
+            else:
+                # we aren't the global instance,
+                # don't make the subapp a global instance
+                # which could take over from another instance
+                make_subapp = subapp
             # instantiating subapp...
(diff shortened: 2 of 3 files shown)


Carreau:

PR #492, opened 2018-06-07
Convert dict to Config before merging to avoid overwriting extant values

mpacer:
Right now merging a dictionary into a Config object will overwrite the values of the Config object. If the dictionary had first been converted into a Config object, it would not overwrite the values. 

This converts arguments passed to merge to a `Config` object before beginning the merge. It appears to not affect existing `Config` objects that are being merged, but if there are any performance concerns, I can add a check to see if it's a dict before applying the transformation.

This also removes some unrelated, unnecessary prints from the tests that were making it hard to read the test output.

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -190,7 +190,7 @@ class Config(dict):
     def merge(self, other):
         """merge another config object into this one"""
         to_update = {}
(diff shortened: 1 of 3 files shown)


Carreau:

PR #498, opened 2018-08-12
adds string parsing ability to CBool

vlaurenzano:
This allows strings to be parsed in CBool. This mitigates unexpected scenarios in user code, for instance when using CBool with the value of 'False' which will return True. 

This is particularly useful when using env variables for traitlets since they will always come in as strings, as was the case in this project: https://github.com/jupyter/kernel_gateway/pull/299

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -2264,7 +2264,7 @@ class CBool(Bool):
 
     def validate(self, obj, value):
         try:
-            if isinstance(value, str):
+            if isinstance(value, six.string_types):
                 return strtobool(value)
             return bool(value)
         except:
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -41,12 +41,13 @@ Inheritance diagram:
 # also under the terms of the Modified BSD License.
 
 import contextlib
+import enum
 import inspect
 import os
 import re
 import sys
 import types
-import enum
+
 try:
     from types import ClassType, InstanceType
     ClassTypes = (ClassType, type)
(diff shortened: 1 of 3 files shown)


rmorshea:
@minrk will this change break anything? [strtobool](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) converts a fair number of strings to false - `n`, `no`, `f`, `false`, `off` and `0`. If not, this seems useful.

traitlets/traitlets.py
```diff
@@ -2262,6 +2264,8 @@ class CBool(Bool):
 
     def validate(self, obj, value):
         try:
+            if isinstance(value, str):
```
rmorshea: You should use `six.string_types` here, and add some byte strings to the `_good_values` test.

vlaurenzano: done

minrk:
I'm not sure. I wouldn't expect it to break anything.

#441 would also solve this issue in a more direct and explicit way, only coercing when loading from env, not on every `obj.bool_trait = 'false'` assignment, which I would generally expect and want to raise. I don't really think that string-casting belongs in the direct assignment of traits, but in general think that the CFoo traits should ~never be used, as the coercion is too relaxed and can result in surprising successful transformed assignments that should have been errors.

For getting a bool trait from env, this is the pattern I would use:

```python
enable = Bool(config=True)
@default('enable')
def _enable_default(self):
    return strtobool(os.environ.get("MY_ENABLE_VAR") or "0")
```

Since there has to be code already to get the config from the env to the trait, that's where I think the type coercion belongs, not on all assignments in general.

vlaurenzano:
@minrk  If you check the pr I linked to, that's what I ended up doing. https://github.com/jupyter/kernel_gateway/pull/299/files#diff-e12fea7b69535836a5f96101b1248316R184 

In general, I'd suggest either making coercion more robust or removing the CTypes, the middle ground is where it's most likely to cause problems.

minrk:
We can't remove them, since backward-compatibility is the number one concern of traitlets, but we can de-emphasize and possibly deprecate them. For now, let's call them "officially supported, but recommended against."

vlaurenzano:
Closed since people seem to not want it.

PR #503, opened 2018-10-15
Add a YAML config loader class 

basileMarchand:
I have add a simple traitlets.config.loader.YAMLFileConfigLoader to simply deal with Yaml file in place of json one. 
And I also add the associated test.

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -11,6 +11,7 @@ import os
 import re
 import sys
 import json
+import yaml
 from ast import literal_eval
 
 from ipython_genutils.path import filefind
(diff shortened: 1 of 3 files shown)


basileMarchand:
Hello everyone, 

I am a bit stupid, I thought that yaml is in the standard lib but in fact it's not the case. Do you authorize the use of non-standard module in traitlets or not ? If it is the case I will update the dependencies list and commit it in my branch. 

See you.

traitlets/config/loader.py
```diff
@@ -447,6 +448,52 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            self._find_file()
+        except IOError as e:
+            raise ConfigFileNotFound(str(e))
+        dct = self._read_file_as_dict()
+        self.config = self._convert_to_config(dct)
+        return self.config
+
+    def _read_file_as_dict(self):
+        with open(self.full_filename) as f:
+            return yaml.load(f)
```
Carreau: You most likely want to use `safe_load` instead of `load`, or this may become a security risk.

Carreau:
I think you should be able to provide that in its own package. There is a long history of issues with yaml config files, where loaders are inconsistent across versions an languages, plus a number of value types are "guessed" which leads to arbitrary bugs. Typically some number get parsed as dates. `country_code:fr` vs `country_code: no` (oh no the second one is boolean False instead of norway)... 

That usually makes is worry of using Yaml. 

Do you have any strong case of having that in the core ? If you have you own application you coudl publish a yaml_loader on PyPI, and if it gains traction we can add it to the core later on, once it's battle tested. 

Also if it gets into the core, we'll probably drop the Json loader as Yaml is a superset of Json.

Carreau:
If you've never published a package I recommend [flit](https://flit.readthedocs.io/en/latest/) which bring you from Zero-to published a few minutes.

Carreau:

PR #509, opened 2019-02-01
do not catch system exceptions like KeyboardInterrupt

albertz:
You never should use `except:` in general. In this case, you probably do not want to catch exceptions like `KeyboardInterrupt`.


minrk:
Good catch, thanks

PR #517, opened 2019-03-29
Use log from parent in LoggingConfigurable.

### LINKED ISSUE #514: Use Parent Log If Exists
I'd like to see the default `log` attribute of a `LoggingConfigurable` be the parent log, rather than the global traitlets log, if the parent is also a LoggingConfigurable.

```python
class LoggingConfigurable(...):
    ...

    log = Instance('logging.Logger')
    @default('log')
    def _log_default(self):
        # (vvv) new behavior (vvv)
        if isinstance(self.parent, LoggingConfigurable):
            return self.parent.log
        # (^^^) new behavior (^^^)
        from traitlets import log
        return log.get_logger()
```

If that sounds okay I can open a merge request with those two lines added.

travigd:
Closes #514.

Ping @minrk.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -435,6 +435,8 @@ class LoggingConfigurable(Configurable):
     log = Instance('logging.Logger')
     @default('log')
     def _log_default(self):
+        if isinstance(self.parent, LoggingConfigurable):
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #528, opened 2019-08-01
Fix formatting.

danielballan:
Make this a bulleted list in sphinx, as intended.


rmorshea:
Thanks for the PR!

danielballan:
I am pretty sure that there does not need to be space between the bullet points.

PR #529, opened 2019-08-09
Minor addition to docs 

krinsman:
Suggested minor addition to the docs to prevent other people from making the same mistake I made.

I know this is fairly obvious in hindsight, but better explicit than implicit I guess. I genuinely did spend around 30-60 minutes puzzled about how to fix this (since I had assumed incorrectly that the default behavior was that the value of `proposal.value` was accepted automatically as long as no errors were raised by the decorating function, but really functions with `@validate` decorators seem to be setters as much as anything else).

Again, my fault for not reading the original docs closely enough. I thought it might be helpful to explain the examples a little bit, is all.

### DIFF
--- a/docs/source/using_traitlets.rst
+++ b/docs/source/using_traitlets.rst
@@ -120,6 +120,13 @@ Basic Example: Validating the Parity of a Trait
         parity_check.value = 1
         parity_check.parity = 1
 
+Notice how all of the examples above return
+``proposal['value']``. This is necessary for validation to work
+properly, since the new value of the trait will be the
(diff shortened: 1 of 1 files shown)


krinsman:
As far as I can tell the reason why this fails one of the Travis tests doesn't relate to the changes I made, since I made no changes to the code, just to the docs, but the failed Travis test was because of one of the lines of code.

rmorshea:
Yeah, this isn't you. A downstream project is failing.

rmorshea:
BTW just added https://github.com/ipython/traitlets/issues/530

PR #531, opened 2019-08-13
Update myapp.py

connerxyz:
Provide an example of `shortname` usage, which is discussed in the comments but not demonstrated.


Carreau:
Thanks !

PR #535, opened 2019-10-14
Ensure default Set values are sorted for a reproducible build

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that traitlets generates non-reproducible output which is affecting the reproducibility of other packages. For example, in nbconvert:

```
  -<dd><p class="first">Default: {'image/jpeg',</span> <span class="pre">'image/svg+xml',</span> [..]
  +<dd><p class="first">Default: {'image/svg+xml',</span> <span class="pre">'application/pdf',</span> [..]
```

(From https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/nbconvert.html as of time-of writing.)

This is due to it not iterating over a `Set` traitlet type in a deterministic ordering when generating the `Default: [..]` human-readable strings. This patch ensures that we display the `Set` traits defaults in a sorted, and thus reproducible, manner.

### DIFF
--- a/traitlets/config/tests/test_configurable.py
+++ b/traitlets/config/tests/test_configurable.py
@@ -62,7 +62,9 @@ class Bar(Foo):
     b = Unicode('gotit', help="The string b.").tag(config=False)
     c = Float(help="The string c.").tag(config=True)
     bset = Set([]).tag(config=True, multiplicity='+')
+    bset_values = Set([2,1,5]).tag(config=True, multiplicity='+')
     bdict = Dict().tag(config=True, multiplicity='+')
+    bdict_values = Dict({1:'a','0':'b',5:'c'}).tag(config=True, multiplicity='+')
 
 foo_help=u"""Foo(Configurable) options
 -------------------------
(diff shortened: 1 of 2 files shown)


Carreau:
(rebased to update with master and re-kick the tests)

Carreau:
That actually make the default a a list instead of a set, which is technically wrong. That should be fix in the representation instead of the data structure itself, it looks like this can be done by definind "`default_value_repr`".

And yes having reproducible build would be good.

lamby:
Ah, hm, I can't remember if that was deliberate - when glancing at the code many months on I seem to recall that there was *something* that made me do it this way, but it really could have been that I had not seen this method. Hmm. :)

Carreau:
No problem, I haven't looked at traitlets in a long time and will need to dive again into the codebase to understand what have changed. 

THanks again for your patience, and sorry if it take some time to get things back in shape.

Carreau:
rebased and should have fixed the test. I tweaked the repr to be `set()`, for empty set and `{1,2,3}` for set, but ordered.

Carreau:
Aaaand it works.

PR #536, opened 2019-10-17
In hold_trait_notifications, don't assume `change` is a Bunch

gjoseph92:
If a change is submitted directly with `HasTraits.notify_change`, `_notify_observers` (which handles turning a `dict` to a `Bunch`) has been replaced with `hold`, which needs to be prepared for plain `dicts`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1164,6 +1164,9 @@ class HasTraits(six.with_metaclass(MetaHasTraits, HasDescriptors)):
                     return past_changes
 
             def hold(change):
+                if not isinstance(change, Bunch):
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -1164,6 +1164,9 @@ def compress(past_changes, change):
                     return past_changes
 
             def hold(change):
+                if not isinstance(change, Bunch):
+                # cast to bunch if given a dict
```
rmorshea, line 1168: Sorry to nitpick. Can you just tab in this comment?

noelportillo (approved):

PR #538, opened 2019-11-04
Add trait-names autocompletion support

martinRenou:
This PR takes over #515


martinRenou:
> Besides a version bump

What do you mean? Should I do something else like pinning funcsigs to a specific version?

rmorshea:
Just bumping the version of `traitlets`. I was planning on doing a release for this since it's small and won't impact downstream projects.

rmorshea:
Maybe I can do that in a different PR though.

PR #544, opened 2020-02-27
Should there be a `write_config_file` method in the config.Application class?

krinsman:
JupyterHub has a method like this, but nothing about the method as defined there seems very specific to JupyterHub:

https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L2438

It seems generally useful to me, which is why I merged in a more or less copy-pasted version of the same method into NBViewer:

https://github.com/jupyter/nbviewer/blob/master/nbviewer/app.py#L336

But copy-pasting large blocks of code like that makes me worry that something is wrong. Two distinct child classes having the same copy-pasted method seems to suggest to me that maybe the parent class should actually be the one which has the method?

I guess the most controversial part of this PR might be a default config file name as an attribute, which seems to be necessary for the `write_config_file` name method to work. But I'm not sure whether it matters, since that can more or less expect to be overwritten by every child class.

Anyway if I'm understanding the design purposes/goals of traitlets incorrectly, please let me know.

**EDIT:** This doesn't pass the last build, but it seems like most recent PR's haven't passed the last build, so it seems like the changes in this PR might not be what is causing the last build to fail.

**EDIT 2:** I think both JupyterLab and classic notebook also have functionality like this somehow, but I couldn't figure out from their codebases how it was implemented. So maybe the correct answer would actually be for JupyterHub and NBViewer to both do what Jupyter Lab/Notebook are doing, instead of adding a method to the `traitlets.config.Application` class. I honestly don't know.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -283,6 +283,8 @@ class Application(SingletonConfigurable):
         """
     )
 
+    config_file = Unicode('application_config.py', help="The config file to load").tag(config=True)
+
     _loaded_config_files = List()
 
     show_config = Bool(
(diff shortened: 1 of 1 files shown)


traitlets/config/application.py
```diff
@@ -826,6 +828,40 @@ def exit(self, exit_status=0):
         self.log.debug("Exiting application: %s" % self.name)
         sys.exit(exit_status)
 
+    def write_config_file(self):
```
rmorshea, line 831: @krinsman can we just make `config_file` a parameter of `Application.write_config_file()`? Subclasses can then pass in a `config_file` attribute if they have them.

krinsman: @rmorshea That makes sense to me. I don't know how the JupyterHub developers feel about it though, so I should refrain from speaking for them.

I probably don't have time in the near future to fix this PR, so if you come up with a better solution that makes this PR obsolete, please feel free to close it!

PR #546, opened 2020-03-19
Export Bunch

jasongrout:
Bunch is a very useful utility class that used to be exported, but then we locked down the things exported from traitlets.py.

Bunch is used in our public APIs, and it would be useful to be able to create one to manually call a change handler.


Carreau:
Bunch should relly just be [SimpleNamespace](https://docs.python.org/3/library/types.html#types.SimpleNamespace) which I beleive does the same, but +1 to reintegrate.

Carreau:
Failure unrelated, merging.

PR #558, opened 2020-05-30
Allow compounding of Lazy traitlets across configuration files.

Carreau:
This is mostly intesting in the case of systemwide configuration in
/etc/local... to provide default configuration to user and still allow
them to extend their configuration.

Current traitlets behavior is as follow.

   /etc/local/config.py ::  c.Foo.value = ['a']
   ~/.jupyter/config.py ::  c.Foo.value.append('b')

Final seen value will be `['b']`

With the changes included here, you would get `['a', 'b']`

It also become possible to do:

   /etc/local/config.py ::  c.Foo.value.append('a')
   ~/.jupyter/config.py ::  c.Foo.value.append('b')

In which case the end result will be `['default', 'a', 'b']`.

This feel like a more reasonable (and requested) behavior than the
current.

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -77,6 +77,7 @@ class LazyConfigValue(HasTraits):
     # list methods
     _extend = List()
     _prepend = List()
+    _inserts = List()
 
     def append(self, obj):
         self._extend.append(obj)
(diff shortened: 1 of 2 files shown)


Carreau:
@meeseeksdev suggestions

minrk:
This pattern:

```python
/etc/local/config.py :: c.Foo.value.append('a')
~/.jupyter/config.py :: c.Foo.value.append('b')
```

has always worked, I believe. I just verfied that it works with current stable, at least.

But 👍 to fixing the assignment-then-append case!

PR #569, opened 2020-06-15
Follow nep 29, remove support for Python < 3.7

Carreau:


Carreau:
Ok, dropping 3.6 for now it's easy to re-add.

PR #582, opened 2020-06-25
Add trait.from_string

minrk:
Re-issue of long-idle #441

The key to this PR is deferring the parsing of strings from the CLI to trait objects themselves, to avoid inconsistency when

This is attempting to resolve an inconsistency in the improved argparse handling for container traits, where a list trait can be specified in the argparse list way of giving the flag multiple times:

```
command -l x -l y
# or
command --Class.list_trait x --Class.list_trait y
```

which results in `Class.list_trait = ['x', 'y']`

The main issue with master is summarized [here](https://github.com/ipython/traitlets/pull/441#issuecomment-323811638):

>  There are three cases: `--short` aliases, `--Class.trait` in the classes list, and `--Class.trait` not in the classes list.

The new handling works for `--short` and when `Class` is in the classes list, but *not* when `Class` is not in the explicit Application.classes list (e.g. an IPython extension). This PR so far improves handling of scalars (solving the longstanding `ipython -c 1` bug), but container traits have been the sticky part.

TODO:

- [x] add TraitType.from_string
- [x] define from_string for simple scalars
- [x] use trait.from_string to evaluate CLI strings at config-loading time instead of CLI-parsing time
- [x] define from_string for container traits (`item_from_string`?)
- [x] make `--Class.trait x --Class.trait y` produce the same config regardless of `Application.classes`
- [x] test confirming above
- [ ] test some sensible Union cases

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -358,8 +358,9 @@ class DeferredConfigString(text_type):
     in the configurable classes.
 
     When config is loaded, `trait.from_string` will be used.
-    
-    If an error is raised
+
+    If an error is raised in `.from_string`,
+    the original string is returned.
 
     .. versionadded:: 5.0
     """
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -12,6 +12,7 @@ import sys
 import json
 import warnings
 from ast import literal_eval
+from collections import defaultdict
 
 from ..utils import cast_unicode
 
@@ -347,12 +348,20 @@ class Config(dict):
             raise AttributeError(e)
 
 
-class DeferredConfigString(str):
+class DeferredConfig:
+    """Class for deferred-evaluation of config from CLI"""
+    pass
+
+    def get_value(self, trait):
+        raise NotImplementedError("Implement in subclasses")
+
+
+class DeferredConfigString(str, DeferredConfig):
     """Config value for loading config from a string
 
     Interpretation is deferred until it is loaded into the trait.
 
-    Subclass of unicode for backward compatibility.
+    Subclass of str for backward compatibility.
 
     This class is only used for values that are not listed
     in the configurable classes.
@@ -377,6 +386,46 @@ class DeferredConfigString(str):
 
     def __repr__(self):
         super_repr = super(DeferredConfigString, self).__repr__()
+        return '%s(%s)' % (self.__class__.__name__, super_repr)
+
+
+class DeferredConfigList(list, DeferredConfig):
+    """Config value for loading config from a list of strings
+
+    Interpretation is deferred until it is loaded into the trait.
+
+    This class is only used for values that are not listed
+    in the configurable classes.
+
+    When config is loaded, `trait.from_string_list` will be used.
+
+    If an error is raised in `.from_string_list`,
+    the original string list is returned.
+
+    .. versionadded:: 5.0
+    """
+    def get_value(self, trait):
+        """Get the value stored in this string"""
+        if hasattr(trait, "from_string_list"):
+            src = list(self)
+            cast = trait.from_string_list
+        else:
+            # only allow one item
+            if len(self) > 1:
+                raise ValueError("Multiply defined option: %s" % trait.name)
+            src = self[0]
+            cast = trait.from_string
+
+        try:
+            return cast(src)
+        except Exception:
+            # exception casting from string,
+            # let the original value lie.
+            # this will raise a more informative error when config is loaded.
+            return src
+
+    def __repr__(self):
+        super_repr = super(DeferredConfigList, self).__repr__()
         return '%s(%s)' % (self.__class__.__name__, super_repr)
 
 
@@ -408,7 +457,7 @@ class ConfigLoader(object):
         """A base class for config loaders.
 
         log : instance of :class:`logging.Logger` to use.
-              By default loger of :meth:`traitlets.config.application.Application.instance()`
+              By default logger of :meth:`traitlets.config.application.Application.instance()`
               will be used
 
         Examples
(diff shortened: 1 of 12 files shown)


minrk:
@Carreau I pushed a commit that makes handling containers more consistent. Tests need to be updated and several more cases covered, but it should be mostly there.

The main sketchy thing I did to accomplish this was to load a private-api-patched ArgumentParser to enable defining `--Class.trait` options on-demand. The goal here was to ensure consistent handling of `--Foo.bar=x`, `--Foo.bar x`, etc. If you have a better solution to that part, I'd be happy to work through it.

Another question to resolve is where deprecated literal-eval handling belongs (`--string "'1'"` and `--dict '{"key": "value"}'`). The easiest way to make sure it's always called is to put it in the new `from_string` itself, but it logically makes the most sense to only apply it when coming from CLI.

Carreau:
I think that the changes of behavior of Cli parsing between 4.x and 5.0 is fair game especially if it's in corner case. It seem fair to say that too complex of command line may not work well. 

It's also common to have regression in major releases that are fixed in the next patched version. 

I'll try to find some cycles to write test and understand what you wrote.

traitlets/config/loader.py
```diff
@@ -709,71 +834,43 @@ def load_config(self, argv=None, aliases=None, flags=None):
...
         if argv is None:
             argv = self.argv
-        if aliases is None:
-            aliases = self.aliases
-        if flags is None:
-            flags = self.flags
+        if aliases:
+            raise ValueError("%s does not support aliases" % self.__class__.__name__)
+        if flags:
+            raise ValueError("%s does not support flags" % self.__class__.__name__)
```
Carreau, line 844: I get from this that this is on purpose; so things like `--Foo.Bar a=1 -F b=2` where foo is an alias of `--Foo.Bar` is not handled ? Or does that just mean it's in progress ?

Carreau:
@minrk can you look at https://github.com/minrk/traitlets/pull/1  (A PR against your PR, likely easier to review).

traitlets/config/loader.py
minrk: This actually doesn't affect command-line parsing in 'real' applications, which use `KVArgParseConfigLoader` below. `KVArgParseConfig` works like this:

1. build an argparse Parser to handle aliases and flags
2. call `parse_known_args` to get alias and flag config,
3. construct one of these `KeyValueConfigLoader`s to handle parsing the remaining *unrecognized* arguments, which should all be arbitrary `--Class.trait=foo` (or unrecognized aliases that should raise).

The result is that these `KeyValueConfigLoader` objects are *only* used to parse full `--Class.trait` args, never aliases or flags. In changing the implementation here to use argparse (mainly to get consistent multiply-specified and `=` vs space handling), I lost the previous support for aliases, which was never used, since these were by definition used only on non-alias/flag options. Rather than complicate things further by reimplementing alias support where it will never be used, I opted to make the lack of support in thus sub-parser explicit.

I'll have a look at docstrings and comments to try to make this clearer.

Carreau:
Force-pushig from f6f195b after rebase/conflicts.

Carreau:
ok, the last "broken" thing is `ipython --profile test test.py`, as the KVparse collect both `test` and `test.py` in a list for `--profile.` 

The workaround is to use either `--profile=test` or `--profile test -- test.py`; thoughts on that ?

@consideRatio, @yuvipanda ; hany tough on how a similar thing may affect JupyterHub ?
@MSeal, @maartenbreddels, @SylvainCorlay same things for nbconvert

MSeal:
`--profile test -- test.py` would imply to me that you're separating to entirely new sub-application argument with `--` by itself. Thus `--profile=test` would be what I would expect to assign in isolation.

minrk:
Yeah, I don't think we want `-f x y` to be collecting into a single config. Will need to look into why that's happening.

Carreau:
> Yeah, I don't think we want `-f x y` to be collecting into a single config. Will need to look into why that's happening.

because the KVParser when it encounter something which is unknown create a parser with `nargs="+"`, to collect things like mapping; for example `--foo a=1 b=2`. 

There are explicit test case for this new behavior.

Carreau:
The kv parser is create only of `=` not in `--whatever`, hence why `--profile=test other` work but not `--profile test other`. I looked into setting `nargs=1`, but that of course break the case with mapping; and I'm not sure we can stop the KVparser to parse argument to `--stuff a=b c=d` when no `=` is found in a item. Note that this of course only affects application only if those unknown options are close to the end.

Carreau:
So I've looked into argparse; `consume_optional` has the followign block:

```
start = start_index + 1
selected_patterns = arg_strings_pattern[start:]
arg_count = match_argument(action, selected_patterns)
stop = start + arg_count
args = arg_strings[start:stop]
action_tuples.append((action, args, option_string))
```

Where `args` end up being `['test','test.py']`; there is no way without subclassing and rewriting ~80 lines to let argparse stop parsing eagerly, as the KVArgParser is nowhere given the current list of things it may consume.

SO I'm not sure can support both varargs, and `--Containers a b c` on the CLI.

traitlets/config/loader.py
```diff
@@ -618,7 +684,11 @@ def _load_flag(self, cfg):
 # rejects:  -foo=value
 #           --foo
 #           --Class.trait
-kv_pattern = re.compile(r'\-\-[A-Za-z][\w\-]*(\.[\w\-]+)*\=.*')
+kv_pattern = re.compile(r'\-\-[A-Za-z][\w\-]*(\.[\w\-]+)+\=.*')
+
+kv_key_pattern = re.compile(r'[A-Za-z]+(\.[\w]+)+')
+
+kv_opt_pattern = re.compile(r'\-\-[A-Za-z][\w]*(\.[\w]+)+')
```
Carreau, line 691: unused now.

traitlets/config/loader.py
```diff
@@ -628,15 +698,76 @@ def _load_flag(self, cfg):
...
+
+    but acts as if all --Class.trait options are predefined
+    """
+
+    def _add_kv_action(self, key):
+        self[key] = _KVAction(
+            option_strings=[key],
+            dest=key.lstrip("-").replace(".", _DOT_REPLACEMENT),
+            nargs="+",
+        )
```
Carreau, line 728: So this is the culprit; and I think this approach is 100% incompatible with trailing varargs.

Carreau:
hugh, I also get `MultipleValueError hist_file` when running a custom nbconvert script... I'll try to dig into that.

I'm starting to feel less confident about traitlets 5....

Carreau:
Dask and conda-forge logo in particular seem to be really low res.

traitlets/config/loader.py
minrk: I thought we needed `--foo` to be specified multiple times, i.e. `--foo a=1 --foo b=2`. That was the goal, at least. I agree and fear it's going to be too ambiguous if that's not the case. I'll look into the variable-arguments docs in argparse to see if we can exert more control over this.

minrk:
I fixed the nargs issue so that you have to pass a given flag for each item, not supporting `-X a b c` all appending to X. I also simplified things a bunch by merging the argparse-unaware KeyValue parser logic into KVArgParse (the only one that's ever directly used), eliminating the two-pass problem, which should fix any ordering mixups with e.g. `--alias value --Class.trait value`. The only change I really see now, which I don't think is bad, is that we can no longer inter-mix flags in positional arguments. Positional arguments must be contiguous (or after `--`), which I think results in a better, simpler CLI.

Carreau:
> The only change I really see now, which I don't think is bad, is that we can no longer inter-mix flags in positional arguments. Positional arguments must be contiguous (or after `--`), which I think results in a better, simpler CLI.

I agree with that. Let's get it in and polish a beta1.

minrk:
Were you able to figure out the cause of the `MultipleValueError`? It would probably be prudent to get everything we find that fails, even if we fixed it eventually, into a test case so it's easier to be confident in our test coverage of realistic cases.

I fear that our command-line config coverage for real apps isn't great because most downstream packages don't extensively test their CLI.

Carreau:
Ah, no I forgot about that.

Carreau:
MultipleValueError  seem to be gone on traitlets master, I have other errors, but that's due to me using private nbconvert API I think.

kernc:
> MultipleValueError seem to be gone on traitlets master

Reoccured in https://github.com/ipython/traitlets/issues/609.

PR #584, opened 2020-06-30
nicer falsy behavior for LazyConfigValue

minrk:
- empty lazy config is falsy, non-empty is truthy
- checking `in` for empty lazy config returns False, indicating there's no config there

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -61,9 +61,11 @@ class ArgumentParser(argparse.ArgumentParser):
 # Config class for holding config information
 #-----------------------------------------------------------------------------
 
+
 def execfile(fname, glob):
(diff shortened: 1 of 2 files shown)


Carreau:
I'm not sure if we should fix the tests, or the `__contains__`. I can't see why we would like `'foo' in c.foo` to be true-ish.

minrk:
Sorry or leaving this forever. The failing tests were mainly verifying that the LazyConfig was stored so that future access would retrieve the same thing, but the asserts were broken by the new falsiness. So I updated the tests to verify presence in another way.

noelportillo (approved):

Carreau:
I'm going to close for innactivity.

PR #597, opened 2020-08-17
Update metadata in setup.py

### LINKED ISSUE #595: update the long description in setup.py

Carreau:
closes #595

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,35 +42,17 @@ with open(pjoin(here, name, '_version.py')) as f:
     exec(f.read(), {}, version_ns)
 
 
-long_desc = """
-Traitlets is a pure Python library enabling:
+with open(os.path.join(here, "README.md")) as f:
+    long_description = f.read()
 
-- the enforcement of strong typing for attributes of Python objects
-  (typed attributes are called *"traits"*);
-- dynamically calculated default values;
-- automatic validation and coercion of trait attributes when attempting a
-  change;
-- registering for receiving notifications when trait values change;
-- reading configuring values from files or from command line
-  arguments - a distinct layer on top of traitlets, so you may use
-  traitlets without the configuration machinery.
-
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -42,16 +42,44 @@
     exec(f.read(), {}, version_ns)
 
 
+long_desc = """
```
rmorshea, line 45: Can we just read in the `README.md` and use that as the `long_desc`?

```python
with open(os.path.join(here, "README.md")) as f:
    long_description = f.read()

long_description = long_description
long_description_content_type = "text/markdown"
```

Carreau: We could, but there are some github specific thing in the readme. Not sure it matters much.

setup.py
```diff
@@ -42,16 +42,44 @@
...
+
+Its implementation relies on the [descriptor](https://docs.python.org/howto/descriptor.html)
+pattern, and it is a lightweight pure-python alternative of the
+[*traits* library](http://code.enthought.com/pages/traits.html).
+
+Traitlets powers the configuration system of IPython and Jupyter
+and the declarative API of IPython interactive widgets.
+"""
+
+
```
Carreau, line 66:
```suggestion
with open(os.path.join(here, "README.md")) as f:
    long_description = f.read()

long_description_content_type = "text/markdown"
```

setup.py
```diff
@@ -42,16 +42,44 @@
...
 setup_args = dict(
     name            = name,
     version         = version_ns['__version__'],
     scripts         = glob(pjoin('scripts', '*')),
     packages        = packages,
-    description     = "Traitlets Python config system",
-    long_description= "A configuration system for Python applications.",
+    description     = "Traitlets Python configuration system",
+    long_description= long_desc,
```
Carreau, line 73:
```suggestion
    long_description= long_description,
```

PR #598, opened 2020-08-18
Remove u string prefixes

rahulporuri:
This PR removes the `u` prefixes on strings. The prefixes dont do anything on Python 3 and they are unnecessary noise in the code. 

Note that this PR is one PR in a series of PRs that aim to remove Python 2 support from the code base. This PR has been standalone given the volume of changes and to aid in the review process.

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -1229,7 +1229,7 @@ class AnyTraitTest(TraitTestBase):
     obj = AnyTrait()
 
     _default_value = None
-    _good_values   = [10.0, 'ten', u'ten', [10], {'ten': 10},(10,), None, 1j]
+    _good_values   = [10.0, 'ten', [10], {'ten': 10},(10,), None, 1j]
     _bad_values    = []
 
 class UnionTrait(HasTraits):
@@ -1271,9 +1271,9 @@ class TestInt(TraitTestBase):
     obj = IntTrait()
     _default_value = 99
     _good_values   = [10, -10]
-    _bad_values    = ['ten', u'ten', [10], {'ten': 10}, (10,), None, 1j,
-                      10.1, -10.1, '10L', '-10L', '10.1', '-10.1', u'10L',
-                      u'-10L', u'10.1', u'-10.1',  '10', '-10', u'10', -200]
+    _bad_values    = ['ten', [10], {'ten': 10}, (10,), None, 1j,
+                      10.1, -10.1, '10L', '-10L', '10.1', '-10.1',
+                      '10', '-10', -200]
 
 
 class CIntTrait(HasTraits):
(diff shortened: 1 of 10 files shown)


Carreau:
Thanks ! I'll have a look at try to get this in the next release ; We usually tend to avoid large scale PR that change all the file as it breaks `git blame`.

If this is not the only project you do that one; these changes can be automatically applied via https://github.com/asottile/pyupgrade

I'm also working on improving [darker](https://github.com/akaihola/darker/pull/51) to apply upgrade only on changed lines which would have the advantage to progressively update the codebase w/o breaking git blame.

traitlets/tests/test_traitlets.py
```diff
@@ -1229,7 +1229,7 @@ class AnyTraitTest(TraitTestBase):
     obj = AnyTrait()
 
     _default_value = None
-    _good_values   = [10.0, 'ten', u'ten', [10], {'ten': 10},(10,), None, 1j]
+    _good_values   = [10.0, 'ten', [10], {'ten': 10},(10,), None, 1j]
```
Carreau, line 1232: 👍 nice catch, pyupgrade would not find this !

traitlets/tests/test_traitlets.py
```diff
@@ -1495,10 +1493,10 @@ class TestDottedObjectName(TraitTestBase):
     obj = DottedObjectNameTrait()
 
     _default_value = "a.b"
-    _good_values = ["A", "y.t", "y765.__repr__", "os.path.join", u"os.path.join"]
-    _bad_values = [1, u"abc.€", "_.@", ".", ".abc", "abc.", ".abc.", None]
+    _good_values = ["A", "y.t", "y765.__repr__", "os.path.join", "os.path.join"]
```
Carreau, line 1496: double os.path.join now :-)

Carreau:
I have to run into a meeting but that looks good to me !

rahulporuri:
> Thanks ! I'll have a look at try to get this in the next release ; We usually tend to avoid large scale PR that change all the file as it breaks git blame.
>
> If this is not the only project you do that one; these changes can be automatically applied via https://github.com/asottile/pyupgrade
>
> I'm also working on improving darker to apply upgrade only on changed lines which would have the advantage to progressively update the codebase w/o breaking git blame.

I've seen different strategies on different projects - some which prefer to make all of the stylistic upgrades in a single PR - squash merged into a single commit. That way, history is only broken in one commit. I can also understand/appreciate making modular changes only in parts of code/files in project which are modified in various PRs. I guess the downside with that approach is that some parts of the codebase are more readable than others - especially if the project as a whole moves slower in general.

I leave it up to y'all to decide what the strategy for traitlets is.

traitlets/tests/test_traitlets.py
rahulporuri: keen eye. i was afraid there'd be a duplicate i would miss.

Carreau:
for new contribution we'll like not reject them, but I've been bitten enough by large changes and it breaking git-bisect that I'm usually on the fence.

Git blame now have the _option_ to ignore some commit, but it's not avail via github... 

Anyway thank, this is a great first contribution, it will make the 5.x codebase cleaner !

rahulporuri:
thanks @Carreau

PR #613, opened 2020-09-03
handle allow_none in from_string

### LINKED ISSUE #612: Passing None on the command line
cc @SylvainCorlay 

Hi all,

congrats with the 5.0 release! But with all big releases come bugs :)

In voila we used to pass to a Unicode config a None value by using
```
$ voila --template=None
```
With traitlets 5, this becomes the string 'None'. Any falsy would work for us, but it also does not seem possible to pass an empty string.

Any ideas for workaround, or would this require a fix?

cheers,

Maarten

Related voila PR: https://github.com/voila-dashboards/voila/pull/691

minrk:
allow_none wasn't taken into account in `from_string`, meaning `--alias=None` didn't work as intended.

closes #612

### DIFF
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -295,6 +295,15 @@ class TestApplication(TestCase):
             assert app.config.TestApp.value == 'cli'
             assert app.value == 'cli'
 
+    def test_cli_allow_none(self):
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Thanks for fixing this so fast @minrk !

PR #615, opened 2020-09-04
handle deprecated eval syntax for Bytes on the command-line

### LINKED ISSUE #608: 5.0.0 fails to parse args
Hi, 
Looks like you are about to release `5.0.0` but it already gets fetched by `jupyter` :
```shell
docker run --rm -it python:3.7 bash
$ pip install jupyter
...
Collecting traitlets>=4.2
  Downloading traitlets-5.0.0-py3-none-any.whl (97 kB)
...
```
And leads to an exception in my tests, you may try to reproduce with:
```
$ python
>> from nbconvert import nbconvertapp
>> nbconvertapp.main(["nbconvert", "--output", "/tmp/qwe.txt", "simple-py36.ipynb"])
...
: error: unrecognized arguments: simple-py36.ipynb
...
```
I have fixed it by pinning to `traitlets>=4.2,<5.0.0`.

minrk:
closes ipython/ipykernel#543
closes #608

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -2690,6 +2690,22 @@ def test_cunicode_from_string(s, expected):
 
 @pytest.mark.parametrize(
     "s, expected",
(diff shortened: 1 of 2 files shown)


Zsailer (approved):
Thanks, @minrk. LGTM

PR #624, opened 2020-09-18
allow from_string to be specified via metadata

minrk:
avoids need to define custom Trait subclasses just to define how strings are parsed, e.g.

```python
class C(Configurable):
    path = List().tag(config=True, from_string=os.pathsep.split)
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -720,12 +720,29 @@ class TraitType(BaseDescriptor):
 
         >>> Int(0).tag(config=True, sync=True)
         """
-        maybe_constructor_keywords = set(metadata.keys()).intersection({'help','allow_none', 'read_only', 'default_value'})
+        maybe_constructor_keywords = set(metadata.keys()).intersection(
+            {"help", "allow_none", "read_only", "default_value"}
+        )
         if maybe_constructor_keywords:
             warn('The following attributes are set in using `tag`, but seem to be constructor keywords arguments: %s '%
                     maybe_constructor_keywords, UserWarning, stacklevel=2)
 
         self.metadata.update(metadata)
+
+        # allow from_string to be overridden via metadata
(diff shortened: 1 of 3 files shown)


traitlets/traitlets.py
```diff
@@ -720,12 +720,29 @@ def tag(self, **metadata):
...
+        )
         if maybe_constructor_keywords:
             warn('The following attributes are set in using `tag`, but seem to be constructor keywords arguments: %s '%
                     maybe_constructor_keywords, UserWarning, stacklevel=2)
 
         self.metadata.update(metadata)
+
+        # allow from_string to be overridden via metadata
+        if self.metadata.get("from_string"):
+            self.from_string = self.metadata["from_string"]
```
rmorshea, line 734: Why not look this up in the `metadata` dict rather than assign to an attr and find it there?

rmorshea (approved):

traitlets/traitlets.py
rmorshea: Nevermind, I guess `from_string_list` works the same way.

minrk: Yeah, it's the TraitType subclass priority I was thinking of. The alternative was to implement `_from_string` private method that does the lookup, but I thought that might get confusing.

minrk:
I was about to merge this when I realized the argument should probably be a Bunch with access to the trait instance itself like we do with all other hooks, to grant access to other tags, default values, etc.

minrk:
Hm, but that makes it not have the same signature as the method. Maybe I should make the signature `(trait, string)` so it's the *same* as the method? I'm not sure.

rmorshea:
I think making it `(trait, string)` makes sense

noelportillo (approved):

Carreau:
I'm going to close for innactivity

Issue #626, opened 2020-09-28
Tuple cannot have default_value=None


martinRenou:
When creating an `HasTraits` class with a tuple value that can be none, we get an error:

```python
from traitlets import HasTraits, Tuple

class Foo(HasTraits):
    value = Tuple(allow_none=True, default_value=None)
```

Shows:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-54c571552fdc> in <module>
----> 1 class Foo(HasTraits):
      2     value = Tuple(allow_none=True, default_value=None)
      3 

<ipython-input-2-54c571552fdc> in Foo()
      1 class Foo(HasTraits):
----> 2     value = Tuple(allow_none=True, default_value=None)
      3 

~/lib/python3.8/site-packages/traitlets/traitlets.py in __init__(self, *traits, **kwargs)
   2414             args = (default_value,)
   2415         else:
-> 2416             raise TypeError('default value of %s was %s' %(self.__class__.__name__, default_value))
   2417 
   2418         self._traits = []

TypeError: default value of Tuple was None
```

martinRenou:
A workaround for me is to wrap it around an Union:

```python
from traitlets import HasTraits, Tuple, Union

class Foo(HasTraits):
    value = Union((Tuple(), ), allow_none=True, default_value=None)
```

minrk:
should be fixed by #630

minrk:
(updated the title to specify that it's `default_value=None` that doesn't work; not `allow_none=True` which works for assignment anywhere *but* the default value in the constructor: So another workaround might be to use a dynamic default:

```python
from traitlets import HasTraits, Tuple, Integer, default

class C(HasTraits):
    s = Tuple((Integer(), Integer()), allow_none=True)
    @default("s")
    def _s_default(self):
        # workaround lack of support for default_value=None in constructor in traitlets <= 5.0.4
        return None

print(C().s)
```

PR #627, opened 2020-09-29
fixes: #626

rmorshea:
allow_none=True and default_value=None do not work in Tuple


rmorshea:
Nope. Not quite that simple.

PR #633, opened 2020-10-14
Fix error configuring nested tuple

zazezizozu:
Adding configuration to a nested tuple in IPython and running `%config RelevantClassName` caused `TypeError: not all arguments converted during string formatting`. This wraps any argument in a one element tuple and prevents the error.


minrk:
Thanks!

PR #635, opened 2020-10-15
stop recommending CUnicode for redundant-quotes

minrk:
it's not actually a valid solution to the problem, and not a good recommendation

cf jupyter/jupyter_server#306


rmorshea (approved):

willingc (approved):
LGTM

PR #638, opened 2020-11-17
speed up setting attributes

maartenbreddels:
Setting an attribute on a trait is slow, because we always trigger callbacks, even when they are not there. This speeds it up by 3x.

Before:
```
------------------------------------------------- benchmark: 1 tests ------------------------------------------------
Name (time in us)           Min      Max    Mean  StdDev  Median     IQR   Outliers  OPS (Kops/s)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_performance_set     5.5110  94.8380  6.9445  2.8176  6.0430  0.2957  3040;3721      143.9978   21575           1
---------------------------------------------------------------------------------------------------------------------
```

After:
```
------------------------------------------------- benchmark: 1 tests ------------------------------------------------
Name (time in us)           Min      Max    Mean  StdDev  Median     IQR   Outliers  OPS (Kops/s)  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_performance_set     1.8560  89.3450  2.2384  1.3132  2.0570  0.0880  2200;2725      446.7546   37368           1
---------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -617,9 +617,9 @@ class TraitType(BaseDescriptor):
             proposal = Bunch({'trait': self, 'value': value, 'owner': obj})
             value = obj._trait_validators[self.name](obj, proposal)
         else:
-            if self.name not in obj._magic_trait_notifiers_found:
-                obj._magic_trait_notifiers_found[self.name] = hasattr(obj, '_%s_validate' % self.name)
-            if obj._magic_trait_notifiers_found[self.name]:
+            if self.name not in obj._magic_trait_validator_found:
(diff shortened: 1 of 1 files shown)


rmorshea:
It looks like you have to go to a lot of effort to look up the "magic" notifier methods.

Probably fine to support them in this PR, but maybe we should think about officially deprecate those? We've had the deprecation message 1+ years, so no one should be surprised if we finally dropped support. Thoughts @minrk.

noelportillo (approved):

Carreau:
closing as stale; If someone want to pick it up; maybe we can get better perf

PR #639, opened 2020-11-19
perf: fast path for simple default construction

maartenbreddels:
Benchmark results will come in a referring PR

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -535,7 +535,11 @@ class TraitType(BaseDescriptor):
             value = obj._trait_values[self.name]
         except KeyError:
             # Check for a dynamic initializer.
-            default = obj.trait_defaults(self.name)
+            key = (obj.__class__, self.name)
+            if key in obj._simple_default:
+                default = obj._simple_default[key](obj)
+            else:
+                default = obj.trait_defaults(self.name)
             if default is Undefined:
                 warn(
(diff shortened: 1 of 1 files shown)


rmorshea:
Given that `_trait_default_generators` is a private attribute (and a relatively recently added on at that) could we just change it to use the `(class, name)` key you're using in `_simple_default`? Default value creation is already complicated enough as it is, so it'd be nice if we could avoid the adding a separate cache layer if we can just fold that into an existing abstraction.

If that's not possible though I'd definitely suggest a name like `_trait_default_generator_cache` since that's a bit more descriptive of what the attribute does. Also, we can probably have the trait continue to use `obj.trait_defaults`, but have that method inspect the cache before tracing the MRO to find the default generator.

rmorshea (changes requested):
Given that `_trait_default_generators` is a private attribute (and a relatively recently added on at that) could we just change it to use the `(class, name)` key you're using in `_simple_default`? Default value creation is already complicated enough as it is, so it'd be nice if we could avoid the adding a separate cache layer if we can just fold that into an existing abstraction.

If that's not possible though I'd definitely suggest a name like `_trait_default_generator_cache` since that's a bit more descriptive of what the attribute does. Also, we can probably have the trait continue to use `obj.trait_defaults`, but have that method inspect the cache before tracing the MRO to find the default generator.

maartenbreddels:
Replaced by #777

PR #642, opened 2021-01-30
Add ability to set a trait link directly

### LINKED ISSUE #641: Support setting a link on traits
I'm looking to start using a standard pattern in our UIs to split our model and views, similar to the points discussed in https://github.com/jupyter-widgets/ipywidgets/issues/2296. However when extending to more complex or nested views it's rather inconvenient to have to define all the links separately as we need to hold a reference to every field.

Instead I'd been keen to introduce the ability to set links directly on traits. This way I can bind to the view's traits without needing to hold a reference to the object. Proposing something along the following lines so that it's backwards compatible with current traitlet usage.

```python
from traitlets import SourceLink, Float, HasTraits
import ipywidgets as widgets

class Model(HasTraits):
   a = Float()

def view(model):
    return widgets.HBox([
        widgets.Label("Value A"),
        widgets.FloatText(value = SourceLink(model, "a"))
    ])

x1 = Model()
view(x1)
```

Having tested locally I think this is easily achievable with minimal changes. However before submitting a PR I wanted to check if any views or comments on the idea or how to approach it.

InfiniteChai:
Proposed change to allow setting of a link directly on a trait property, rather than needing to use an explicit link between target and source.
Closes #641

### DIFF
--- a/traitlets/tests/test_traitlets.py
+++ b/traitlets/tests/test_traitlets.py
@@ -2024,7 +2024,7 @@ class TestLink(TestCase):
             a = Int()
 
         x1 = A()
-        x2 = A(a = SourceLink(x1, "a", (lambda x: x+1, lambda x: x-1)))
+        x2 = A(a = SourceLink(x1, "a", transform=(lambda x: x+1, lambda x: x-1)))
         x1.a = 3
         self.assertEqual(x2.a, 4)
         x2.a = 1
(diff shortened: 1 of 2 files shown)


noelportillo (approved):

Carreau:
closing as stale, sorry this did not went further

PR #646, opened 2021-02-27
Fix Container.from_string

### LINKED ISSUE #645: Issue parsing a single instance of TagRemovePreprocessor
```
pip show traitlets
Name: traitlets
Version: 5.0.5
```
Calling `voila` with the `TagRemovePreprocessor.remove_cell_tags` option fails for a single tag:
```
voila notebook.ipynb --TagRemovePreprocessor.remove_cell_tags tag
....
traitlets.traitlets.TraitError: The 'remove_cell_tags' trait of a TagRemovePreprocessor instance expected a set, not the str 'tag'.
```

But when providing 2 tags, there's no issue:

`voila .\notebook.ipynb --TagRemovePreprocessor.remove_cell_tags tag0 --TagRemovePreprocessor.remove_cell_tags tag1`

rmorshea:
Closes: #645 

It looks like #645 might actually be a slightly bigger problem than it would initially seem. You can't even do `"{'tag'}"` and get it to work because the `Container` calls `self.validate(None, value)` and fails because, down the callstack, a `HasTraits` instance is expected instead of `None`.

Assuming we fix that though, it looks like this issue would still persist because `DeferredConfigString` calls `Container.from_string` which blindly attempts a `literal_eval` and fails (because `tag` is not a literal Python expression), resulting in a fallback value (the raw string) being used; thus the `TraitError`. The solution to this (from what I can tell) is for `Container.from_string(s)` to call `Container.from_string_list([s])`.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -608,7 +608,7 @@ class TraitType(BaseDescriptor):
             return value
         if hasattr(self, 'validate'):
             value = self.validate(obj, value)
-        if obj._cross_validation_lock is False:
+        if obj is not None and obj._cross_validation_lock is False:
             value = self._cross_validate(obj, value)
         return value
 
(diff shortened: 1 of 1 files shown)


willingc:
@rmorshea Closed/reopened this PR to run GH Actions.

PR #649, opened 2021-03-15
Add TOML support

### LINKED ISSUE #648: Add toml support
It would be very nice if the `ConfigLoader` class would support `toml` configuration files.
Given that `toml.load()` returns `dict` object and the implementation for JSON exists, the changes should be rather small.

mexanick:
- implement class TOMLFileConfigLoader
- implement corresponding test case
- add toml to requires list in setup.py

Fixes #648

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,7 @@ setuptools_args = {}
 
 install_requires = setuptools_args['install_requires'] = [
     'ipython_genutils',
+    'toml',
 ]
 
 extras_require = setuptools_args['extras_require'] = {
--- a/setup.py
+++ b/setup.py
@@ -105,11 +105,11 @@ setuptools_args = {}
 
 install_requires = setuptools_args['install_requires'] = [
     'ipython_genutils',
-    'toml',
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['pytest'],
+    'test': ['pytest', 'toml'],
+    'with-toml': ['toml'],
 }
 
 if 'setuptools' in sys.modules:
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -9,7 +9,11 @@ import os
 import re
 import sys
 import json
-import toml
+try:
+    import toml
+    HAS_TOML = True
+except ImportError:
+    HAS_TOML = None
 import warnings
 
 from ..utils import cast_unicode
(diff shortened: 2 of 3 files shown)


mexanick:
can it be related to https://github.com/jupyter/notebook/pull/5986?

ivanov:
It's great to see an addition of a format for traitlets, but I think it would make more sense to not make it a hard dependency. TOMLFileConfigLoader could wrap the currently top-level "import toml" in a try-except to not force the installation of toml for anyone who won't be using that feature, but giving a helpful error message for those wanting to use it to install `toml`.

willingc:
@rmorshea @ivanov I've closed/reopened this PR to trigger build on GH-Actions. Tests are green now on GH.

Probably the next step would be to determine offer @mexanick some guidance to avoid the hard dependency.

mexanick:
I have added the optional import of toml. The only thing which is suboptimal now is the testing of optional dependency (it is a bit ugly)

rmorshea:
Normally in these situations, I try to isolate this sort of "environment dependent code" in its own module so you don't have to do any `try/except` handling. Then, when the user tries to import it, if they haven't installed `toml`, they get an import error complaining that `import toml` failed. If no-one else has objections, my suggestions would be to add a new `traitlets.config.toml_loader` module containing the `TOMLFileConfigLoader`. This would set a precedent for future modules of the form `{format}_loader`. For example, if one wanted a YAML config loader, you'd create an isolated `traitlets.config.yaml_loader` module.

traitlets/config/loader.py
```diff
@@ -9,6 +9,11 @@
 import re
 import sys
 import json
+try:
+    import toml
```
bollwyvl, line 13: we should likely now prefer py3.11+ stdlib's `tomllib` and fall back to `tomli`

setup.py
```diff
@@ -108,7 +108,8 @@
 ]
 
 extras_require = setuptools_args['extras_require'] = {
-    'test': ['pytest'],
+    'test': ['pytest', 'toml'],
+    'with-toml': ['toml'],
```
bollwyvl, line 112: would probably just add `tomli ; python_version < 3.11`

traitlets/config/loader.py
```diff
@@ -598,6 +604,69 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        dct = self._read_file_as_dict()
+        self.config = self._convert_to_config(dct)
+        return self.config
+
+    def _read_file_as_dict(self):
+        with open(self.full_filename) as f:
+            return toml.load(f)
+
+    def _convert_to_config(self, dictionary):
+        if 'version' in dictionary:
```
bollwyvl, line 644: i don't recall having seen `version` on any of the other sources, and don't see why we'd want to add that constraint here unless it was added to all of them, which would be a major version bump

noelportillo (approved):

Carreau:
closing as stale. Sorry this did not land,.

PR #650, opened 2021-03-18
DOC: change flag -l -> -d

mgeier:
I guess this is a hypothetical example and there is no way to test this, but I guess it makes more sense to use the same flag twice. Otherwise it would create to separate `dict` objects I guess?


ivanov:
Thank you @mgeier ! Looks good to me (and the Travis CI failure is unrelated).

PR #653, opened 2021-04-10
Switch from Travis CI to Github actions

### LINKED ISSUE #652: Move CI to GH Actions
See https://discourse.jupyter.org/t/ipython-github-org-out-of-travis-ci-hours/8711

takluyver:
Remember me? :stuck_out_tongue: This seemed like something I could tackle fairly easily.

For now, I've combined the tests of downstream projects into a single job. This is a little bit slower, but it seemed simpler to do them like that than using variables or templating. It's easily changed if you prefer. Other than that, I've aimed to directly port what Travis was doing to Github actions.

Closes #652.


rmorshea:
Thanks! Will try and review this tomorrow.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
+
+jobs:
+  tests:
+
+    runs-on: ubuntu-latest
```
rmorshea, line 12: Should we also do a matrix on OS too?

takluyver: I usually wouldn't bother for a pure-Python package unless there's some particular reason to think its behaviour will differ between platforms. But it should be easy to do - at least for Mac - if you'd like.

willingc (approved):
Thanks @takluyver. I'm going to go ahead and merge. I'll open a new issue re: https://github.com/ipython/traitlets/pull/653#discussion_r611137655

willingc:
Here's the new issue re: os matrix https://github.com/ipython/traitlets/issues/654

Issue #661, opened 2021-04-20
Test fails with python 3.10


opoplawski:
Fedora is looking to move to python 3.10 soon.  traitlets test fails with:
```
platform linux -- Python 3.10.0a7, pytest-6.2.3, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python3

________________ TestUseEnum.test_assign_scoped_enum_value_name ________________

self = <traitlets.tests.test_traitlets_enum.TestUseEnum testMethod=test_assign_scoped_enum_value_name>

    def test_assign_scoped_enum_value_name(self):
        # -- CONVERT: string => Enum value (item)
        scoped_names = ["Color.red", "Color.green", "Color.blue", "Color.yellow"]
        for value in scoped_names:
            example = self.Example()
            example.color = value
            self.assertIsInstance(example.color, Color)
>           self.assertEqual(str(example.color), value)
E           AssertionError: 'red' != 'Color.red'
E           - red
E           + Color.red

traitlets/tests/test_traitlets_enum.py:90: AssertionError
```

hroncok:
There were some changes in repr() in latest alpha of Python 3.10.
https://docs.python.org/3.10/whatsnew/changelog.html#python-3-10-0-alpha-7

bpo-40066: Enum: adjust repr() to show only enum and member name (not value, nor angle brackets) and str() to show only member name. Update and improve documentation to match.

bpo-40066: Enum’s repr() and str() have changed: repr() is now EnumClass.MemberName and str() is MemberName. Additionally, stdlib Enum’s whose contents are available as module attributes, such as RegexFlag.IGNORECASE, have their repr() as module.name, e.g. re.IGNORECASE.
https://bugs.python.org/issue40066

olgarithms:
Hello! I am going to work on this :)

PR #663, opened 2021-05-07
Fix string assertion for python3.10

olgarithms:
Closes https://github.com/ipython/traitlets/issues/661.


hroncok:
Note that this got reverted in Python 3.10.0b4, see https://mail.python.org/archives/list/python-dev@python.org/message/LSTMFAPSPD3BGZ4D6HQFODXZVB3PLYKF/

olgarithms:
Ah, just saw this. Should I revert the commit?

PR #667, opened 2021-06-16
Setting one element of a trait list from cli not working

fcollonval:
Specifying only one element in a list is not working:

--C.list_trait x

The element will be interpreted as string => no matching type error.

Cause: https://github.com/ipython/traitlets/blob/191ff75e93d56446b2c051f4dbb33e2f0e7be71d/traitlets/config/loader.py#L671


fcollonval:
Superseeded by #670

PR #670, opened 2021-08-17
Fixes #668 ( passing a single element for a list on CLI)

### LINKED ISSUE #668: Single list element interpreted as string => error
Reported on JupyterLab Gitter by @mlucool

> Hello. What's the correct way to pass --LabApp.extra_labextensions_path=/path/to/dir on the CLI when there is exactly one extra path? Passing that gives "The 'app_dir' trait of a LabApp instance expected a unicode string, not the NoneType None.", but if I do --LabApp.extra_labextensions_path=/path/to/dir --LabApp.extra_labextensions_path=/dev/null it works as expected. More generally, how do you pass a traitlet list of a single value via CLI?

I opened #667 with a unit test that fails.

Carreau:
cc @fcollonval

### DIFF
--- a/traitlets/config/tests/test_loader.py
+++ b/traitlets/config/tests/test_loader.py
@@ -331,7 +331,6 @@ class TestKeyValueCL(TestCase):
         cl = self.klass(log=log)
         argv = ["--C.list_trait", "x"]
         config = cl.load_config(argv)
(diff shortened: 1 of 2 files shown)


fcollonval:
Thanks a lot @Carreau I can confirm it fixes #668

PR #678, opened 2021-09-27
FIX: account for changes in py310

tacaswell:
In https://github.com/python/cpython/pull/28420 the order of setting the
default values on the Namespace object and processing the Actions was reversed.  This means
that in the `__call__` method of `_FlagAction` the `_flags` attribute has not
yet been put on the namespace.

This change makes `_FlagAction.__call__` forgiving if the `_flags` attribute
does not exist (by creating it!).


minrk:
Thanks!

PR #683, opened 2021-10-17
Use codecov Github action v2

takluyver:
The v1 action is deprecated due to security concerns, and will stop working in a few months:

https://github.com/marketplace/actions/codecov
https://about.codecov.io/blog/introducing-codecovs-new-uploader/

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -35,13 +35,16 @@ jobs:
         iptest
     - name: Run tests nbconvert
       run: |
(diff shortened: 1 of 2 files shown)


takluyver:
I noticed it was trying to upload coverage info after running tests of downstream projects, but those tests weren't collecting coverage data. I've configured the test suites using pytest to collect that information. But this can easily be removed if you prefer not to count coverage from downstream tests. It doesn't appear to make any difference at the moment.

PR #698, opened 2022-04-01
Application: add logging_config trait

### LINKED ISSUE #688: log: allow log handlers to be configured
We can configure the log level and the log format, however, I don't think we can configure the log handlers?

There are cases where it would be helpful to be able to configure the application log to write to a file for reference as well as outputting to the default stream handler.

A simple solution would be to add a new trait for defining additional log handlers. A more complex solution could use a [logging config](https://docs.python.org/3/library/logging.config.html?highlight=logging) to configure one or more handlers, potentially with different formats, etc. I don't think users would expect this to be reactive to config changes.

Has this been discussed before? Does this sound sensible?

oliver-sanders:
Closes #688

Allows fine configuration of logging via a `logging.config.dictConfig`, useful for configuring additional log handlers/formatters and advanced logging setups.

* Fixes doctests and gets them running with pytest.
* Unifies the pre-existing logging configurations into a standard "base" logging configuration dict.
* Adds a new trait for providing a logging config dict which is merged into the "base" configuration allowing adding configs as well as redefining existing ones.
* Changes the default logger level from WARN to DEBUG and the default handler level from undefined to WARN (I think the log_level should apply to the handler not the logger otherwise).

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,8 @@
 [pytest]
-addopts = --durations=10 -ra --showlocals
+addopts = --durations=10 -ra --showlocals --doctest-modules
+testpaths =
+    traitlets
+    examples
 
 filterwarnings=
    ignore
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -4,16 +4,17 @@
 # Distributed under the terms of the Modified BSD License.
 
 
+from collections import defaultdict, OrderedDict
+from contextlib import suppress
+from copy import deepcopy
 import functools
 import json
 import logging
+from logging.config import dictConfig
 import os
 import pprint
 import re
 import sys
-import warnings
-from collections import OrderedDict, defaultdict
-from copy import deepcopy
 from textwrap import dedent
 
 from traitlets.config.configurable import Configurable, SingletonConfigurable
@@ -38,6 +39,7 @@ from traitlets.traitlets import (
     observe_compat,
 )
 from traitlets.utils.text import indent, wrap_paragraphs
+from traitlets.utils.nested_update import nested_update
 
 from ..utils import cast_unicode
 from ..utils.importstring import import_item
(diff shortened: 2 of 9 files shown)


blink1073:
Hi @oliver-sanders, it looks like the switch to using `black` caused some merge conflicts.  

The recommended way to handle this is to:

Run Black on modified files, keep a copy aside, start a new branch from "main", move the modified files back into place, and commit the result.

oliver-sanders:
Ok, that should do it, have kept a copy of the original branch for reference - https://github.com/ipython/traitlets/compare/main...oliver-sanders:688.20220411T1120?expand=1

blink1073 (approved):
This is great, thank you!

blink1073:
Looks like another doctest skip is needed:

```
 =================================== FAILURES ===================================
_______________ [doctest] traitlets.tests._warnings.all_warnings _______________
023     ...     warnings.warn(RuntimeWarning("bar"))
024 
025     We raise the warning once, while the warning filter is set to "once".
026     Hereafter, the warning is invisible, even with custom filters:
027     >>> with warnings.catch_warnings():
028     ...     warnings.simplefilter('once')
029     ...     foo()
030 
031     We can now run ``foo()`` without a warning being raised:
032     >>> from numpy.testing import assert_warns
UNEXPECTED EXCEPTION: ModuleNotFoundError("No module named 'numpy'")
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/doctest.py", line 1337, in __run
    compileflags, 1), test.globs)
  File "<doctest traitlets.tests._warnings.all_warnings[3]>", line 1, in <module>
ModuleNotFoundError: No module named 'numpy'
```

oliver-sanders:
> Looks like another doctest skip is needed:

Test was failing in CI because numpy dependency would be needed, added skips for now.

blink1073:
Thanks again!  The downstream failure is being handled in #701

Zsailer:
I think this might have broken some things: https://github.com/ipython/traitlets/issues/722

I'll try to dive into this today/tomorrow.

PR #699, opened 2022-04-07
Add pre-commit and run auto-formatters

blink1073:
This should be merged as a merge commit so git-blame-ignore-revs works.


blink1073:
Downstream test failure is unrelated

Carreau:
+1 , I trust you if you think precommit will help. I prefer merge commit
anyway as many things get screwed up by squash and merge and co,  like
finding all merged branches.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,79 @@
...
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        files: \.py$
+        args: [--profile=black]
+
+  # - repo: https://github.com/pre-commit/mirrors-mypy
+  #   rev: v0.942
+  #   hooks:
+  #     - id: mypy
```
rmorshea, line 37: I think this can be removed

rmorshea (commented):
Could you update the development installation instructions to explain how to set up pre-commit?

I've also found that people can sometimes find pre-commit frustrating when hooks cannot automatically fix the issues they discover. For example, `black` works great as a pre-commit step for most people because it automatically fixes any stylistic mistakes, but `flake8` can be annoying being it reports the error without fixing it and will block you from committing until you do. If the commit is really just a WIP and you don't want to fix everything you have to know to go look at the pre-commit docs to know how to turn off its checks for that commit. 

Pre-commit checks like `flake8` are fine to run in CI, but maybe not by default on ever commit a developer might make as they're writing code. Given this, it looks like pre-commit allows you to configure when hooks will run [using `stages`](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages). Perhaps for those hooks which do not automatically correct mistakes they could be marked as `manual` and only run in CI.

rmorshea:
Shoot, didn't see this was already merged. Maybe these minor nits could be done in a follow up?

blink1073:
I'll follow up with docs and moving some of the checks to ci-only by using the manual stage.

.pre-commit-config.yaml
blink1073: I am adding this in a follow up PR as well

rmorshea:
Awesome!

blink1073:
Great feedback by the way @rmorshea, I'll use that approach across repos.

PR #701, opened 2022-04-08
Fix downstream test

blink1073:


.github/workflows/downstream.yml
```diff
@@ -24,24 +24,33 @@ jobs:
...
+          pip install ipython[test] nbconvert[test] \
                       notebook[test] ipywidgets[test]
+          # Work around https://github.com/jupyter/nbconvert/issues/1754
+          pip install "bleach<5"
           pip freeze
 
       - name: Run tests IPython
         run: |
-          iptest
+          cd $HOME
```
rmorshea, line 35: why `cd $HOME`?

blink1073: To avoid using our own pytest.ini as the settings

PR #705, opened 2022-04-13
Add mypy support

blink1073:
- Adds minimal typing support with semi-strict config so that downstream projects can use traitlets typings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,11 +37,14 @@ repos:
         files: \.py$
         args: [--profile=black]
 
-  # - repo: https://github.com/pre-commit/mirrors-mypy
-  #   rev: v0.942
-  #   hooks:
-  #     - id: mypy
-  #       stages: [manual]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.942
+    hooks:
+      - id: mypy
+        args: ["--config-file", "pyproject.toml"]
+        exclude: "traitlets/.*tests/.*.py"
+        additional_dependencies: [pytest]
+        stages: [manual]
 
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.6.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
       - id: mypy
         args: ["--config-file", "pyproject.toml"]
         exclude: "traitlets/.*tests/.*.py"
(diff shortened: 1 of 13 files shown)


bollwyvl:
Hooray more static analysis!

Can't claim I understand the downstream issue, would it have been fixed by the recent release?

Another thing, and down here at a no-dep place is as good as any to talk about it: might we start normalizing the generation, and archiving as build artifacts, the human- and machine-readable reports from more of these tools? 

For example, the mypy html report (needs `lxml`) is kind of interesting, while its json/xml reports are perhaps even more interesting as data over time. `pytest-html` isn't terrible, but the built-in xunit2 output is a _de facto_ standard which can be consumed by many tools.

If these report bundles were repeatably showing up with predictable names, we could have something downstream that periodically harvested them, checked them in, and either built sites (or just binders) that could help show some of these impacts over time _and_ across/down the stack.

Best of luck on these efforts!

blink1073:
> Can't claim I understand the downstream issue, would it have been fixed by the recent release?

Right now we just have to treat `traitlets` as `Any` and ignore imports, this minimal step give us some inferred typing from mypy.

> might we start normalizing the generation, and archiving as build artifacts, the human- and machine-readable reports from more of these tools?

Where would those reports live, as part of the sdist?

bollwyvl:
I guess I meant, explicitly, the [_Test downstream projects_ fail](https://github.com/ipython/traitlets/runs/6005518104?check_suite_focus=true#step:7:52) on notebook... didn't look into too hard

> Where would those reports live, as part of the sdist?

Nah, just uploaded to CI: e.g. after running

```bash
- name: lint
  run: |
    mypy --cobertura-xml-report build/reports/mypy/cobertura --html-report build/reports/mypy/html
- name: test
  run: |
    pytest --junitxml=build/reports/pytest/xunit.xml
```

And then

```yaml
      - name: upload test reports
        if: always()
        uses: actions/upload-artifact@v3
        with:
          name: traitlets-test-artifacts-${{ github.run_number }}
          path: build/reports
```

Then have a separate job come along and pluck them for e.g. merges to `main`, and stuff _those_ just right into an ever-growing branch on a repo :shrug:?

Some similar stuff happens on conda-forge, e.g. https://github.com/regro/cf-graph-countyfair

blink1073:
The notebook tests themselves are failing on the 6.4.x branch, Eric Charles is actively working on getting them fixed.

> Nah, just uploaded to CI: e.g. after running

Sounds reasonable to me if you want to experiment with it.

PR #713, opened 2022-04-23
Typing: tell mypy we have type, update __all__.

Carreau:
Without the empty file py.typed mypy is unsure it should follow type
annotation on projects that depends on traitlets.

Once added it looks in __all__ and says that things like Unicode, Bool,
are not supposed to exists, as not in __all__


blink1073 (approved):
Thanks!

PR #724, opened 2022-05-12
Fix KeyError in Application.get_default_logging_config

tonyroberts:
Fixes #723

PR #733, opened 2022-05-27
fix KeyError "loggers"

Roffild:
Fix run in pythonw.exe


Roffild:
https://github.com/ipython/traitlets/pull/724

PR #749, opened 2022-06-17
fix version_info

minrk:
no '.' in '.dev'


minrk:
Tools like hatch/tbump have a hard time with `version_info` tuples and their leading dots, etc.

We could switch to generating version_info from `__version__` which tools have an easier time dealing with. It feels wrong to me to go from version string to version info instead of the other way around, but tooling supports updating the string way better, so that's probably the practical solution

PR #773, opened 2022-09-08
Fix union from string

### LINKED ISSUE #772: Union trait does not parse correctly from strings
Union trait does not enumerate the types in the union when parsing from string which means that incorrect type gets inferred.

Correct inference of union types is desirable in https://github.com/jupyter-server/jupyter_server/pull/949 and https://github.com/ipython/ipython/pull/13745.

krassowski:
Fixes #772, adds unit test.


blink1073 (approved):
Nice, thank you!

blink1073:
I'm going to wait until next week to cut a release.

krassowski:
I will also take a day look tomorrow because I'm not sure if it should raise error at the end - probably should call super instead to check for none. I will follow up.

PR #774, opened 2022-09-09
Cover more Union parsing scenarios

krassowski:
Follow-up after #773 with more test cases and added notes in the documentation.


blink1073 (approved):
Thanks again!

PR #778, opened 2022-09-16
Update tests and docs to use non-deprecated functions

jasongrout:
As a followup to https://github.com/ipython/traitlets/pull/306, this updates tests and docs to use the non-deprecated value_trait and per_key_traits arguments.

I also fixed two spelling typos.

### DIFF
--- a/docs/source/using_traitlets.rst
+++ b/docs/source/using_traitlets.rst
@@ -143,8 +143,8 @@ properties.
 
     class Nested(HasTraits):
 
-        value = Dict(traits={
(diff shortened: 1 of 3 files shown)


jasongrout:
This fixes an issue we have in ipywidgets testing. It seems that when we import `from traitlets.tests.test_traitlets import TraitTestBase`, the deprecation warnings are triggered, which means we have deprecation warnings in our test suite from the traitlets codebase.

blink1073 (approved):
Thank you!

PR #788, opened 2022-10-20
Allow type checkers to infer trait types

vidartf:
We are currently encoding typing information for traits without integrating with the Python static typing system. This change is for allowing traitlets users to get static typing "for free" by using our trait types.

Ideally, the typing system would allow for default types, such that `TraitType == TraitType[t.Any, t.Any]`, and `TraitType[U] == TraitType[U, U]`, but for now we have to be verbose for this to work.

This is currently in a draft state, as some of the typings are still not accurately reflecting all the information we have. Potential improvements:
- [ ] The Union type should correctly surface its types as the typings union of those of its components
- [ ] Make the `Optional` part depend on allow_none (is it possible?).
- [ ] Improve types of class based traits
- [ ] Can we make the coercing variants not need to re-inherit `TraitType` ?
- [ ] Make Enum and UseEnum types better
- [ ] etc.

cc @maartenbreddels

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -518,6 +518,7 @@ class BaseDescriptor:
 G = t.TypeVar("G")
 S = t.TypeVar("S")
 
+Self = t.TypeVar("Self", bound="TraitType")  # Holdover waiting for typings.Self in Python 3.11
 
 class TraitType(BaseDescriptor, t.Generic[G, S]):
     """A base class for all trait types."""
@@ -855,7 +856,7 @@ class TraitType(BaseDescriptor, t.Generic[G, S]):
         warn("Deprecated in traitlets 4.1, " + msg, DeprecationWarning, stacklevel=2)
         self.metadata[key] = value
 
-    def tag(self, **metadata):
+    def tag(self, **metadata) -> "TraitType[G, S]":
         """Sets metadata and returns self.
 
         This allows convenient metadata tagging when initializing the trait, such as:
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -2316,7 +2316,7 @@ class CFloat(Float, TraitType[float, t.Any]):
         return _validate_bounds(self, obj, value)
 
 
-class Complex(TraitType[complex, complex | tuple[float, int]]):
+class Complex(TraitType[complex, complex | float | int]):
     """A trait for complex numbers."""
 
     default_value = 0.0 + 0.0j
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -515,8 +515,9 @@ class BaseDescriptor:
         pass
 
 
-G = t.TypeVar('G')
-S = t.TypeVar('S')
+G = t.TypeVar("G")
+S = t.TypeVar("S")
+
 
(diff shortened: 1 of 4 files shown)


blink1073:
Nice!  You may want to play with the `Application` object in your mypy tests, to see if we can remove the explicit typings there.

vidartf:
@maartenbreddels I thought mypy already did that for class access (it isn't using `__get__` type for class access, is it?).

rmorshea:
@vidartf, it does not. MyPy checks for polymorphic behavior on `__get__` to determine class vs instance access return type. Here's [a sample demonstrating this](https://mypy-play.net/?mypy=latest&python=3.10&gist=868f9943de0319850445cc46f9e8981d).

maartenbreddels:
> it isn't using `__get__` type for class access, is it?

it is, therefore the `if obj is None:` check

Thanks @rmorshea for the gist. Do you have an idea how to approach the `allow_none` -> Optional issue? Would that be possible with overload or should dataclass_transform be used for that?

rmorshea:
[This](https://mypy-play.net/?mypy=latest&python=3.10&gist=1bd03a489365c03ad5ddecbf282226a7) appears to work as expected. You can overload the `__new__` method (just while type checking) to narrow the type. Unfortunately I don't see a clear way to avoid copy-pasting this to each `TraitType` because there's no way to generalize add args to type variables (e.g. `def __new__(cls: type[C], kind: T) -> C[T]: ...`).

vidartf:
Thanks for the input. I thought I had checked it, but I must have dreamed it. Now added a test for the typing implicitly via using .tag()

traitlets/traitlets.py
```diff
@@ -855,7 +857,7 @@ def set_metadata(self, key, value):
         warn("Deprecated in traitlets 4.1, " + msg, DeprecationWarning, stacklevel=2)
         self.metadata[key] = value
 
-    def tag(self, **metadata):
+    def tag(self, **metadata) -> "TraitType[G, S]":
```
maartenbreddels, line 860: I think with the `typing_extension` package we can do this:
```suggestion
    def tag(self, **metadata) -> typing_extension.Self":
```

blink1073:
The failures are due to https://github.com/davidfritzsche/pytest-mypy-testing/issues/35

maartenbreddels:
> [This](https://mypy-play.net/?mypy=latest&python=3.10&gist=1bd03a489365c03ad5ddecbf282226a7) appears to work as expected. You can overload the `__new__` method

Whow, thanks, you never stop learning with typing tricks! I think it's acceptable, the number of traits types isn't huge, and it will give us really good type checks with pylance and mypy!

blink1073:
We can pin to `pytest<7.2` for now to unblock this PR.

blink1073:
I opened https://github.com/vidartf/traitlets/pull/1 exploring the idea @rmorshea had, with a bit of [tweaking](https://mypy-play.net/?mypy=latest&python=3.10&gist=1bd03a489365c03ad5ddecbf282226a7) to get it to work with other arguments like `help`

blink1073:
It turns out we will still need the explicit types on the `Application` object to allow subclasses to set `name = "foo"` for example, which is not a huge problem.

PR #809, opened 2022-12-07
Adopt ruff and address lint

blink1073:
cf https://github.com/jupyter-server/jupyter_server/pull/1114

We weren't actually running mypy...

### DIFF
--- a/traitlets/config/tests/test_application.py
+++ b/traitlets/config/tests/test_application.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import sys
+import typing as t
 from io import StringIO
 from tempfile import TemporaryDirectory
 from unittest import TestCase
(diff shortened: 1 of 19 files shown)


traitlets/config/tests/test_application.py
```diff
@@ -375,7 +372,7 @@ def test_aliases_multiple(self):
         # Test multiple > 2 aliases for the same argument
         class TestMultiAliasApp(Application):
             foo = Integer(config=True)
-            aliases = {("f", "bar", "qux"): "TestMultiAliasApp.foo"}
+            aliases = {("f", "bar", "qux"): "TestMultiAliasApp.foo"}  # type:ignore
```
azjps, line 375: I think this might be fixed by https://github.com/ipython/traitlets/pull/810/commits/b805c9ed63330df126bc337f4b028b27def81c97#diff-5453d413070d06bcf7ba7f017b9cf364e774627fad69964a6471449e5c16dc37R361?

PR #810, opened 2022-12-07
Minor fixes for Application.aliases

azjps:
1. Fix the type annotations for Applications.flags / aliases, which support tuples as keys and results in linting errors (e.g. in the test suite).
2. Move the logic for determining what to do with an unrecognized command-line argument to a helper method. Currently a warning is logged, but in past versions of traitlets an error was raised, and this might be more user-friendly in some cases especially where the stderr can be pretty verbose (e.g. launching Jupyterlab).


blink1073 (approved):
Thank you!

Issue #539, opened 2019-11-07
Add support for bash completion to Application


telamonian:
A traitlets Application is able to accept various command line arguments. For aliases, flags, and subcommands, at least, I think it would make a lot of sense if traitlets provided an automatic (or at least easy) way to set up the related completions for bash (and possibly other shells). This would make an Application's command line arguments easier to enter, and also easier to discover.

I work a lot on JupyterLab core and the related ecosystem. My use case for this feature is so that I never have to fully type out

```
jupyter labextension install ...
```

ever again. I think this feature belongs in the `traitlets` package specifically since:
- It makes the most sense to implement completion support adjacent to the command line argument handling
- A general solution to this issue would be potentially of use to every project that uses Application

In the past, I've found that by far the simplest way to implement bash completion for a Python script is to use the `argcomplete` package. Given that a script is parsing cmd line args using the `argparse` standard lib package, you can set up bash completion with `argcomplete` by adding 3 lines to the script.

From skimming the code in `traitlets/config/application.py` and `traitlets/config/loader.py` it appears that at a low level traitlets is indeed using `argparse` to parse cmd line args (though please correct me if I'm wrong about that). So I'll take a stab at making a PR for this issue based on `argcomplete`.

azjps:
Should be (mostly) done by #812 & #829 😄

PR #811, opened 2022-12-07
Initial support for argcomplete for KVArgParseConfigLoader

azjps:
After enabling [`argcomplete`](https://github.com/kislyuk/argcomplete) shell completion for a `traitlets.Application` based script (e.g. via `activate-global-python-argcomplete`), this commit sets up tab auto-completion for command-line flags and aliases. (Note that `argcomplete` would be completely optional and not a dependency.)

Argcomplete completers can be added for a trait by using an `"argcompleter"` metadata tag, which should be a function that takes keyword arguments (passed down from `argcomplete`) and returns a list of string completions. Completers are set up for `Bool` (`"true", "false", "1", "0"`) & `Enum`.

This commit does **not** yet add general support for arbitrary class traits of the form `--Class.trait=xyz`, as these are currently not added to the `ArgumentParser` instance, but rather directly parsed from `sys.argv`. It is probably possible to add this support for the classes in `Application.classes`. IMHO, the most user-friendly way of doing this would be to allow `--<TAB>` to also show the completions `--Class1. --Class2.` and so forth, enumerating the Configurable classes but not spamming all of the available traits. Only once the class has been completed out e.g. `--Class1.<TAB>` would the traits of the class become completable. While this should be possible with `argcomplete` I haven't looked closely enough to figure out what the easiest way to accomplish this would be.

Issue: #539

Example:

```bash
~/dev/traitlets$ python examples/myapp.py --[TAB]
--debug      --disable    --enable     --enabled    --help       --i          --j          --log_level  --mode       --name       --running

~/dev/traitlets$ python examples/myapp.py --running [TAB]
0      1      false  true

~/dev/traitlets$ python examples/myapp.py --running true --[TAB]
--debug      --disable    --enable     --enabled    --help       --i          --j          --log_level  --mode       --name       --running

~/dev/traitlets$ python examples/myapp.py --running true --mode o[TAB]
off    on     other
```

If this idea seems good to pursue, I can add documentation, update dev dependencies, and so forth.


blink1073:
Very cool!  I think it makes sense to add `argcomplete` as a test dependency.  I am also 👍🏼 on completing the class names, that would be very helpful IMO.

azjps:
Turns out `argcomplete` doesn't currently have built-in support to partially complete options, in particular if one adds `--Class.` as an `argparse` option, then `argcomplete` will add a space afterwards when its tab-completed, which is user-unfriendly.

I managed to hack around a bit and directly add all of the config traits as `parser.add_argument("--Class.<trait>")`, but only when we detect that something that looks like `--Class.*` or its following arguments are being completed. It does use `argcomplete`'s internals so maybe there's something better that could be done.

I've added an example script `examples/argcomplete_app.py`, the docstring has different completions I've tried to test.

Another example: add `# PYTHON_ARGCOMPLETE_OK` as the second line for `bin/ipython`, and tab-complete away:

```bash
$ ipython --
--Application.               --StoreMagics.               --config                     --m                          --nosep
--BaseFormatter.             --TerminalIPythonApp.        --confirm-exit               --matplotlib                 --pdb
--BaseIPythonApplication.    --TerminalInteractiveShell.  --debug                      --no-autoedit-syntax         --pprint
--Completer.                 --autocall                   --ext                        --no-autoindent              --profile
--HistoryAccessor.           --autoedit-syntax            --gui                        --no-automagic               --profile-dir
--HistoryManager.            --autoindent                 --help                       --no-banner                  --pylab
--IPCompleter.               --automagic                  --i                          --no-color-info              --quick
--InteractiveShell.          --banner                     --ignore-cwd                 --no-confirm-exit            --quiet
--InteractiveShellApp.       --c                          --init                       --no-ignore-cwd              --simple-prompt
--LoggingMagics.             --cache-size                 --ipython-dir                --no-pdb                     --term-title
--PlainTextFormatter.        --classic                    --log-level                  --no-pprint
--ProfileDir.                --color-info                 --logappend                  --no-simple-prompt
--ScriptMagics.              --colors                     --logfile                    --no-term-title

$ ipython --gui=
asyncio  glut     gtk      gtk2     gtk3     osx      pyglet   qt       qt4      qt5      tk       wx

$ ipython --gui g
glut  gtk   gtk2  gtk3

$ ipython --InteractiveShellApp.
--InteractiveShellApp.code_to_run                         --InteractiveShellApp.hide_initial_ns
--InteractiveShellApp.exec_PYTHONSTARTUP                  --InteractiveShellApp.ignore_cwd
--InteractiveShellApp.exec_files                          --InteractiveShellApp.matplotlib
--InteractiveShellApp.exec_lines                          --InteractiveShellApp.module_to_run
--InteractiveShellApp.extensions                          --InteractiveShellApp.pylab
--InteractiveShellApp.extra_extension                     --InteractiveShellApp.pylab_import_all
--InteractiveShellApp.file_to_run                         --InteractiveShellApp.reraise_ipython_extension_failures
--InteractiveShellApp.gui
```

Sub-commands might be tricky since IIRC they're handled by `traitlets` and not by `argparse` (and thus `argcomplete` won't know about them), I haven't started looking into it yet.

blink1073:
Looking good so far!

azjps:
@blink1073 I think I've gotten into a style battle with `pre-commit` which [re-adds some whitespace lines](https://github.com/ipython/traitlets/pull/811/commits/7279e3896e0b2b3b36ac393c5a8eaa2346482c46) that causes `ruff` lint to fail with `I001 Import block is un-sorted or un-formatted`. 🤣 

[`downstream_check > nbconvert` is also failing, I'm not sure what this indicates?](https://github.com/ipython/traitlets/actions/runs/3703688685/jobs/6275584915)

While stashing my changes related to subcommand handling, I decided instead to just go ahead with it, its probably very fragile but it works for the simple examples at least.

I added some stuff to the `config.rst` docs and `examples/` directory, haven't proofread it that closely though.

I've gotten some unit tests working for `argcomplete` & `traitlets.Application` which "mock" the environment variables being used while completing and redirecting the completion output to a `TemporaryFile`, its working in my local environments but failing in CI with `OSError` and a few other errors. I have to admit I don't really know what's going on there, but for now I've just marked them with `pytest.mark.skip`.

Besides these points, I'm satisfied with the state of these commits, probably the next thing I'll look into is opening PRs against the major traitlets-based projects (ipython, jupyter, etc) to actually enable `argcomplete`.

blink1073:
#814 fixes the linting issue

blink1073:
The nbconvert failure is unrelated.  I'll take a look at the CI skipping, I'd rather not merge without that in place.  It might be a bit before I get to it, I'm recovering from an illness.

azjps:
No worries. I re-enabled the tests and they seem to deterministically fail here, I've been trying to bisect towards the issue in #815 but haven't gotten a simpler example to fail yet 😞

azjps:
Changing `TemporaryFile` -> `StringIO` was enough to get all the unit tests to pass, unfortunately pytest still crashes even though all tests pass. Possibly in coverage checking or something else and the errors are a bit difficult to scrutinize -- probably some side effect of `argcomplete` but having trouble narrowing it down

blink1073:
> Use pytest-mock to patch os.fdopen

That seems reasonable.  Great work!

pyproject.toml
```diff
@@ -45,14 +45,14 @@ nowarn = "test -W default {args}"
 
 [tool.hatch.envs.cov]
 features = ["test"]
-dependencies = ["coverage", "pytest-cov"]
+dependencies = ["coverage", "pytest-cov", "argcomplete"]
```
blink1073, line 48: "argcomplete" is already handled by the "test" feature, here and in the typing env.

traitlets/config/argcomplete_config.py
```diff
@@ -0,0 +1,203 @@
...
+import os
+import typing as t
+
+try:
+    import argcomplete  # type: ignore[import]
+    from argcomplete import CompletionFinder
+except ImportError:
+    # This module and its utility methods are written to not crash even
+    # if argcomplete is not installed.
+    class StubModule:
```
azjps, line 12: I tried to be a bit defensive here and allow this module to be importable even if `argcomplete` is not installed, maybe there's a more common pattern that can be used here?

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
+## 5.8.0
```
blink1073, line 5: The changelog is updated automatically during the release process, this change would interfere with that logic.

traitlets/config/argcomplete_config.py
blink1073: This is a nice approach, I'd say leave it.

blink1073:
I'll manually update the release notes during the release to add your language for the documentation, inlining here:

```
### Enhancements

- Shell command-line tab-completion via `argcomplete` [#811](https://github.com/ipython/traitlets/pull/811) ([@azjps](https://github.com/azjps))

### Documentation

- Additional `Application` examples and docs [#811](https://github.com/ipython/traitlets/pull/811) ([@azjps](https://github.com/azjps))
``

blink1073:
If you beat me to it, please remove the changelog and update the branch, otherwise I'll do it tomorrow and cut a minor release.  Thanks again!

CHANGELOG.md
azjps: Ah cool! Thought it might be the case but wasn't sure.

blink1073:
https://github.com/ipython/traitlets/releases/tag/v5.8.0

PR #777, opened 2022-09-14
Performance: decrease runtime overhead for constructing HasTraits (up to 20x faster)

maartenbreddels:
This makes creating ipywidgets about 2x faster (ignoring comm construction).

Related to, but superseded https://github.com/ipython/traitlets/pull/639

PR #812, opened 2022-12-09
fix: some descriptors raise AttributeError

maartenbreddels:
Regression, noticed in https://github.com/ipython/traitlets/pull/777#issuecomment-1344691036

@jeffyjefflabs I think this should fix your issue, but I'd like to be able to add a test to prevent further regression, any ideas on this?

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1006,7 +1006,13 @@ class MetaHasTraits(MetaHasDescriptors):
         mro = cls.mro()
 
         for name in dir(cls):
-            value = getattr(cls, name)
+            # Some descriptors raise AttributeError like zope.interface's
+            # __provides__ attributes even though they exist.  This causes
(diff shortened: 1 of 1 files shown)


rmorshea (approved):

jeffyjefflabs:
I can confirm that fixes it - thank you!  I added a print statement and the class being set up when the error occurs is `jupyter_client.provisioning.provisioner_base.KernelProvisionerBase`.  From the traceback, this is happening when sphinx sets up the myst_nb extension.  Unfortunately our code isn't public, but we are actually calling sphinx programmatically after some other code runs.  We do have zope.interface (mentioned in that comment) in the environment via twisted, so I'm guessing there's some interplay going on there.  Unfortunately I'm not having any luck reproducing with a minimal repo generated by `mystnb-quickstart`, but hopefully this will give you some ideas.

blink1073:
I'll go ahead and cut a release with this.

PR #816, opened 2022-12-16
define trait.__doc__ = trait.help

minrk:
improves inspection via tools like sphinx autodoc

related discussion: https://github.com/jupyterhub/autodoc-traits/issues/30


blink1073 (approved):
Thank you!

PR #824, opened 2023-01-06
fix: mro_trait can be unbound when the class is not in mro

### LINKED ISSUE #823: Bug in version 5.8.0
https://github.com/ipython/traitlets/blob/2cdbe59606ea23ee835f651303f8f75e0aeb9efd/traitlets/traitlets.py#L1022-L1030

Getting error UnboundLocalError: local variable 'mro_trait' referenced before assignment.

maartenbreddels:
Closes #823

Not so happy that it doesn't include a test yet, and I think a linter should have picked up this mistake. @blink1073 do you know why CI didn't see this lint issue?

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1020,6 +1020,7 @@ class MetaHasTraits(MetaHasDescriptors):
                 cls._traits[name] = value
                 trait = value
                 default_method_name = "_%s_default" % name
+                mro_trait = mro
                 try:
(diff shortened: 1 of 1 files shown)


blink1073:
Unbound locals are apparently tricky to catch with a linter, I found issues on both mypy and pyflakes where they punt on it.

blink1073 (approved):
Thanks!  I'll release this tomorrow.

PR #825, opened 2023-01-16
Maint: Remove potentially extra usage of Unicode.

Carreau:
As far as I can tell setting this to `Unicode` does not achieve much. For example the IPython application that derives from it set it to a plain string and it works.

Same for description.

From my dev folder I haven't found any usage of name or description that needs this to be a `Unicode`, we we might as well make it always a str, which simplify a bit the codebase/types.

Usually the simpler the better.

It is _technically_ backward incompatible.

See https://github.com/ipython/ipython/pull/13895 that remove the usage of Unicode in description

### DIFF
--- a/examples/subcommands_app.py
+++ b/examples/subcommands_app.py
@@ -27,7 +27,7 @@ class PrintHello(Configurable):
 
 
 class FooApp(Application):
-    name = Unicode("foo")
+    name = "foo"
     classes = [PrintHello]
     aliases = {
         "print-name": "PrintHello.greet_name",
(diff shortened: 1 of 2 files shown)


rmorshea:
Seems like a reasonable change, but wouldn't this also impact any assignment of `name` and `description` via the application's constructor, not just sub-classing?

Given that this is a breaking change, would this be included in a major release?

Carreau:
> but wouldn't this also impact any assignment of `name` and `description` via the application's constructor

Hum, that's a good question. 

Either it does not seem to do so, or there is not test for it. 
Let me check it and maybe add tests.

No objection to have this wait for a major release if it's breaking.

Carreau:
You are right, if it is not set to ` = Unicode()`, it can't indeed be modified via the constructor. 

Now the main question is do we want name to be settable via the constructor for the default application. And a test need to be added. 

Good catch.

Carreau:
See #826, that adds a test, I'll change that to draft and we can discuss it for next major release.

PR #826, opened 2023-01-20
Test that name and description can be set via constructor.

Carreau:
Alternative to #825, that adds tests for behavior.

I don't think we _need_ the ability to set the value via a constructor, And I'm generally in favor of simplifying, but at least this should have tests.


blink1073 (approved):
Thanks!

PR #834, opened 2023-02-18
Added type hints to HasTraits.observe() and friends. (#834

scholer:
When writing something like  `obj.observe(names='value')`, the type-checker in VScode (Pylance) will complain that `'value'` is not an instance of `Sentinel`, which is the type used for the default `Any` value for the `names` argument.

This PR fixes that by adding type-hints to `observe()`, `unobserve()`, `unobserve_all()`, and `parse_notifier_name()`, telling the type-checker that str-typed arguments are OK to use.

Let me know if this PR requires any further updates or changes.

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -234,7 +234,7 @@ def is_trait(t):
     return isinstance(t, TraitType) or (isinstance(t, type) and issubclass(t, TraitType))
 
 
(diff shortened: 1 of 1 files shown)


scholer:
@meeseeksdev tag enhancement

blink1073 (approved):
LGTM, thanks!

PR #855, opened 2023-08-21
Even though returning a value from a validation method is ideal, by no means is `proposal['value']` the only thing that can be returned

metaperl:
The [current docs on validation and coercion](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation-and-coercion) make it seem like `proposal['value']`  must be returned even though this is not the case.

Also, there are no examples of coercion in this section. But that's a topic for another time.


blink1073 (approved):
Thanks!

PR #858, opened 2023-09-09
rename `MyClass` to `School`

metaperl:
In the docs and sample code, `MyClass` is a rather confusing and artificial name. To give the docs and example a more practical flavor, I have renamed `MyClass` to `School` in both the docs and the example code.


blink1073 (approved):
LGTM, thanks!

Issue #871, opened 2023-09-26
More typing edge cases


blink1073:
Seen while updating `jupyter_server`:

```bash
jupyter_server/serverapp.py: note: In member "init_webapp" of class "ServerApp":
jupyter_server/serverapp.py:2103: error: Incompatible types in assignment (expression has type "None", variable has
type "dict[Any, Any]")  [assignment]
                self.ssl_options = None
```

```python
ssl_options = Dict(
        allow_none=True,
        config=True,
        help=_i18n(
            """Supply SSL options for the tornado HTTPServer.
            See the tornado docs for details."""
        ),
    )
```

Issue #869, opened 2023-09-26
`__doc__` no longer contains the traitlets help string for reference type traitlets (Dict, List, ...)


consideRatio:
https://github.com/jupyterhub/autodoc_traits is a project that help generate documentation for sphinx by inspecting traitlets. Since #818 was merged and released in 5.10.0, it has experienced issues detecting traitlets help strings via a `__doc__` attribute for some traitlet types.

Specifically the help string is no longer found in `__doc__` attribute of the traitlet types Dict, List, Instance, and Command traitlets, but it still is found for the traitlet types Bool, Integer, Unicode. Maybe this is relating to reference/value type differences?

I've opened https://github.com/jupyterhub/autodoc-traits/issues/55 about it, but since I'm unsure if its a bug in this project or `autodoc_traits` at this point, I'm opening this issue as well for now.

blink1073:
I tried the autodoc test suite with `traitlets@5.10.0` and was able to reproduce the failures.  With 5.10.1 all the tests pass:

```
======================================== 16 passed in 6.05s ========================================
```

consideRatio:
Thank you @blink1073 !!!

PR #873, opened 2023-09-29
More Typing Adoption

blink1073:
- Gets us closer to the goal of strict typing with the adoption of the `ANN` check in ruff, which lets us more easily toggle where to enforce typing.

- Moves the tests directories into a top level `tests` folder.

- Adds a regression test for #869 

- Fixes handling of `Type` trait and adds tests, partially addresses #871


blink1073:
`tests/test_traitlets.py` is a pure file move but it is showing up as a new file, maybe because it is so big.

blink1073:
Oh duh, the big diff is because we had to leave a few items in `traitlets/tests/test_traitlets.py` that are used downstream.

PR #881, opened 2023-10-14
Pass **kwds in MetaHasDescriptors __new__ and __init__

fleming79:
This is a modest change that adds `**kwds` to  `MetaHasDescriptors`  `__new__` and `__init__`  to match the signature defined in the [Python docs](https://docs.python.org/3.8/reference/datamodel.html?highlight=__init_subclass__#creating-the-class-object):

```
metaclass(name, bases, namespace, **kwds)
```

This enables the usage [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html?highlight=__init_subclass__#object.__init_subclass__) without having to overload `MetaHasDescriptors` to intercept the arguments passed to a subclass.

## Trivial example

```
class Philosopher(HasTraits):
    def __init_subclass__(cls, /, default_name, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.default_name = default_name

class AustralianPhilosopher(Philosopher, default_name="Bruce"):
    pass
```
### Before
image

### After
image

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -965,7 +965,7 @@ class MetaHasDescriptors(type):
     instantiated and sets their name attribute.
     """
 
-    def __new__(mcls, name, bases, classdict):  # noqa
+    def __new__(mcls, name, bases, classdict, **kwds):  # noqa
         """Create the HasDescriptors class."""
         for k, v in classdict.items():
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Sounds reasonable to me, thank you!

PR #884, opened 2023-10-27
Use passed-in `info` argument for `TraitType.error()`

kylebarron:
In my own project I'm subclassing `TraitType` to create my own traitlets, and working to provide good validation inside `validate()`, but my custom `info` messages weren't being shown to the user.


blink1073 (approved):
Thank you!

PR #895, opened 2023-12-12
FIX: map - to _ for environement variable prefix.

Carreau:
Loading from env use the name of the application as prefix, though dashes (-) are not valid in env variable names, so map ti underscore.


blink1073 (approved):
Thanks!

PR #906, opened 2024-06-10
Remove py2-compat hack from `bunch.__dir__`

sethtroisi:
This python2 hack was added in #238 but is no longer needed.

### DIFF
--- a/traitlets/utils/bunch.py
+++ b/traitlets/utils/bunch.py
@@ -23,6 +23,7 @@ class Bunch(dict):  # type:ignore[type-arg]
         self.__setitem__(key, value)
 
     def __dir__(self) -> list[str]:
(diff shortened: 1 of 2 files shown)


traitlets/utils/bunch.py
```diff
@@ -23,7 +23,6 @@ def __setattr__(self, key: str, value: Any) -> None:
         self.__setitem__(key, value)
 
     def __dir__(self) -> list[str]:
-        # py2-compat: can't use super because dict doesn't have __dir__
-        names = dir({})
+        names = dir(Bunch)
```
minrk, line 26: This should be the right super invocation, now that we can assume recent Python:

```suggestion
        names = super().__dir__()
```

PR #912, opened 2024-11-28
Suppress exceptions when closing handlers during `__del__`

minrk:
when `__del__` is called during process teardown, any amount of things may already be torn down and fail. Suppress those errors.

currently seeing this with `typing.cast` being None, preventing access of `self.log` (or any trait) from succeeding: https://github.com/jupyterhub/repo2docker/issues/1379

I'll work on a separate PR to remove most if not all use of `return cast(...)` which doesn't do anything more than `return x # type:ignore` other than introduce possible errors like this and (negligible) performance cost.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -1062,7 +1062,11 @@ class Application(SingletonConfigurable):
         sys.exit(exit_status)
 
     def __del__(self) -> None:
-        self.close_handlers()
+        # __del__ may be called during process teardown,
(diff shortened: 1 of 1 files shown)


minrk:
note that this suppresses the error that `close_handlers` on `__del__` actually fails to do what it's supposed to because the attribute-access of `self.log` fails. #913 is what fixes the underlying error.

Carreau (approved):

PR #917, opened 2025-02-04
Fix whitespace formatting in CLI help.

Carreau:
For traitlets that have long descriptions, and in particular that uses multiline strings, this lead to inconsistant indentation – and mmissing newlines in rewrapped paragraphs.

 - Each paragraph in wrap_paragraphs would be a list item, and thus need to be joined by 2 newlines (or we need to ensure that each list item in wrap_paragraph would end in a NL, but that woudl change API
 - When help is ina multiline string, if the line start on the first line:

    """like in

    this kind of help"""

    but not like

    """
    in this kind

    of documentation help
    """

    Wrap parragraph would not properly dedent.

### DIFF
--- a/traitlets/config/configurable.py
+++ b/traitlets/config/configurable.py
@@ -300,7 +300,7 @@ class Configurable(HasTraits):
         if helptext is None:
             helptext = trait.help
         if helptext != "":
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #918, opened 2025-03-09
Update `__new__` method to use Self type for improved type hinting

fleming79:
Recently Pylance (Pyright) started producing  `Any` as the type hint for HasTraits objects since Pylance v 2025.2.102 & Pyright v 1.1.395. This may be related to the first dot point in the Pyright release notes here: https://github.com/microsoft/pyright/releases/tag/1.1.395.

### Before
image

### After
image

Reference: https://typing.python.org/en/latest/spec/constructors.html#consistency-of-new-and-init

### DIFF
--- a/traitlets/traitlets.py
+++ b/traitlets/traitlets.py
@@ -1290,7 +1290,7 @@ class DefaultHandler(EventHandler):
 class HasDescriptors(metaclass=MetaHasDescriptors):
     """The base class for all classes that have descriptors."""
 
-    def __new__(*args: t.Any, **kwargs: t.Any) -> t.Any:
+    def __new__(*args: t.Any, **kwargs: t.Any) -> Self: # type:ignore[misc, type-var]
         # Pass cls as args[0] to allow "cls" as keyword argument
         cls = args[0]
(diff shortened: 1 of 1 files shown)


traitlets/traitlets.py
```diff
@@ -1290,11 +1290,7 @@ def class_init(self, cls: type[HasTraits], name: str | None) -> None:
 class HasDescriptors(metaclass=MetaHasDescriptors):
     """The base class for all classes that have descriptors."""
 
-    def __new__(*args: t.Any, **kwargs: t.Any) -> t.Any:
-        # Pass cls as args[0] to allow "cls" as keyword argument
-        cls = args[0]
-        args = args[1:]
-
+    def __new__(cls, /, *args: t.Any, **kwargs: t.Any) -> Self:
```
fleming79, line 1293: Use of the position-only character `/` means `cls` can be passed as a keyword argument. 

Refs: 
- https://peps.python.org/pep-0570/
- https://docs.python.org/3/glossary.html#term-parameter

minrk:
Thanks!

PR #923, opened 2025-05-22
The --config option with absolute paths will be loaded only once.

qraqras:
I needed to fix this project, not “ipython”, so I re-created the PR.
ipython/ipython#14902

Thanks.

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -900,7 +900,7 @@ class Application(SingletonConfigurable):
         yield each config object in turn.
         """
         if os.path.isabs(basefilename):
-            path = None
+            path = [None]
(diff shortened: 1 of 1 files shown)


traitlets/config/application.py
```diff
@@ -899,6 +899,8 @@ def _load_config_files(
 
         yield each config object in turn.
         """
+        if os.path.isabs(basefilename):
+            path = None
```
minrk, line 903: Probably need to make sure this case is tested, because this will fail on line 906 with:

```
---->  reversed(path)

TypeError: 'NoneType' object is not reversible
```

I think maybe you want:

```suggestion
            path = [None]
```

PR #943, opened 2026-05-29
Fix CI (docs build, mypy tests, downstream)

Carreau:


krassowski:
Looking at test failures we should pin mypy: https://github.com/ipython/traitlets/blob/77c31d32005d58f3f58aab75094d3c53f6e54250/pyproject.toml#L36

I mean of course it is better to address them but I find random failures showing up in unrelated PRs to be counterpoductive.

krassowski:
It looks like this is due to changes in mypy 1.20. https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-20. I will push a commit aligning the format.

krassowski (approved):

Carreau:
THanks. I'm suggesting we do a release soon; then drop 3.9

krassowski:
Starting a patch release now

krassowski:
https://github.com/ipython/traitlets/releases/tag/v5.15.1

PR #946, opened 2026-07-13
Fix three correctness bugs found in code review by claude Fable

Carreau:
- `HasTraits.class_own_trait_events` called the nonexistent `cls.events()` and raised `AttributeError` on every
  call — nothing in the suite exercised it. Fixed to call `trait_events()`, corrected the docstring, and removed an
  unreachable `tags` branch in `trait_events` that could never trigger (`EventHandler` never gets a `.tags`
  attribute).
 - `Application.flatten_flags` checked `isinstance(aliases, tuple)` — the output dict being built, never a tuple
  — instead of the loop variable `alias`. Tuple-valued alias keys like `("f", "foo")` were never exploded into
  individual names, so a flag sharing a name with one member of a tuple alias crashed argparse with `conflicting
  option strings` instead of merging correctly.
- `get_logger()` permanently cached whichever logger it returned first. If called before any `Application`
  existed (easy to hit via a config-file warning), the fallback library logger was cached forever and the app's
  configured logger was never returned afterward. Now checks `Application.initialized()` on every call; the
  now-pointless module-level cache is removed in favor of stdlib's own logger-by-name caching.

- Adds regression tests for all three.

- Also Deduplicate TraitTestBase in favor of the shipped copy.
    - `traitlets/tests/test_traitlets.py` and `tests/test_traitlets.py` each defined their own copy of `TraitTestBase`. Downstream projects (e.g. ipywidgets, in `widgets/tests/test_traits.py`) import `TraitTestBase` from the installed `traitlets.tests.test_traitlets` module, so that copy must stay. Dropped the duplicate from `tests/test_traitlets.py` in favor of importing it, and added a docstring noting the downstream dependency so it isn't mistaken for dead code again.

### DIFF
--- a/tests/config/test_application.py
+++ b/tests/config/test_application.py
@@ -510,6 +510,33 @@ class TestApplication(TestCase):
         # this would be app.config.Application.log_level if it failed:
         self.assertEqual(app.config.MyApp.log_level, "CRITICAL")
 
+    def test_flatten_aliases_tuple_keys(self):
+        # tuple alias keys must be exploded into one entry per name, so
+        # that the loader can detect collisions between flags and aliases
+        app = MyApp()
+        flags, aliases = app.flatten_flags()
+        self.assertEqual(aliases["fooi"], "Foo.i")
+        self.assertEqual(aliases["i"], "Foo.i")
+        self.assertNotIn(("fooi", "i"), aliases)
+
+    def test_flag_tuple_alias_collision(self):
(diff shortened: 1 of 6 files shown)


Carreau:
`from traitlets.tests.test_traitlets import TraitTestBase` is used by ipywidgets, either we deduplicate in the other direction; or fix / make a duplicate in ipywidgets.

PR #947, opened 2026-07-13
Remove dead code and stale Python 2 leftovers (with the help of Claude Fable)

Carreau:
- ClassTypes tuple in traitlets.py, a Python 2/3 compat constant never used and not exported
- execfile() shim in config/loader.py, shadowing the removed Python 2 builtin, never called
- Config._merge in config/loader.py, a deprecated private alias for Config.merge that was never called and never warned

### DIFF
--- a/traitlets/config/application.py
+++ b/traitlets/config/application.py
@@ -67,7 +67,7 @@ For instance, to set `C.a=[0,1,2]`, you may type this:
 """.strip()  # trim newlines of front and back
 
 # sys.argv can be missing, for example when python is embedded. See the docs
-# for details: http://docs.python.org/2/c-api/intro.html#embedding-python
+# for details: https://docs.python.org/3/c-api/intro.html#embedding-python
 if not hasattr(sys, "argv"):
     sys.argv = [""]
 
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -77,11 +77,6 @@ class ArgumentParser(argparse.ArgumentParser):
 # -----------------------------------------------------------------------------
 
 
-def execfile(fname: str, glob: dict[str, Any]) -> None:
-    with open(fname, "rb") as f:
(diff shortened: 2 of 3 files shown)


ccordoba12:
> Config._merge in config/loader.py, a deprecated private alias for Config.merge that was never called and never warned

@Carreau, unfortunately this change broke both Spyder and Spyder-kernels. Could you release 5.16.1 reverting it? That would give us time to address it in versions 6.1.7/3.1.6 of those packages, to be released at the end of August.

By the way, I don't remember why we're using `_merge` (perhaps because there was no `merge` at the time we wanted to merge configs), but it's been a stable API for at least 10 years.

Carreau:
just `_merge`? nothing else ?

I'll do that on monday.

Carreau:
Please feel free to send a PR next time (And see #963 for `_merge`)

And `_merge` had been marked as deprecated for at least 11 years; plus there is a leading underscore, so it's technically private.

ccordoba12:
> just _merge? nothing else ?

I think so. I mean, that's what I saw in our failing tests. But I'll check more thoroughly during the weekend.

> I'll do that on monday.

Thanks!

> Please feel free to send a PR next time (And see #963 for _merge)

Will do. I just wanted to see first if you were ok with the idea.

> And _merge had been marked as deprecated for at least 11 years; plus there is a leading underscore, so it's technically private.

As I said, I don't recall why we use `_merge`. Perhaps we started to use it more than 11 years ago. And sure, we should have changed it but you know how the life of a maintainer is.

Carreau:
We should also start to publish nightly of traitlets maybe.

ccordoba12:
This package is very stable, so I don't know if it's worth the effort (unless you're thinking to make important changes to it, of course).

Carreau:
Not many changes no, but even for the smallest one to trigger CI failures downstream in case like this.

PR #952, opened 2026-07-17
Update pre-commit hooks and fix the resulting fallout

Carreau:
Bump every pre-commit hook to a current release, including the mdformat repo move from executablebooks to hukkin, mypy v2.3.0 and ruff v0.15.22, then fix what the newer tools surface:

- mypy 2.0 narrowed the `int()` overloads, so `Int.validate` needs a `call-overload` ignore. Older versions report that ignore as unused, so require mypy>=2.0 rather than >=1.20; the reveal_type assertions in tests/test_typing.py still pass on 2.x.
- Annotate `_deprecations_shown` in utils/warnings.py.
- Document why `_CallbackWrapper` deliberately leaves `__hash__` unset.
- Reformat CHANGELOG.md (stray double spaces, bracket escaping).
- Ignore ruff PLC0415, since traitlets defers imports to break cycles and to keep optional/slow imports off the import path, and repo-review PP006, since the published extras stay the source of truth.
- Set pytest log_level = "info".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,8 +37,6 @@ test = [
     # `reveal_type()` representations were simplified (dropping the `builtins.`
     # prefix) in https://github.com/python/mypy/pull/20929; test assertions in
     # tests/test_typing.py target that newer output, so require >=1.20.
-    # mypy 2.0 narrowed the `int()` overloads, so traitlets.Int carries a
-    # `call-overload` ignore that older versions flag as unused: require >=2.0.
     "mypy>=2.0; implementation_name != 'pypy'",
     "pre-commit",
     "pytest-mock",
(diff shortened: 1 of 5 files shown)


Carreau:
I'm not sure whay pre-commit ci is not sending updates since 2024, so I removed and re-added it to the org.

pyproject.toml
```diff
@@ -37,7 +37,9 @@ test = [
     # `reveal_type()` representations were simplified (dropping the `builtins.`
     # prefix) in https://github.com/python/mypy/pull/20929; test assertions in
     # tests/test_typing.py target that newer output, so require >=1.20.
-    "mypy>=1.20; implementation_name != 'pypy'",
+    # mypy 2.0 narrowed the `int()` overloads, so traitlets.Int carries a
+    # `call-overload` ignore that older versions flag as unused: require >=2.0.
```
Carreau:
```suggestion
```

PR #956, opened 2026-07-21
Speed up Application/Configurable instance startup

Carreau:
Speeds up the **runtime** cost of building an `Application` and its many
`Configurable` sub-components at startup — the pattern used by IPython, Jupyter
server, ipywidgets, etc., which define hundreds of `HasTraits` subclasses at
import and instantiate a large component graph at launch.

The import-time half of this work ("Defer heavy imports to speed up import time")
has already landed separately on `main` (626bbe5), so this PR is now purely the
runtime optimizations. Each commit is independent and can be reviewed (or pulled)
on its own.

### Commits

1. **Reuse the class-namespace walk across the two metaclasses.**
   `MetaHasDescriptors.setup_class` already walks the full class namespace via
   `getmembers(cls)`; it now returns that `(name, value)` list so
   `MetaHasTraits.setup_class` can reuse it to find `TraitType` members instead
   of doing a second, redundant `dir(cls)` + `getattr` walk over every class.
   Same members in the same order, so semantics are identical (`getmembers`
   already skips members whose `getattr` raises, which is what the removed
   `try/except` handled).
   → class definition ~118 µs → ~96 µs per class.

2. **Cache metadata-filtered `class_traits()`/`traits()` results per class.**
   Filtering by metadata (e.g. `class_traits(config=True)`) was recomputed from
   scratch on every call — the single largest cost of `Application` startup
   (~25-30%), invoked ~45× per startup for results that are static per class
   (from `Application._classes_with_config_traits`,
   `KVArgParseConfigLoader._add_arguments`, and each `Configurable._load_config`).
   Both methods now delegate to a shared classmethod that memoizes the filtered
   dict per class and returns a `.copy()`, preserving the existing "fresh dict"
   contract (the cached dict never escapes by reference). `cls._traits` is frozen
   after class creation (`add_traits()` builds a new class rather than mutating),
   so the only way a filtered result can change is a post-hoc metadata mutation
   via `tag()`/`set_metadata()`; those bump a module-level generation counter and
   stale cache entries are recomputed. Only constant (non-callable, hashable)
   filters are cached; callable predicates stay on the uncached path.
   → `class_traits(config=True)` ~8.3 µs → ~1.3 µs per call.

   *Reviewer note:* the cache is invalidated by the supported post-construction
   metadata APIs (`tag()`/`set_metadata()`). Mutating `trait.metadata` as a raw
   dict after the class has already been queried is not reflected until the next
   generation bump — a pattern not used in traitlets and vanishingly rare in
   practice, but called out for awareness.

3. **Skip redundant re-validation of constructor kwargs.**
   `HasTraits.__init__` validated every trait kwarg twice: once via `setattr` in
   the fast loop, then again via `_cross_validate` + `set_trait`. The second pass
   is only needed for traits that actually have a cross-validator; the loop now
   guards on the same condition `_cross_validate` itself uses and, for traits
   without one, records the already-stored (possibly coerced) value for the
   notification instead of re-validating. Notification payloads are byte-for-byte
   identical.
   → instantiation with kwargs and no cross-validators ~1.2-1.4× faster.

4. **Skip config loading for Configurables with no matching config.**
   `Configurable._load_config` computed `traits(config=True)` and entered
   `hold_trait_notifications()` unconditionally, even for the many leaf
   Configurables whose config has no matching keys. It now computes `my_config`
   first and returns early when empty. Also removes a dead
   `section_names = self.section_names()` local that was computed (twice per
   instance) but never used.
   → ~1.2× on leaf Configurables with no matching config.

### End-to-end

On a synthetic app modeled on Jupyter/IPython scale (12 `Configurable`
components, 373 traits, config applied), `Application()` construct + `initialize()`
drops **~2196 µs → ~1740 µs (~21%)** on Python 3.11. The wins compound with the
number of Configurables built and traits scanned at startup.

All changes are behavior-preserving. Full test suite passes (including new tests
covering the cache's invalidation/copy contract, unhashable/callable filters, and
the deferred-import cold paths), along with mypy and ruff.

### DIFF
--- a/tests/test_traitlets.py
+++ b/tests/test_traitlets.py
@@ -937,6 +937,42 @@ class TestHasTraits(TestCase):
         traits = a.traits(config_key=lambda v: True)
         self.assertEqual(traits, dict(i=A.i, f=A.f, j=A.j))
 
+    def test_traits_metadata_filter_caching(self):
+        # metadata-filtered class_traits()/traits() results are memoized per
+        # class; make sure the cache preserves the "fresh dict" contract and is
+        # invalidated when metadata is mutated after class creation.
+        class A(HasTraits):
+            i = Int().tag(config=True)
+            j = Int()
+
+        # returned dict is a fresh copy the caller may mutate freely
+        first = A.class_traits(config=True)
+        self.assertEqual(first, dict(i=A.i))
+        first["injected"] = "oops"
+        self.assertEqual(A.class_traits(config=True), dict(i=A.i))
+
+        # tagging a trait after the result was cached must be reflected
+        A.j.tag(config=True)
+        self.assertEqual(A.class_traits(config=True), dict(i=A.i, j=A.j))
+        self.assertEqual(A().traits(config=True), dict(i=A.i, j=A.j))
+
+        # a subclass has its own cache and does not pollute the parent's
+        class B(A):
+            k = Int().tag(config=True)
+
+        self.assertEqual(B.class_traits(config=True), dict(i=A.i, j=A.j, k=B.k))
+        self.assertEqual(A.class_traits(config=True), dict(i=A.i, j=A.j))
+
+        # filters with non-hashable or callable values bypass the cache without
+        # error and still filter correctly
+        self.assertEqual(A.class_traits(config=[1, 2]), {})  # unhashable -> uncached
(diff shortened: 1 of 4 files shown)


Carreau:
It does not affects ipython startup visibly, but does hove an effect on raw 'python -c "import traitlets.config"'

PR #957, opened 2026-07-22
Add a document of 'best practices' on designing with traitlets

yuvipanda:
The Jupyter ecosystem uses traitlets a lot, but there's not a lot of 'design documentation' that tells you how to design your applications to use traitlets. This is the start that at least provides a collection of tips for folks

### DIFF
--- a/docs/source/best-practices.md
+++ b/docs/source/best-practices.md
@@ -1,6 +1,6 @@
-# Tips for using Traitlets
+# Best Practices
 
-A collection of useful design tips for using traitlets in your
+A collection of useful design tips & best practices for using traitlets in your
 application or library.
 
 ## Avoid using "pure config" objects
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -23,7 +23,7 @@ the configuration machinery.
     trait_types
     defining_traits
     api
-    tips
+    best-practices
     config
     config-api
     utils
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -23,6 +23,7 @@ the configuration machinery.
     trait_types
     defining_traits
     api
+    tips
     config
     config-api
     utils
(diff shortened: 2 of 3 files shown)


Carreau:
A few notes:

- It may seem obvious but this makes any classname and traitlets part of public API that can't be changed without breaking backward compatibility. 

- If you think come config will be useful by multiple classes you can actually create a empty superclass just with configuration option.

- Traitlets _can_, load config from .py files, json files **and from environment variable** though this is opt-in in the app that need to load from env _explicitely_ (all config have a APPLICATIONNAME_TRAIT_PATH=VALUE that is automatically exposed). But this can have security implication as thos _could_ contain executable code.

- register your configurables with the app to be shown when using --help and --help-all

- Maybe think / speak about `parents`:


```
class MyLeaf(Configurable):
	...

class SuperFoo(Configurable):
    
	def __init__(self):
		self.leaf = Leaf(parent=self)



class UltraBar(Configurable):
    
	def __init__(self):
		self.leaf = Leaf(parent=self)
```

then 

```
C.MyLeaf.values = ...
c.SuperFoo.MyLeaf.values = ...
c.UltraBar.MyLeaf.values = ...
```


Are all valid config options that apply in right context and can allow to avoid defining options with different names.

I would also add that it is tempting to add config options of everything, but it might be easier to rely on `DottedObjectName` and `Instance`, and usually just subclass in a config file to patch _just_ what is needed, than to provide hooks for anything and everything

yuvipanda:
Ty @Carreau!

Does the PR as is feel valuble enough to merge?

Carreau:
Always, i thought you wanted to expand on it :-) merging

yuvipanda:
ty ty @Carreau :) I appreciate the merge! I think all the ones in your comments are great, and I learnt a few things from there too (like env vars!)

PR #963, opened 2026-08-01
Re-add _merge to un-break spyder temporarily.

### LINKED ISSUE #962: Revert removal of _merge.
> Config._merge in config/loader.py, a deprecated private alias for Config.merge that was never called and never warned

@Carreau, unfortunately this change broke both Spyder and Spyder-kernels. Could you release 5.16.1 reverting it? That would give us time to address it in versions 6.1.7/3.1.6 of those packages, to be released at the end of August.

By the way, I don't remember why we're using `_merge` (perhaps because there was no `merge` at the time we wanted to merge configs), but it's been a stable API for at least 10 years.

_Originally posted by @ccordoba12 in https://github.com/ipython/traitlets/issues/947#issuecomment-5146922607_

Carreau:
closes #962

### DIFF
--- a/traitlets/config/loader.py
+++ b/traitlets/config/loader.py
@@ -245,6 +245,18 @@ class Config(dict):  # type:ignore[type-arg]
             if _is_section_key(key) and isinstance(obj, dict) and not isinstance(obj, Config):
                 setattr(self, key, Config(obj))
 
+    def _merge(self, other: t.Any) -> None:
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks @Carreau! I can confirm this is the only thing that's necessary to revert to not break the current Spyder/Spyder-kernels versions.

PR #972, opened 2026-08-09
Add LazyType trait for deferred class imports

Carreau:
## Summary

`Type` resolves a string `klass` — that is, imports it — as soon as the owning `HasTraits` object is created. For a default like `"ipykernel.debugger.Debugger"` that means paying for the `debugpy` import on every kernel startup, even though most sessions never debug.

`LazyType` subclasses `Type` and waits until the class is actually needed: reading or writing the trait, or reading `klass` / `default_value` off the trait itself.

```python
class A(HasTraits):
    encoder = LazyType("json.JSONEncoder")

a = A()      # json not imported
a.encoder    # now it is
```

**`Type` itself is untouched.**

## Key changes

- **New `LazyType(Type)`** — overrides `instance_init` to skip the eager resolution, and makes `klass` / `default_value` properties that import on first read and store the result in place.
- **`TraitType._resolve_lazily`** — tells `MetaHasTraits.setup_class` not to read `default_value` while building `_static_immutable_initial_values`. Without it the import fires at *class-creation* time and laziness is dead. None of that block's cases ever matched a `Type` subclass, so skipping it changes nothing else.

## Design notes

**Why properties rather than intercepting `validate()` / `default()`.** The smaller approach — resolve inside those two methods, leave the attributes as raw strings — is not equivalent. It makes `issubclass(x, trait.klass)` raise `TypeError`, hands `@observe` a string as `change.old`, and leaves the class permanently unresolved when a `@default` returns `None` under `allow_none`. The properties close all three.

**Strings, but not enforced.** Strings are what make this useful and the type hints say so, but a class object is accepted and stored as is — it is already imported, so there is nothing left to defer and such a trait simply behaves like `Type`.

## Consequences worth knowing

- A bad class name surfaces at first use rather than at construction.
- Generating help resolves the class, so `--help-all` costs the same as it does for `Type`.

## Testing

12 tests covering deferral through class definition and instantiation, resolution on read and on assignment, caching, `allow_none` with a `None` default staying lazy, observers receiving a resolved `change.old`, `@default`-returns-`None` not stranding the trait, forcing resolution via `_resolve_classes`, accepting an already-imported class, and that `Type` still resolves eagerly.

Full suite green, `mypy traitlets` clean, ruff and formatting unchanged from `main`.


tests/test_traitlets.py
```diff
@@ -1115,6 +1119,214 @@ class A(HasTraits):
...
+        with _importable_probe_module(tmp_path, name):
+            (tmp_path / f"{name}_re.py").write_text(f"from {name} import Klass, Sub\n")
+
+            class Svc(Configurable):
+                klass = LazyType(f"{name}_re.Klass", help="a class").tag(config=True)
+
+            assert name not in sys.modules
+            assert Svc.klass.info() == f"a subclass of '{name}.Klass'"
+            assert name in sys.modules
+            sys.modules.pop(f"{name}_re", None)
```
Carreau: @claude 

Remove this test, this is not a _requirement_ this is just something we are fine happening, we don't want to constrain future evolution

traitlets/traitlets.py
```diff
@@ -2178,6 +2186,90 @@ def default_value_repr(self) -> str:
...
+            allow_none: Literal[True] = ...,
+            **kwargs: t.Any,
+        ) -> None: ...
+
+        def __init__(self, *args: t.Any, **kwargs: t.Any) -> None: ...
+
+    # NOTE: no runtime __init__. Type's constructor already applies every
+    # defaulting rule and assigns through ``self.klass`` / ``self.default_value``,
+    # which the properties below intercept, validate and store.
+
```
Carreau: @claude delete this comment

traitlets/traitlets.py
```diff
@@ -2178,6 +2186,90 @@ def default_value_repr(self) -> str:
...
+    def instance_init(self, obj: t.Any) -> None:
+        """Deliberately does not resolve the class; see the class docstring."""
```
Carreau, line 2267:
```suggestion
        """Deliberately does not resolve the class; see the class docstring."""
        pass
```

PR #973, opened 2026-08-09
Drop type:ignore comments made unused by argcomplete 3.7

Carreau:
## Summary

The `Test Lint` job on `main` has been failing: argcomplete 3.7 ships inline annotations for `split_line()`, `debug()` and `CompletionFinder.__call__`, so the `no-untyped-call` ignores around those calls are now reported as `unused-ignore` and mypy exits non-zero.

(For the record, `packaging` was not involved — I confirmed the suite passes with `packaging` 26.3, and reverting argcomplete to 3.6.2 makes mypy clean again.)

## Changes

- Removed the `no-untyped-call` code from three ignores in `traitlets/config/argcomplete_config.py` (the `attr-defined` code is still needed and stays).
- Removed the whole ignore on `ExtendedCompletionFinder()` in `traitlets/config/loader.py`.
- Floored the mypy pre-commit hook's `argcomplete` dependency at `>=3.7` (was `>=3.1`), so the check means the same thing locally and in CI — on an older argcomplete the ignores would be required again.

The `no-untyped-call` ignores on `super()._get_completions` / `_get_option_completions` are untouched: those private methods are still unannotated.

## Verification

mypy clean over 24 source files and 706 tests passing against argcomplete 3.7.2. `Test Lint` is green on this branch.

https://claude.ai/code/session_0141ZmpAbm5FVsds8iNEotpp

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,7 +59,7 @@ repos:
         files: "^traitlets"
         stages: [manual]
         args: ["--install-types", "--non-interactive"]
-        additional_dependencies: ["argcomplete>=3.1"]
+        additional_dependencies: ["argcomplete>=3.7"]
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: "v1.10.0"
--- a/traitlets/config/argcomplete_config.py
+++ b/traitlets/config/argcomplete_config.py
@@ -45,7 +45,7 @@ def get_argcomplete_cwords() -> list[str] | None:
             _cword_suffix,
             comp_words,
             _last_wordbreak_pos,
-        ) = argcomplete.split_line(comp_line, comp_point)  # type:ignore[attr-defined,no-untyped-call]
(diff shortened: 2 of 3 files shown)


Carreau:
`Test Lint` is green now, which was the point of this PR. The only red check left is `codecov/patch`, and I don't plan to chase it — here's why.

This diff only edits `# type: ignore` comments; no executable code changed. Codecov still counts the three touched lines as "patch", and one of them — the `argcomplete.debug(...)` call at `traitlets/config/argcomplete_config.py:76`, inside the fallback for a non-integer `$_ARGCOMPLETE` — has never been covered. That gives 2/3 = 66.66% against the 80% patch target. The `traitlets` component delta settled at `(ø)`: no coverage was actually lost.

Making that check green would mean adding a test whose only purpose is to exercise a line this PR doesn't touch the behaviour of. Happy to do it if you'd rather have the check green than the diff minimal — just say so.

---

PR #974, opened 2026-08-09
Add proper deprecation warnings for deprecated module attributes

Carreau:
## Summary
This PR improves deprecation handling across the traitlets codebase by implementing proper `DeprecationWarning` emissions for deprecated APIs instead of silent aliases. This follows PEP 562 for module-level `__getattr__` and ensures users are properly notified when using deprecated features.

## Key Changes

- **traitlets/traitlets.py**: Replaced silent alias `NoDefaultSpecified = Undefined` with a module-level `__getattr__` implementation that emits a `DeprecationWarning` when the deprecated name is accessed. Added `__dir__()` to keep deprecated attributes discoverable.

- **traitlets/config/configurable.py**: Added explicit `DeprecationWarning` to `Configurable._get_log_handler()` method with updated docstring indicating the deprecation version and recommended alternative.

- **traitlets/config/loader.py**: Converted `Config.has_key` from a simple alias to a proper method that emits a `DeprecationWarning` before delegating to `__contains__()`.

- **tests/test_traitlets.py**: Added `test_no_default_specified_deprecated()` to verify the deprecation warning is emitted and that the deprecated name remains discoverable via `dir()`.

- **tests/config/test_loader.py**: Added `test_has_key_deprecated()` to verify `Config.has_key()` emits appropriate deprecation warnings.

- **tests/config/test_configurable.py**: Added `test_get_log_handler_deprecated()` to verify `_get_log_handler()` emits a deprecation warning.

## Implementation Details

The changes use Python's standard `warnings.warn()` with `DeprecationWarning` and `stacklevel=2` to ensure warnings point to the caller's code. The module-level `__getattr__` approach (PEP 562) allows deprecated names to be accessed with warnings while keeping the module namespace clean and maintaining discoverability through `__dir__()`.

https://claude.ai/code/session_01AEeaWWBdmyp9x2DaCombXD

### DIFF
--- a/tests/config/test_configurable.py
+++ b/tests/config/test_configurable.py
@@ -969,3 +969,9 @@ def test_logger_adapter(caplog, capsys):
     app.log.info("test message")
 
     assert "adapted test message" in capsys.readouterr().err
+
+
+def test_get_log_handler_deprecated():
+    app = Application(log=logging.getLogger("test_get_log_handler_deprecated"))
+    with pytest.deprecated_call(match="_get_log_handler has been deprecated"):
+        app._get_log_handler()
--- a/tests/config/test_loader.py
+++ b/tests/config/test_loader.py
@@ -503,6 +503,14 @@ class TestConfig(TestCase):
         del c.A
         self.assertEqual(c.A, Config())
 
+    def test_has_key_deprecated(self):
+        c = Config()
+        c.a = 10
+        with pytest.deprecated_call(match="Config.has_key has been deprecated"):
+            assert c.has_key("a")
+        with pytest.deprecated_call(match="Config.has_key has been deprecated"):
+            assert not c.has_key("b")
+
     def test_merge_doesnt_exist(self):
         c1 = Config()
         c2 = Config()
--- a/tests/test_traitlets.py
+++ b/tests/test_traitlets.py
@@ -367,6 +367,17 @@ class TestTraitType(TestCase):
             class C(HasTraits):
                 t = Dict(Int)
 
+    def test_no_default_specified_deprecated(self):
+        import traitlets.traitlets as tt
+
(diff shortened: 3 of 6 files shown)


Carreau:
The two red downstream checks (`downstream / jupyter_server` and the `downstream_check` gate that aggregates it) are **not caused by this PR**, and can't be fixed from this repo.

Single failure, in jupyter_server's own suite:

```
FAILED tests/test_utils.py::test_check_version - packaging.version.InvalidVersion: Invalid version: 1.0
1 failed, 1121 passed, 32 skipped
```

Root cause is a `packaging` behaviour change. `test_check_version` deliberately passes a float:

```python
assert check_version(1.0, "1.0.1")  # type:ignore[arg-type]
```

and jupyter_server's `check_version` only catches `TypeError`:

```python
try:
    return bool(Version(v) >= Version(check))
except TypeError:
    return True
```

Older `packaging` raised `TypeError` from the regex on a non-str; `packaging` >= 26 raises `InvalidVersion` instead, which isn't caught. Reproduced locally against **unmodified** traitlets:

| packaging | `Version(1.0)` | test |
|---|---|---|
| 24.0 | `TypeError` (caught) | passes |
| 26.3 | `InvalidVersion: Invalid version: 1.0` | fails |

Nothing in this PR touches version parsing — the diff only adds `DeprecationWarning`s to three already-deprecated APIs. The last `downstream` run on `main` was 2026-08-03, before `packaging` 26 landed, which is why `main` isn't showing this yet; it should go red on its next run too.

The fix belongs in jupyter_server (catch `InvalidVersion` as well, or stop passing a float). Every other check on this PR is green, including `Test Lint`.

---

Carreau:
Failures will be fixed by https://github.com/jupyter-server/jupyter_server/pull/1688

PR #977, opened 2026-08-16
Use org-level security policy (ipython/.github/SECURITY.md)

Yann-P:
Do not define SECURITY.mds at the repo-level but rather as org-level. This also fixes the wrong/old email address being used here.

jupyter/security#125 and jupyter/security#124


Carreau:
thanks !

PR #984, opened 2026-08-28
support mock.patch.object on HasTraits

minrk:
`mock.patch.object` requires allowing `delattr(has_traits, "trait")` to work and `hasattr(has_traits, "trait")` to be False afterward.

This fixes mock.patch.object on HasTraits objects, with a previously failing test.

Without this patch, test fails with:

```
    def __exit__(self, *exc_info):
        """Undo the patch."""
        if not self.is_started:
            return

        if self.is_local and self.temp_original is not DEFAULT:
            setattr(self.target, self.attribute, self.temp_original)
        else:
>           delattr(self.target, self.attribute)
E           AttributeError: __delete__
```

Initially, I had it just clear the state without the sentinel (could be useful for restoring defaults, for example), but in that case `hasattr(obj, 'trait')` returns True and mock.patch doesn't restore the before-patch value, it always returns to the default state.

I think the logic in unittest.mock is wrong in how it decides whether to set the original value back, but this is what we need to work with the current standard library.

### DIFF
--- a/tests/test_traitlets.py
+++ b/tests/test_traitlets.py
@@ -12,7 +12,7 @@ import pathlib
 import pickle
 import re
 import typing as t
-from unittest import TestCase
+from unittest import TestCase, mock
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


Carreau:
pushed a commit to fix linting.

PR #986, opened 2026-09-08
Make Set trait generic

### LINKED ISSUE #927: traitlets.Set should support generic type annotations for type safety
## Problem

`traitlets.Set` currently lacks generic type support, making it impossible to specify the type of elements it contains for proper type checking.

## Current Situation

```python
from traitlets import Set
import traitlets

class MyClass(traitlets.HasTraits):
    # This works but provides no type information
    my_set = Set()
    
    # This doesn't work - no generic support
    # my_set: Set[str] = Set()  # Type error
```

## Desired Behavior

`traitlets.Set` should support generic type annotations similar to how Python's built-in collections work:

```python
from traitlets import Set
import traitlets

class MyClass(traitlets.HasTraits):
    # Should be possible for type safety
    room_ids: Set[str] = Set()
    numbers: Set[int] = Set(default_value={1, 2, 3})
```

## Reference

The `Set` class is implemented in [`traitlets/traitlets.py`](https://github.com/ipython/traitlets/blob/main/traitlets/traitlets.py) as a container trait.

GruffElixir:
Fixes #927

Make Set preserve its element type parameter, matching List and the other container traits. This enables annotations such as values: Set[str] = Set() to expose Set[str] on the class and set[str] on instances.

Tests: python -m pytest tests/test_typing.py -q


GruffElixir:
The project tests, lint, Codecov, and downstream ipython/ipywidgets/nbconvert/notebook checks are passing. The two remaining failures appear external or transient: Read the Docs stopped during checkout with "fatal: reference is not a tree: b16490a6268bb7f0174266ecd44e3f5e0559a798", and the jupyter_server downstream job had one kernel-cull timing failure after 1,153 passes. Could a maintainer rerun those checks when convenient?

GruffElixir:
The two red checks are environmental rather than Set typing failures: Read the Docs could not resolve the fork head SHA (`fatal: reference is not a tree`), and the jupyter_server downstream job failed in its timing-sensitive `test_cull_connected` case while the rest of the matrix passed. Could you rerun those checks on the current head?

